PR #4, opened 2020-01-16
🐛 Use existing Trove classifiers in dist metadata

### LINKED ISSUE #3: Trove classifiers in dist metadata are invalid
If you try to publish a dist with trove classifiers (https://github.com/ansible/pylibssh/blob/362a7e3/setup.py#L22-L26) that aren't in the list https://pypi.org/classifiers/, it'll be rejected.

webknjaz:
This change gets rid of invalid Trove classifiers in favor of existing
ones. Basically, it untangles these classifiers:

* Operating System :: MacOS
* Operating System :: POSIX :: Linux
* Topic :: Software Development :: Libraries :: Python Modules
* Topic :: Security

Ref: https://pypi.org/classifiers/

Fixes #3

Signed-Off-By: Sviatoslav Sydorenko <webknjaz@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,10 +19,12 @@ setup(
     packages=find_packages('lib'),
     description="Python bindings for libssh client",
     classifiers=[
-        "Programming Language :: Cython",
         "Development Status :: 2 - Pre-Alpha",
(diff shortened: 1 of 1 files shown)


ganeshrn (approved):
Thank you!

PR #9, opened 2020-01-20
Add unit test structure

ganeshrn:


test/units/test_channel.py
```diff
@@ -0,0 +1,5 @@
+import unittest
```
webknjaz, line 1: The modern-day standard is to use pytest

ganeshrn: noted, will update it as I add actual test cases

PR #13, opened 2020-01-21
Modify installation instructions

Akasurde:
Instructions are missing step to install requirements.txt using pip

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -35,6 +35,7 @@ Build the extension:
 
     git clone https://github.com/ansible/pylibssh.git
     cd pylibssh
+    pip install -r requirements.txt
     python3 setup.py build_ext --inplace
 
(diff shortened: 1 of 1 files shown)


ganeshrn (commented):
As per @webknjaz the file name should end with `.in` extension. Also since it is build time dependency the file can be renamed as `requirements-dev.in`

webknjaz:
@ganeshrn yep

webknjaz:
But this can be done in a separate PR, no pressure here.

webknjaz (approved):

webknjaz:
I'd probably use "build" instead of "dev" since it better reflects the purpose.

PR #22, opened 2020-01-23
Update setup.py to dynamically fetch sources list 

ganeshrn:


.gitignore
```diff
@@ -185,14 +185,16 @@ __pycache__/
 # C extensions
 *.so
 
+# C files
+*.c
```
webknjaz, line 189: This still should be in manifest

setup.py
```diff
@@ -12,23 +13,32 @@
 
 LIB_NAME = 'ssh'
 
-
 sys.path.insert(0, os.path.abspath('lib'))
 
 
+def _get_sources(path):
```
webknjaz, line 19: could as well be `_get_sources = glob`

lib/pylibssh/version.pyx
```diff
@@ -0,0 +1,3 @@
+from pylibssh.includes.libssh cimport LIBSSH_VERSION
+
+libssh_version = LIBSSH_VERSION.decode("ascii")
```
webknjaz, line 3: Maybe keep this uppercase as a constant?

PR #38, opened 2020-04-28
Build macOS wheels in GitHub Actions workflow

webknjaz:


bin/pep517_backend.py
```diff
@@ -191,5 +191,4 @@ def func_wrapper(*args, **kwargs):  # noqa: WPS210, WPS430
     return func_wrapper
 
 
-build_sdist = pre_build_cython(build_sdist)  # pylint: disable=invalid-name
```
webknjaz, line 194: NOTE: This was probably removed because sdists don't need `*.so` files. OTOH, it'd be great to include `C`-files I think.

PR #40, opened 2020-05-05
Add bulk response recv support

ganeshrn:
##### SUMMARY

Add API to receive a bulk response from the remote shell

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -15,8 +15,13 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 #
-from libc.string cimport memset
+import time
+
+from io import BytesIO
+
 from cpython.bytes cimport PyBytes_AS_STRING
+from cpython.mem cimport PyMem_Malloc, PyMem_Realloc, PyMem_Free
+from libc.string cimport memcpy, memset
 
 from pylibsshext.session cimport get_libssh_session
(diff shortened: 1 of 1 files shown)


src/pylibsshext/channel.pyx
```diff
@@ -105,6 +112,25 @@ cdef class Channel:
     def sendall(self, data):
         return self.write(data)
 
+    def read_bulk_response(self, size=1024, stderr=0, timeout=0.001, retry=5):
+        recv = BytesIO()
```
webknjaz, line 116: It's better to always use a context manager because currently, you'll have an unclosed resource if any exception happens in the middle of the loop below.

src/pylibsshext/channel.pyx
```diff
@@ -105,6 +112,25 @@ cdef class Channel:
...
+            if not data:
+                if timeout:
+                    time.sleep(timeout)
+            else:
+                recv.write(data)
+                offset = recv.tell() - size if recv.tell() > size else 0
+                recv.seek(offset)
+                response += recv.read()
+
+            retry = retry - 1
```
webknjaz, line 129: There's a better way to do retries — exponential backoff: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

Here's a lib that we used for the migration script: https://pypi.org/project/backoff/

PR #41, opened 2020-05-05
Fix py2 compatibility issues

ganeshrn:
##### SUMMARY

*  `CompletedProcess` used for response handling
   is added in py >= 3.5
*  Add `CompletedProcess` class defination
   for channel response handling in channel.pyx
   to make it py2 compilant.
*  Fix review comments to use context handling for
   Bytes buffer.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
 
 from pylibsshext.session cimport get_libssh_session
 from pylibsshext.errors cimport LibsshChannelException
-from subprocess import CalledProcessError, CompletedProcess
+from subprocess import CalledProcessError
+
+
+# Defined CompletedProcess here to be compliant with py2
+class CompletedProcess(object):
+    """A process that has finished running.
+    This is returned by run().
+    Attributes:
+      args: The list or str args passed to run().
+      returncode: The exit code of the process, negative for signals.
+      stdout: The standard output (None if not captured).
+      stderr: The standard error (None if not captured).
+    """
+    def __init__(self, args, returncode, stdout=None, stderr=None):
+        self.args = args
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+
+    def __repr__(self):
+        args = ['args={!r}'.format(self.args),
+                'returncode={!r}'.format(self.returncode)]
+        if self.stdout is not None:
+            args.append('stdout={!r}'.format(self.stdout))
+        if self.stderr is not None:
+            args.append('stderr={!r}'.format(self.stderr))
+        return "{}({})".format(type(self).__name__, ', '.join(args))
+
+
+    def check_returncode(self):
+        """Raise CalledProcessError if the exit code is non-zero."""
+        if self.returncode:
+            raise CalledProcessError(self.returncode, self.args, self.stdout,
+                                     self.stderr)
+
+
+class BufferManager():
+    def __init__(self, filename, mode):
+        self.recv = BytesIO()
+
+    def __enter__(self):
+        return self.recv
+
+    def __exit__(self):
+        if self.recv:
+            self.recv.close()
 
 
 cdef int _process_outputs(libssh.ssh_session session,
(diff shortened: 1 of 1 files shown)


src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+
+
+    def check_returncode(self):
+        """Raise CalledProcessError if the exit code is non-zero."""
+        if self.returncode:
+            raise CalledProcessError(self.returncode, self.args, self.stdout,
+                                     self.stderr)
+
+
+class BufferManager():
```
webknjaz, line 64: No need to have braces here
```suggestion
class BufferManager:
```

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
 
 from pylibsshext.session cimport get_libssh_session
 from pylibsshext.errors cimport LibsshChannelException
-from subprocess import CalledProcessError, CompletedProcess
+from subprocess import CalledProcessError
```
webknjaz, line 28: This should go after line 20 because it's an stdlib import, not a local one

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+
```
webknjaz, line 29: Add this boilerplate to make classes in py2 new-style by default
```suggestion

__metaclass__ = type

```

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+
+# Defined CompletedProcess here to be compliant with py2
+class CompletedProcess(object):
```
webknjaz, line 32: No need to explicitly inherit object with that boilerplate above:
```suggestion
class CompletedProcess:
```

webknjaz, line 32: Also, I'd probably put it into some `_compat` module instead. It doesn't really belong here.

Besides, I'd use the following style:
```python
try:
    from subprocess import CompletedProcess
except ImportError:
    class CompletedProcess:
        ...
```

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+    """A process that has finished running.
```
webknjaz, line 33: a docstring title must be separated from the body as per PEP257
```suggestion
    """A process that has finished running.

```

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+        self.stderr = stderr
+
+    def __repr__(self):
+        args = ['args={!r}'.format(self.args),
+                'returncode={!r}'.format(self.returncode)]
+        if self.stdout is not None:
+            args.append('stdout={!r}'.format(self.stdout))
+        if self.stderr is not None:
+            args.append('stderr={!r}'.format(self.stderr))
+        return "{}({})".format(type(self).__name__, ', '.join(args))
```
webknjaz, line 54: this should probably be 
```suggestion
        return "<{}({})>".format(type(self).__name__, ', '.join(args))
```

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+      stdout: The standard output (None if not captured).
+      stderr: The standard error (None if not captured).
+    """
+    def __init__(self, args, returncode, stdout=None, stderr=None):
+        self.args = args
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+
+    def __repr__(self):
```
webknjaz, line 47: Should it also have `__str__`?

src/pylibsshext/channel.pyx
```diff
@@ -25,7 +25,52 @@ from libc.string cimport memcpy, memset
...
+
+    def check_returncode(self):
+        """Raise CalledProcessError if the exit code is non-zero."""
+        if self.returncode:
+            raise CalledProcessError(self.returncode, self.args, self.stdout,
+                                     self.stderr)
+
+
+class BufferManager():
+    def __init__(self, filename, mode):
```
webknjaz, line 65: This wouldn't work because you're not passing the args. OTOH this class is unnecessary anyway.

src/pylibsshext/channel.pyx
```diff
@@ -113,22 +158,21 @@ cdef class Channel:
...
-            else:
-                recv.write(data)
-                offset = recv.tell() - size if recv.tell() > size else 0
-                recv.seek(offset)
-                response += recv.read()
-
-            retry = retry - 1
-
-        recv.close()
+        with BufferManager() as recv:
```
webknjaz, line 162: This extra wrapper is unnecessary, it's already a CM:
```suggestion
        with BytesIO() as recv:
```

src/pylibsshext/channel.pyx
```diff
@@ -113,22 +158,21 @@ cdef class Channel:
         return self.write(data)
 
     def read_bulk_response(self, size=1024, stderr=0, timeout=0.001, retry=5):
-        recv = BytesIO()
         response = b""
```
webknjaz, line 161: This should be closer to `memoryview`/`buffer` or at least `BytesIO` up until returned for performance. I'll refactor it.

PR #42, opened 2020-05-05
Enable cython checks in pre-commit

webknjaz:


.pre-commit-config.yaml
```diff
@@ -4,11 +4,18 @@ repos:
...
+    #   ^.*\.p(xd|y|yx)$
+    # types: [file]
 
 - repo: git://github.com/asottile/add-trailing-comma
   rev: v0.7.1
   hooks:
   - id: add-trailing-comma
+    files: >-
+      ^.*\.p(xd|y|yx)$
+    types: [file]
```
asottile, line 18: this doesn't work fwiw, add-trailing-comma uses the ast to find the things to fix

webknjaz: But also this doesn't fail...

asottile: because syntax errors are currently silently ignored because people got grumpy when it was trying to fix invalid code 🤷

PR #51, opened 2020-05-15
Implement initial `ssh_client_session` fixture

webknjaz:
##### SUMMARY

This PR attempts to implement a public-private keypair auth in the session
with a specific key that is already present in memory.

The idea is that the caller should be able to feed bytes and an optional password
into the connect method.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

atm the auth attempt gets rejected for the private key that is 100% correct.

### DIFF
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -358,15 +358,18 @@ cdef class Session(object):
             NULL, NULL,
             &_private_key,
         )
+
         rc = libssh.ssh_userauth_publickey(
-            self._libssh_session, NULL, &_private_key,
+            self._libssh_session, NULL, _private_key,
         )
+
         if rc != libssh.SSH_AUTH_SUCCESS:
             raise LibsshSessionException(
                 "Failed to authenticate a specific public key: "
                 "{!s} (RC={!r})".
                 format(self._get_session_error_str(), rc),
             )
+        libssh.ssh_key_free(_private_key)
 
     def authenticate_pubkey(self):
         cdef int rc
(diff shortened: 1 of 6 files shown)


src/pylibsshext/session.pyx
```diff
@@ -302,6 +320,54 @@ cdef class Session(object):
...
+        if private_key_password is not None:
+            if isinstance(private_key_password, bytes):
+                b_password = private_key_password
+            else:
+                b_password = private_key_password.encode()
+            c_password = b_password
+        libssh.ssh_pki_import_privkey_base64(
+            c_private_key_b64, c_password,
+            NULL, NULL,
+            &_private_key,
```
automate16, line 359: this variable is dynamic memory allocated and should be freed before existing the function

src/pylibsshext/session.pyx
```diff
@@ -302,6 +320,54 @@ cdef class Session(object):
...
+        )
+        rc = libssh.ssh_userauth_publickey(
+            self._libssh_session, NULL, &_private_key,
+        )
```
ganeshrn, line 364: Free `_private_key` variable as it is dynamically allocation memory
`libssh.ssh_key_free(_private_key)`

src/pylibsshext/session.pyx
```diff
@@ -302,6 +320,54 @@ cdef class Session(object):
                                       self._fingerprint_py,
                                       know_host_msg)
 
+    def authenticate_specific_pubkey(
```
ganeshrn, line 323: I tested after applying the changes here https://gist.github.com/ganeshrn/46eafd00e1f91ba4d5a05d7b1a441ab5 on top of your PR and it is working fine for me now

webknjaz: Applied, let's see what happens.

PR #54, opened 2020-05-30
Track down the cause of `test_get` fs permission error under Python 2.7

ganeshrn:
##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -100,5 +100,10 @@ cdef class SFTP:
                     raise LibsshSFTPException("Opening remote file [%s] for read failed" % remote_file)
         sftp.sftp_close(rf)
 
+    def close(self):
+        if self._libssh_sftp_session is not NULL:
+            sftp.sftp_free(self._libssh_sftp_session)
+            self._libssh_sftp_session = NULL
+
     def _get_sftp_error_str(self):
         return sftp.sftp_get_error(self._libssh_sftp_session)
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -114,7 +114,13 @@ def ssh_client_session(sshd_addr, ssh_clientkey_path):
         host_key_checking=False,
         look_for_keys=False,
     )
-    return ssh_session
+    try:  # noqa: WPS501
+        yield ssh_session
+    finally:
+        if hasattr(ssh_session, 'sftp') and ssh_session.sftp is not None:
```
webknjaz: Why would this attr be missing? If there are cases where it's true, it's a design flaw.

src/pylibsshext/sftp.pyx
```diff
@@ -100,5 +100,10 @@ cdef class SFTP:
                     raise LibsshSFTPException("Opening remote file [%s] for read failed" % remote_file)
         sftp.sftp_close(rf)
 
+    def close(self):
+        if self._libssh_sftp_session is not NULL:
+            sftp.sftp_free(self._libssh_sftp_session)
+            self._libssh_sftp_session = NULL
```
webknjaz: I usually like the guard expression style for reducing indents:
```suggestion
        if self._libssh_sftp_session is NULL:
            return
        sftp.sftp_free(self._libssh_sftp_session)
        self._libssh_sftp_session = NULL
```

ganeshrn: Since this is one level indentation, it shouldn't be an issue. I try to avoid return in between the method whereever possible.

webknjaz: Why? `return` does nothing in terms of changing the method's return value.

webknjaz:
I've cherry-picked the "close" fix from this branch into devel and kept only `test_get` debug attempts here.

PR #58, opened 2020-06-04
Fix sftp file get issue on py2

ganeshrn:
##### SUMMARY

*  the file write() method returns None on py2
   is bytes are written to file successfully, whereas
   on py3 it returns total number of bytes written
   to file.
*  Add fix to check for the number of bytes written only
   in the case when write() is not returning None

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -91,13 +91,17 @@ cdef class SFTP:
                 break
             elif file_data < 0:
                 sftp.sftp_close(rf)
-                raise LibsshSFTPException("Opening remote file [%s] for read failed" % remote_file)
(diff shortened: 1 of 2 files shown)


webknjaz:
This whole linter pleasing dance could've been avoided by simple `git revert 0bf0a6c482f2ddb853ea1551dfdac00e6716dd10`

ganeshrn:
@webknjaz Yeah my bad...Adding a commit pre-hook to fail on lint issues is due for me now :-)

webknjaz:
FWIW I set up the CI so that the linters don't cancel test jobs.

webknjaz:
@ganeshrn do you think it's ready to go?

ganeshrn:
@webknjaz Yup should be ready to go

PR #107, opened 2020-07-22
Support keyboard-authentication auth

### LINKED ISSUE #105: Password authentication fails with Arista vEos
##### ISSUE TYPE
- Bug Report

```paste below
devel
```

##### STEPS TO REPRODUCE
clone repo and build the binaries
Refer: https://github.com/ansible/pylibssh#building-the-module

Add binary file PYTHONPATH
```console
$ export PYTHONPATH="${PYTHONPATH}:<path-to-cloned-repo>/src"
```

Run below python script

```python
import os

from pylibsshext.errors import LibsshSessionException

ssh = Session()

HOST = "<changeme>"
USER = "<changeme>"
PASSWORD = BECOME_PASSWORD = "<changeme>"


try:
    ssh.connect(
        host=HOST,
        user=USER,
        password=PASSWORD,
        timeout=30,
        port=22,
        look_for_keys=False
    )
except LibsshSessionException as ex:
    print(str(ex))

print(ssh.is_connected)
```

##### EXPECTED RESULTS

Password authentication should work

##### ACTUAL RESULTS

```paste below
Password authentication fails
```

Qalthos:
##### SUMMARY

Adds support for keyboard-interactive auth and tries to intelligently pick auth type

Fixes #105

##### ISSUE TYPE

- Bugfix Pull Request
- Feature Pull Request

### DIFF
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -416,26 +416,29 @@ cdef class Session(object):
         :rtype: NoneType
         """
         cdef int rc
+        cdef char should_echo
         rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
-        if rc == libssh.SSH_AUTH_SUCCESS:
-            return
 
-        if rc == libssh.SSH_AUTH_INFO:
+        while rc == libssh.SSH_AUTH_INFO:
             prompt_count = libssh.ssh_userauth_kbdint_getnprompts(self._libssh_session)
-            if prompt_count == 0:
-                # No prompts, but still need to call kbdint again
-                pass
-            elif prompt_count == 1:
-                rc = libssh.ssh_userauth_kbdint_setanswer(self._libssh_session, 0, password.encode())
-            else:
-                raise LibsshSessionException("Server aksed more questions than I was expecting ({count}), bailing...".format(count=prompt_count))
+            if prompt_count > 0:
+                for prompt in range(prompt_count):
+                    prompt_text = libssh.ssh_userauth_kbdint_getprompt(self._libssh_session, prompt, &should_echo)
+                    if prompt_text == b"Password: ":
+                        break
+                else:
+                    raise LibsshSessionException("None of the prompts looked like password prompts to me")
+                rc = libssh.ssh_userauth_kbdint_setanswer(self._libssh_session, prompt, password.encode())
+
+            # We didn't get SSH_AUTH_SUCCESS this time, so we have to call
+            # userauth_kbdint again until we do (or fail)
+            rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
 
         if rc in (libssh.SSH_AUTH_ERROR, libssh.SSH_AUTH_DENIED):
             raise LibsshSessionException("Failed to authenticate with keyboard-interactive: {err}".format(err=self._get_session_error_str()))
 
-        # If we didn't get SSH_AUTH_SUCCESS this time, we have to call
-        # userauth_kbdint again until we do (or fail)
-        return self.authenticate_interactive(password)
+        if rc == libssh.SSH_AUTH_SUCCESS:
+            return
 
     def new_channel(self):
         return Channel(self)
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -441,8 +441,6 @@ cdef class Session(object):
         if rc in (libssh.SSH_AUTH_ERROR, libssh.SSH_AUTH_DENIED):
             raise LibsshSessionException("Failed to authenticate with keyboard-interactive: {err}".format(err=self._get_session_error_str()))
 
-        if rc == libssh.SSH_AUTH_SUCCESS:
-            return
 
     def new_channel(self):
         return Channel(self)
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -227,7 +227,7 @@ cdef class Session(object):
             # Huh, it worked?
             return
         if rc == libssh.SSH_AUTH_ERROR:
-            raise LibsshSessionException("Something went wrong ):")
+            raise LibsshSessionException("Error while fetching list of supported authentication methods")
 
         supported_auth = libssh.ssh_userauth_list(self._libssh_session, NULL)
 
@@ -238,33 +238,37 @@ cdef class Session(object):
                     kwargs['private_key'],
                     kwargs.get('private_key_password'),
                 )
-                return
             except LibsshSessionException as ex:
                 saved_exception = ex
+            else:
+                return
 
         # try authenticating with a password
         if kwargs.get('password') and supported_auth & libssh.SSH_AUTH_METHOD_PASSWORD:
             try:
                 self.authenticate_password(kwargs["password"])
-                return
             except LibsshSessionException as ex:
                 saved_exception = ex
+            else:
+                return
 
         # try authenticating with keyboard-interactive (without the interactive)
         if kwargs.get('password') and supported_auth & libssh.SSH_AUTH_METHOD_INTERACTIVE:
             try:
                 self.authenticate_interactive(kwargs["password"])
-                return
             except LibsshSessionException as ex:
                 saved_exception = ex
+            else:
+                return
 
         if kwargs.get('look_for_keys', True) and supported_auth & libssh.SSH_AUTH_METHOD_PUBLICKEY:
             # try authenticating with public keys
             try:
                 self.authenticate_pubkey()
-                return
             except LibsshSessionException as ex:
                 saved_exception = ex
+            else:
+                return
 
         if saved_exception is not None:
             libssh.ssh_disconnect(self._libssh_session)
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -243,8 +243,8 @@ cdef class Session(object):
             else:
                 return
 
-        # try authenticating with a password
         if kwargs.get('password') and supported_auth & libssh.SSH_AUTH_METHOD_PASSWORD:
+            # try authenticating with a password
             try:
                 self.authenticate_password(kwargs["password"])
             except LibsshSessionException as ex:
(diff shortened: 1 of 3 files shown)


src/pylibsshext/session.pyx
```diff
@@ -231,14 +233,22 @@ cdef class Session(object):
...
                 saved_exception = ex
 
-        if kwargs.get('look_for_keys', True):
+        # try authenticating with keyboard-interactive (but not really)
+        if kwargs.get('password') and supported_auth & libssh.SSH_AUTH_METHOD_INTERACTIVE:
+            try:
+                self.authenticate_interactive(kwargs["password"])
+                return
+            except LibsshSessionException as ex:
+                saved_exception = ex
```
webknjaz: It's usually a bad idea to have more than one statement in try/except block:
```suggestion
                self.authenticate_interactive(kwargs["password"])
            except LibsshSessionException as ex:
                saved_exception = ex
            else:
                return
```

src/pylibsshext/session.pyx
```diff
@@ -383,6 +393,19 @@ cdef class Session(object):
         if rc == libssh.SSH_AUTH_ERROR or rc == libssh.SSH_AUTH_DENIED:
             raise LibsshSessionException("Failed to authenticate with password: %s" % self._get_session_error_str())
 
+    def authenticate_interactive(self, password):
```
webknjaz, line 413: add a PEP257 compliant docstring plz

src/pylibsshext/session.pyx
```diff
@@ -383,6 +393,19 @@ cdef class Session(object):
...
+        cdef int rc
+        rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
+        if rc == libssh.SSH_AUTH_INFO:
+            prompt_count = libssh.ssh_userauth_kbdint_getnprompts(self._libssh_session)
+            if prompt_count == 1:
+                rc = libssh.ssh_userauth_kbdint_setanswer(self._libssh_session, 1, password.encode())
+            else:
+                raise LibsshSessionException("Server aksed more questions than I was expecting (%s), bailing..." % prompt_count)
```
webknjaz: Let's turn this into a guard expression and dedent some bits. Also, let's use str.format for forward compat:
```suggestion
            if prompt_count != 1:
                raise LibsshSessionException(
                    "Server aksed more questions than I was expecting "
                    "({prompt_num!s}), bailing...".
                    format(prompt_num=prompt_count)
                )

            rc = libssh.ssh_userauth_kbdint_setanswer(self._libssh_session, 1, password.encode())
```

webknjaz (dismissed):
I have a few minor requests here

src/pylibsshext/session.pyx
Qalthos: I'm not familiar with PEP257 to know if this passes, but this matches the style of the rest of the file.

webknjaz: Unfortunately, most of the linters cannot check Cython so I just check some things myself...
FWIW here's the PEP, it's quite simple https://www.python.org/dev/peps/pep-0257/.
The idea of good docstrings is roughly the same as for the commit messages.
The first line should be a sentence (starting with a verb if it's a function/method) and there should be one empty line if you want to add an extra description. The title line is limited to 72 chars including the quotes. That's about it.

Those para/type things are not from the PEP, they are sphinx-specific and are currently unused but nice to have for humans.

src/pylibsshext/session.pyx
webknjaz: @Qalthos ^

src/pylibsshext/session.pyx
```diff
@@ -219,36 +219,59 @@ cdef class Session(object):
...
-        if kwargs.get('password'):
+        if kwargs.get('password') and supported_auth & libssh.SSH_AUTH_METHOD_PASSWORD:
             try:
                 self.authenticate_password(kwargs["password"])
+            except LibsshSessionException as ex:
+                saved_exception = ex
+            else:
                 return
+
+        # try authenticating with keyboard-interactive (without the interactive)
```
webknjaz: What do you mean by "without the interactive"?

src/pylibsshext/session.pyx
```diff
@@ -383,6 +406,43 @@ cdef class Session(object):
...
+        cdef int rc
+        cdef char should_echo
+        rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
+
+        while rc == libssh.SSH_AUTH_INFO:
+            prompt_count = libssh.ssh_userauth_kbdint_getnprompts(self._libssh_session)
+            if prompt_count > 0:
+                for prompt in range(prompt_count):
+                    prompt_text = libssh.ssh_userauth_kbdint_getprompt(self._libssh_session, prompt, &should_echo)
+                    print(prompt_text)
```
webknjaz: is the print necessary?

src/pylibsshext/session.pyx
```diff
@@ -383,6 +406,43 @@ cdef class Session(object):
...
+
+            # We didn't get SSH_AUTH_SUCCESS this time, so we have to call
+            # userauth_kbdint again until we do (or fail)
+            rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
+
+        if rc in (libssh.SSH_AUTH_ERROR, libssh.SSH_AUTH_DENIED):
+            raise LibsshSessionException("Failed to authenticate with keyboard-interactive: {err}".format(err=self._get_session_error_str()))
+
+        if rc == libssh.SSH_AUTH_SUCCESS:
+            return
```
webknjaz: This looks pointless.
Did you mean
```suggestion
        if rc == libssh.SSH_AUTH_SUCCESS:
            return

        raise LibsshSessionException('Something went wrong')
```

src/pylibsshext/session.pyx
```diff
@@ -383,6 +406,43 @@ cdef class Session(object):
...
+        :raises LibsshSessionException: If authentication failed.
+
+        :return: Nothing.
+        :rtype: NoneType
+        """
+        cdef int rc
+        cdef char should_echo
+        rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
+
+        while rc == libssh.SSH_AUTH_INFO:
```
webknjaz, line 428: Does this need some retries limit/timeout?

webknjaz:
@Qalthos do you want to try adding tests for this?

src/pylibsshext/session.pyx
Qalthos: I suppose "without the keyboard" would be more accurate, just trying to convey that we aren't actually expecting there to be a user interactive element as one might expect.

src/pylibsshext/session.pyx
Qalthos: Holdover from the older version, thanks for spotting

src/pylibsshext/session.pyx
webknjaz: maybe write that down? "emulate a keyboard input without requesting it from the user"

src/pylibsshext/session.pyx
Qalthos: It isn't retries per se, and for a well behaving server I wouldn't think we would need to, but I'll add a check that makes sure we don't loop more than three times.

src/pylibsshext/session.pyx
```diff
@@ -383,6 +408,39 @@ cdef class Session(object):
...
+            prompt_count = libssh.ssh_userauth_kbdint_getnprompts(self._libssh_session)
+            if prompt_count > 0:
+                for prompt in range(prompt_count):
+                    prompt_text = libssh.ssh_userauth_kbdint_getprompt(self._libssh_session, prompt, &should_echo)
+                    if prompt_text == b"Password: ":
```
ganeshrn: The password prompt can be updated to be case insensitive.

src/pylibsshext/session.pyx
```diff
@@ -383,6 +408,39 @@ cdef class Session(object):
...
+        """
+        cdef int rc
+        cdef char should_echo
+        rc = libssh.ssh_userauth_kbdint(self._libssh_session, NULL, NULL)
+
+        while rc == libssh.SSH_AUTH_INFO:
+            prompt_count = libssh.ssh_userauth_kbdint_getnprompts(self._libssh_session)
+            if prompt_count > 0:
+                for prompt in range(prompt_count):
+                    prompt_text = libssh.ssh_userauth_kbdint_getprompt(self._libssh_session, prompt, &should_echo)
```
ganeshrn, line 432: ~If `should_echo` variable is dynamically allocated, it should be freed explicitly to avoid mem leaks~

src/pylibsshext/session.pyx
ganeshrn: Based on the libssh example here https://api.libssh.org/stable/libssh_tutor_authentication.html  for `The keyboard-interactive authentication method` loop if for the total number of prompts.

Qalthos: There's also the outer `while (rc == SSH_AUTH_INFO)` loop. That page is largely what I was working from to understand what's going on so I could put this together.

ganeshrn: Ah..that's right. I missed the outer loop.

src/pylibsshext/session.pyx
```diff
@@ -383,6 +409,39 @@ cdef class Session(object):
...
+                    if re.match(br"Password:\s*", prompt_text, flags=re.IGNORECASE):
```
webknjaz: wouldn't it be better/easier/more efficient to avoid using regex?
```suggestion
                    if b"password:" == prompt_text.lower()[:9]:
```

webknjaz:
@Qalthos mind rebasing?

webknjaz:
@Qalthos this will keep failing until rebased

ganeshrn (approved):
Please add changelog fragments per https://ansible-pylibssh.readthedocs.io/en/latest/contributing/guidelines.html#adding-change-notes-with-your-prs

webknjaz:
@Qalthos FYI this is how you can preview the change log entry you added: https://ansible-pylibssh--107.org.readthedocs.build/en/107/changelog.html#vrelease-unreleased-draft-2020-08-06

webknjaz (approved):

PR #119, opened 2020-08-01
Refactor sphinx RST parsing in towncrier extension

ewjoachim:
##### SUMMARY
Following an [exchange on Twitter](https://twitter.com/Ewjoachim/status/1289323468270395393), I wanted to explore why the `nested_parse_with_titles` Sphinx function (which is [documented](https://www.sphinx-doc.org/en/master/extdev/markupapi.html?highlight=nested_parse_with_titles#parsing-directive-content-as-rest)) wasn't used instead of interacting directly with the state machine.

The answer is "It's longer and not particularly more readable". Still, I'm posting this PR and letting you decide what you want to do with it.

I didn't remember that it was this convoluted. But I managed to make it work 🎉 

Capture d’écran 2020-08-01 à 02 14 34

##### ISSUE TYPE

- Refactor Pull Request :)

##### ADDITIONAL INFORMATION
As usually with docutils, part of the solution came from a [random Google Groups](https://groups.google.com/forum/#!topic/sphinx-dev/IXQdvsFi1qM) answer which I'm linking here for the record.

### DIFF
--- a/docs/_ext/towncrier_draft_ext.py
+++ b/docs/_ext/towncrier_draft_ext.py
@@ -10,7 +10,7 @@ from typing import Any, Dict, List, Union
 
 from sphinx.application import Sphinx
 from sphinx.util.docutils import SphinxDirective
-from sphinx.util.nodes import nodes
+from sphinx.util.nodes import nested_parse_with_titles, nodes
 
 
 # isort: split
@@ -106,6 +106,21 @@ def _get_draft_version_fallback(strategy: str, sphinx_config: Dict[str, Any]):
     return ' '.join(msg_chunks)
 
 
+def _nodes_from_rst(state: statemachine.State, rst: str) -> List[nodes.Node]:
+    """Turn an RST string into a node that can be used in the document."""
+    node = nodes.Element()
+    node.document = state.document
+    nested_parse_with_titles(
+        state=state,
+        content=statemachine.ViewList(
+            rst.splitlines(),
+            source='towncrier-fragments',
+        ),
+        node=node,
+    )
+    return node.children
+
+
 class TowncrierDraftEntriesDirective(SphinxDirective):
     """Definition of the ``towncrier-draft-entries`` directive."""
 
@@ -119,7 +134,6 @@ class TowncrierDraftEntriesDirective(SphinxDirective):
                 f'Error in "{self.name!s}" directive: '
                 'only one argument permitted.',
             )
-
         config = self.state.document.settings.env.config  # noqa: WPS219
         autoversion_mode = config.towncrier_draft_autoversion_mode
         include_empty = config.towncrier_draft_include_empty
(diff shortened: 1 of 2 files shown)


docs/_ext/towncrier_draft_ext.py
```diff
@@ -106,6 +106,21 @@ def _get_draft_version_fallback(strategy: str, sphinx_config: Dict[str, Any]):
...
+    node.document = state.document
+    nested_parse_with_titles(
+        state=state,
+        content=statemachine.ViewList(
+            rst.splitlines(),
+            source='towncrier-fragments',
+        ),
+        node=node,
+    )
+    return node.children
```
webknjaz: @ewjoachim Nice! This is something that should be contributed to Sphinx IMHO :)

docs/_ext/towncrier_draft_ext.py
```diff
@@ -106,6 +106,21 @@ def _get_draft_version_fallback(strategy: str, sphinx_config: Dict[str, Any]):
     return ' '.join(msg_chunks)
 
 
+def _nodes_from_rst(state: statemachine.State, rst: str) -> nodes.Element:
```
webknjaz: Doesn't it return `List[nodes.Node]`? `node.children` is a list
```suggestion
def _nodes_from_rst(state: statemachine.State, rst: str) -> List[nodes.Node]:
```

ewjoachim: Ah yes you're right !

webknjaz (approved):

webknjaz:
@ewjoachim could you please re-submit this to https://github.com/sphinx-contrib/towncrier too? I cannot do this myself as it has a different license.

ewjoachim:
I hereby grant you full rights to re-license any of my pylibssh contributions to any BSD-3-licensed project including sphinx-contrib/towncrier :)

(I'll do it at some point, but if you don't want to wait for me to do it, now I guess you can, but IANAL)

docs/_ext/towncrier_draft_ext.py
```diff
@@ -106,6 +106,21 @@ def _get_draft_version_fallback(strategy: str, sphinx_config: Dict[str, Any]):
...
 
 
+def _nodes_from_rst(state: statemachine.State, rst: str) -> List[nodes.Node]:
+    """Turn an RST string into a node that can be used in the document."""
+    node = nodes.Element()
+    node.document = state.document
+    nested_parse_with_titles(
+        state=state,
+        content=statemachine.ViewList(
+            rst.splitlines(),
```
webknjaz, line 116: Is this better than `statemachine.string2lines()`?

webknjaz:
@ewjoachim I'll be happy to wait :) I'm delegating the work on that project to my mentee for the sake of learning and will be trying to avoid doing things instead of him. So I expect that it'll take some time anyway.

docs/_ext/towncrier_draft_ext.py
ewjoachim: no idea.

ewjoachim: http://code.nabla.net/doc/docutils/api/docutils/statemachine/docutils.statemachine.string2lines.html

PR #122, opened 2020-08-04
[WIP]Update close method to cpdef

ganeshrn:
##### SUMMARY

Define ``close()`` as ``cpdef``

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -159,16 +159,11 @@ cdef class Channel:
         callbacks.ssh_callbacks_init(&cb)
         callbacks.ssh_set_channel_callbacks(channel, &cb)
 
-        libssh.ssh_channel_send_eof(channel)
-        result.returncode = libssh.ssh_channel_get_exit_status(channel)
         if channel is not NULL:
             libssh.ssh_channel_close(channel)
             libssh.ssh_channel_free(channel)
 
         return result
-
-    def get_channel_exit_status(self):
-        return libssh.ssh_channel_get_exit_status(self._libssh_channel)
 
     def close(self):
         if self._libssh_channel is not NULL:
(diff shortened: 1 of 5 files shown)


src/pylibsshext/session.pyx
```diff
@@ -101,20 +101,15 @@ cdef class Session(object):
...
             self.set_ssh_options(key, kwargs[key])
 
-    def __dealloc__(self):
-        if self._libssh_session is not NULL:
-            if libssh.ssh_is_connected(self._libssh_session):
-                libssh.ssh_disconnect(self._libssh_session)
-            libssh.ssh_free(self._libssh_session)
-            self._libssh_session = NULL
+    def __del__(self):
+        self.close()
```
webknjaz, line 119: Why do you think it's better to use `__del__` in Python-land?

src/pylibsshext/channel.pyx
```diff
@@ -46,7 +46,7 @@ cdef int _process_outputs(libssh.ssh_session session,
     return len
 
 cdef class Channel:
-    def __cinit__(self, session):
```
webknjaz, line 48: why?

webknjaz:
@ganeshrn what is it that you're trying to fix here?

webknjaz:
@ganeshrn is this still relevant?

webknjaz:
@Jakuje would you mind looking into this patch? I'm not able to figure out the motivation behind it, but perhaps you could?

Jakuje:
> @Jakuje would you mind looking into this patch? I'm not able to figure out the motivation behind it, but perhaps you could?

Sorry, I have no idea. It looks like just moving things around without good explanation. Given that we did not hear from the reporter for 4 years, I would not bother.

webknjaz:
Thanks for confirming!

Issue #150, opened 2020-10-04
[TODO] Replace `set-env` and `add-path` with env files in GHA


webknjaz:
These are deprecated and will need to be replaced ASAP. Here's why: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

fmsabatini:
Any news on this? the deadline is close, and we rely on Manubot ;-)

PR #158, opened 2020-12-03
Fix linking to libssh.so

### LINKED ISSUE #63: [TODO] Enable testing sdists under Ubuntu in GHA

### LINKED ISSUE #103: Installs from sdists produce broken C-extensions
##### SUMMARY

It's really low-priority right now (because it's unlikely that anybody would hit it) but installs from sdist under Ubuntu produce C-extensions that explode with missing symbols:
```console
E   ImportError: /home/runner/work/pylibssh/pylibssh/.tox/test-source-dists/lib/python2.7/site-packages/pylibsshext/session.so: undefined symbol: ssh_disconnect
```

##### ISSUE TYPE
- Bug Report

##### OS / ENVIRONMENT

https://github.com/ansible/pylibssh/pull/102/checks?check_run_id=879173917#step:12:178

Qalthos:
##### SUMMARY

Fixes #103 and resolves #63.

The issue seems to be that the LDFLAGS env is not picked up or used
properly in some instances. Explicitly listing the link in the pxd files
causes the files to be built with `-lssh` properly.

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -100,5 +100,9 @@ pytest-parametrize-names-type = tuple
 pytest-parametrize-values-type = tuple
 pytest-parametrize-values-row-type = tuple
 
+# flake8-eradicate
+# E800:
+eradicate-whitelist-extend = distutils:\s+libraries\s+=\s+
+
 # wemake-python-styleguide
 show-source = true
--- a/.github/workflows/build-test-n-publish.yml
+++ b/.github/workflows/build-test-n-publish.yml
@@ -589,10 +589,6 @@ jobs:
         exclude:
         - dist-type: source
           runner-vm-os: ubuntu-16.04  # has too old libssh in the repos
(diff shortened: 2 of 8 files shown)


webknjaz:
I've fixed the CI on devel (it was broken for a while because of #150) and will now rebase this branch.

webknjaz:
@Qalthos Weird... Does https://github.com/ansible/pylibssh/blob/bc911cc/pyproject.toml#L26 not help with this?

webknjaz (approved):
Honestly, I was thinking that it's related to something like pygit2 troubleshooting describes at  https://www.pygit2.org/install.html#the-run-path. But Let's merge it and see if that will fix our problem long-term.

I'll add a linting patch before merging because it now reports false-positives in flake8.

PR #216, opened 2021-05-14
Report sftp_error when write fails

Qalthos:
##### SUMMARY

Add additional error reporting to SFTP write errors

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/docs/changelog-fragments/216.bugfix.2.rst
+++ b/docs/changelog-fragments/216.bugfix.2.rst
@@ -1,0 +1,1 @@
+Changed ``sftp.sftp_get`` to write files as bytes rather than assuming files are valid UTF8 -- by :user:`Qalthos`
--- a/docs/changelog-fragments/216.bugfix.2.rst
+++ b/docs/changelog-fragments/216.bugfix.2.rst
@@ -1,1 +1,0 @@
-Changed ``sftp.sftp_get`` to write files as bytes rather than assuming files are valid UTF8 -- by :user:`Qalthos`
--- a/docs/changelog-fragments/216.bugfix.rst
+++ b/docs/changelog-fragments/216.bugfix.rst
@@ -1,0 +1,1 @@
+Added additional error reporting to SFTP write errors -- :user: Qalthos
--- a/docs/changelog-fragments/216.bugfix.rst
+++ b/docs/changelog-fragments/216.bugfix.rst
@@ -1,1 +1,1 @@
-Added additional error reporting to SFTP write errors -- :user: Qalthos
+Added additional error reporting to SFTP write errors -- by :user:`Qalthos`
--- a/docs/changelog-fragments/216.bugfix.rst
+++ b/docs/changelog-fragments/216.bugfix.rst
@@ -1,1 +1,1 @@
-Added additional error reporting to SFTP write errors -- by :user:`Qalthos`
+Changed ``sftp.sftp_get`` to write files as bytes rather than assuming files are valid UTF8 -- by :user:`Qalthos`
--- a/docs/changelog-fragments/216.misc.rst
+++ b/docs/changelog-fragments/216.misc.rst
@@ -1,0 +1,1 @@
+Added additional details when SFTP write errors are raised -- by :user:`Qalthos`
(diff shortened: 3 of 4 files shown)


webknjaz:
@Qalthos this needs a changelog fragment and maybe a test? Oh and please fill out the summary.

ganeshrn (approved):

webknjaz:
@Qalthos is https://github.com/ansible/pylibssh/issues/222 related?

webknjaz:
@Qalthos plz rebase to pull in the CI+linter+docs build fixes from devel.

docs/changelog-fragments/216.bugfix.2.rst
```diff
@@ -0,0 +1 @@
+Changed ``sftp.sftp_get`` to write files as bytes rather than assuming files are valid UTF8 -- by :user: `Qalthos`
```
webknjaz: @Qalthos FYI this is a syntax error in RST :point_up: — there shouldn't be a space between the role `:user:` and its argument for Sphinx to parse that as one thing. Otherwise, it tries to interpret them separately.

This caused the failures here:
* https://github.com/ansible/pylibssh/pull/216/checks?check_run_id=2806860150
* https://readthedocs.org/projects/ansible-pylibssh/builds/13990866/

```suggestion
Changed ``sftp.sftp_get`` to write files as bytes rather than assuming files are valid UTF8 -- by :user:`Qalthos`
```

docs/changelog-fragments/216.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added additional error reporting to SFTP write errors -- by :user: `Qalthos`
```
webknjaz:
```suggestion
Added additional error reporting to SFTP write errors -- by :user:`Qalthos`
```

docs/changelog-fragments/216.bugfix.rst
```diff
@@ -0,0 +1 @@
+Added additional error reporting to SFTP write errors -- by :user:`Qalthos`
```
webknjaz: Is this one really a bugfix? Do you want to use a fragment number here too, to control ordering in the changelog?

webknjaz: The order is visible here: https://ansible-pylibssh--216.org.readthedocs.build/en/216/changelog/#vrelease-unreleased-draft-2021-06-16

webknjaz:
@Qalthos this needs rebasing

.pre-commit-config.yaml
```diff
@@ -57,7 +57,7 @@ repos:
       ^docs/_samples/.*\.py$
 
 - repo: https://gitlab.com/pycqa/flake8.git
-  rev: 3.8.2
+  rev: 3.9.2
```
webknjaz: Why did you change it in this PR?

bin/pep517_backend/_backend.py
```diff
@@ -76,15 +76,12 @@ def get_config():
 
         [tool.local.cythonize.kwargs.directives]
         # This section can contain compiler directives
-        # NAME = "VALUE"
```
webknjaz: This should remain as it's the only example of how things are supposed to be configured right now. You could either uncomment it or extend the exclusion at https://github.com/ansible/pylibssh/blob/820a2d0/.flake8#L86.

.pre-commit-config.yaml
Qalthos: Oh you're right, I was trying things out and did not undo that

bin/pep517_backend/_backend.py
Qalthos: Ah, so that is how that is configured. That seems much more sensible.

webknjaz: Yeah, I also think that it's a false-positive because it's a comment inside a docstring. Probably need to report this to `eradicate`.

webknjaz: Looks like it's https://github.com/wemake-services/flake8-eradicate/issues/215.

webknjaz:
FYI don't worry about the Fedora:34 job, it's not related to this PR.

.flake8
```diff
@@ -83,7 +83,7 @@ pytest-parametrize-values-row-type = tuple
 
 # flake8-eradicate
 # E800:
-eradicate-whitelist-extend = distutils:\s+libraries\s+=\s+|isort:\s+\w+
+eradicate-whitelist-extend = distutils:\s+libraries\s+=\s+|isort:\s+\w+|NAME = "VALUE"
```
webknjaz: Maybe you could submit this separately with its own misc change note?

Qalthos: Yeah, I'm doing that now

PR #277, opened 2021-12-07
Start building for 3.10

Qalthos:
##### SUMMARY

Notable changes:
* Loss of 2.7 and 3.5 on macOS. This could probably be restored with enough effort
* No longer building centos8 RPM.
  * Still building centos-stream 8 - this probably isn't an issue?

##### ISSUE TYPE

- Feature Pull Request?

##### ADDITIONAL INFORMATION

```paste below

```


webknjaz:
@Qalthos you need to update the container images first, I think.

PR #323, opened 2022-05-02
Simplfy matrix.include

Qalthos:
##### SUMMARY

This should still result in 49 build jobs

##### ISSUE TYPE

- CI Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


Qalthos:
I don't think this actually works this way

PR #343, opened 2022-07-01
Support specifying socket fd for connection.

sabedevops:
##### SUMMARY

The libssh library advertises "[Your sockets: the app hands over the socket, or uses libssh sockets](https://www.libssh.org/features/)".

The pylibssh library should support this capability. With respect to Ansible, it opens the door for new connections plugins to seamlessly operate over different sockets  (like Unix sockets, or  WebSockets) while allowing for code reuse.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```python
import socket

sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
s = Session()
s.connect(fd=sock.fileno())
```

### DIFF
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -26,6 +26,7 @@ from pylibsshext.sftp import SFTP
 
 
 OPTS_MAP = {
+    "fd": libssh.SSH_OPTIONS_FD,
     "host": libssh.SSH_OPTIONS_HOST,
     "user": libssh.SSH_OPTIONS_USER,
     "port": libssh.SSH_OPTIONS_PORT,
@@ -147,6 +148,7 @@ cdef class Session(object):
             return ret
 
     def set_ssh_options(self, key, value):
+        cdef int value_fd
         cdef int value_int
         cdef unsigned int value_uint
         cdef long value_long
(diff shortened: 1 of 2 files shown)


sabedevops:
Is there anything I can do to improve this PR?

sabedevops:
Is there anything I can do that would help get this PR accepted?

src/pylibsshext/session.pyx
```diff
@@ -147,6 +148,7 @@ cdef class Session(object):
             return ret
 
     def set_ssh_options(self, key, value):
+        cdef int value_fd
```
Qalthos: We don't actually need this, you can just reuse `value_int`. These are just here to stuff `value` in to the correctly typed C variable so that libssh's `ssh_options_set` knows what to do with them.
```suggestion
```

src/pylibsshext/session.pyx
```diff
@@ -158,7 +160,10 @@ cdef class Session(object):
             key_m = OPTS_MAP[key]
         else:
             raise LibsshSessionException("Unknown attribute name [%s]" % key)
-        if key == "gssapi_delegate_credentials":
+        if key == "fd":
+            value_fd = value
+            libssh.ssh_options_set(self._libssh_session, key_m, &value_fd)
+        elif key == "gssapi_delegate_credentials":
```
Qalthos: Likewise, we're not doing anything in particular in this method other than getting the Python input to match libssh's conventions, so there isn't any reason to keep these apart.
```suggestion
        if key in ("fd", "gssapi_delegate_credentials"):
```

Qalthos (changes requested):

src/pylibsshext/session.pyx
sabedevops: Thanks for the suggestions @Qalthos. I didn't know that applying the suggestions would auto-resolve the conversation here. I appreciate you taking the time to review this change.

Qalthos (approved):

PR #472, opened 2023-01-04
Third party libraries are located under /usr/local in FreeBSD

donnerhacke:
##### SUMMARY
The filesystem hierarchy at FreeBSD requires non-essential libraries to be installed in the `/usr/local` tree. Hence the compiler needs to search `/usr/local/include` for the required header file `libssh/libssh.h`.

The project does not compile without specifying this path.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Before
```
cc -pthread -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -DCYTHON_TRACE=1 -DCYTHON_TRACE_NOGIL=1 -fPIC -I/tmp/build-env-w5pvtdrg/include -I/usr/local/include/python3.9 -c /tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.c -o /tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/tmpnmc2mh29/tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.o
/tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.c:757:10: fatal error: 
      'libssh/libssh.h' file not found
#include "libssh/libssh.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.
error: command '/usr/bin/cc' failed with exit code 1
```

After
```
24 warnings generated.
cc -pthread -shared -L/usr/local/lib -fstack-protector-strong -lssh -I /usr/local/include -DCYTHON_TRACE=1 -DCYTHON_TRACE_NOGIL=1 /tmp/build-via-sdist-7zkzyplu/ansible-pylibssh-0.1.dev1359+g7e61df1/src/tmpp0lv82d8/tmp/build-via-sdist-7zkzyplu/ansible-pylibssh-0.1.dev1359+g7e61df1/src/pylibsshext/scp.o -L/usr/local/lib -lssh -o build/lib.freebsd-13.1-STABLE-amd64-3.9/pylibsshext/scp.cpython-39.so
```

PR #477, opened 2023-01-04
Third party libraries are located under /usr/local in FreeBSD

donnerhacke:
##### SUMMARY
The filesystem hierarchy at FreeBSD requires non-essential libraries to be installed in the `/usr/local` tree. Hence the compiler needs to search `/usr/local/include` for the required header file `libssh/libssh.h`.

The project does not compile without specifying this path.

Supersedes #472 (wrong branch) 

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Before
```
cc -pthread -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -O2 -pipe -fstack-protector-strong -fno-strict-aliasing -DCYTHON_TRACE=1 -DCYTHON_TRACE_NOGIL=1 -fPIC -I/tmp/build-env-w5pvtdrg/include -I/usr/local/include/python3.9 -c /tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.c -o /tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/tmpnmc2mh29/tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.o
/tmp/build-via-sdist-6bpb1mwc/ansible-pylibssh-0.1.dev1358+g11b206a/src/pylibsshext/_libssh_version.c:757:10: fatal error: 
      'libssh/libssh.h' file not found
#include "libssh/libssh.h"
         ^~~~~~~~~~~~~~~~~
1 error generated.
error: command '/usr/bin/cc' failed with exit code 1
```

After
```
24 warnings generated.
cc -pthread -shared -L/usr/local/lib -fstack-protector-strong -lssh -I /usr/local/include -DCYTHON_TRACE=1 -DCYTHON_TRACE_NOGIL=1 /tmp/build-via-sdist-7zkzyplu/ansible-pylibssh-0.1.dev1359+g7e61df1/src/tmpp0lv82d8/tmp/build-via-sdist-7zkzyplu/ansible-pylibssh-0.1.dev1359+g7e61df1/src/pylibsshext/scp.o -L/usr/local/lib -lssh -o build/lib.freebsd-13.1-STABLE-amd64-3.9/pylibsshext/scp.cpython-39.so
```

### DIFF
--- a/docs/changelog-fragments/477.bugfix.rst
+++ b/docs/changelog-fragments/477.bugfix.rst
@@ -1,2 +1,2 @@
-Extended CFLAGS to include header locations typical to FreeBSD
+Extended ``CFLAGS`` to include header locations typical to FreeBSD
 (:file:`/usr/local/include`) -- by :user:`donnerhacke`
--- a/docs/changelog-fragments/477.bugfix.rst
+++ b/docs/changelog-fragments/477.bugfix.rst
@@ -1,2 +1,2 @@
-- Extended CFLAGS to include header locations typical to FreeBSD
-  (:file:/usr/local/include) -- by :user:`donnerhacke`
+Extended CFLAGS to include header locations typical to FreeBSD
+(:file:/usr/local/include) -- by :user:`donnerhacke`
--- a/docs/changelog-fragments/477.bugfix.rst
+++ b/docs/changelog-fragments/477.bugfix.rst
@@ -1,2 +1,2 @@
 Extended CFLAGS to include header locations typical to FreeBSD
-(:file:/usr/local/include) -- by :user:`donnerhacke`
+(:file:`/usr/local/include`) -- by :user:`donnerhacke`
--- a/docs/changelog-fragments/477.bugfix.rst
+++ b/docs/changelog-fragments/477.bugfix.rst
@@ -1,0 +1,2 @@
+- Extended CFLAGS to include header locations typical to FreeBSD
+  (:file:/usr/local/include) -- by :user:`donnerhacke`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ src = ["src/**/*.pyx"]
 #ANSIBLE_PYLIBSSH_TRACING = "1"
 #CFLAGS = "-DCYTHON_TRACE=1 ${CFLAGS}"
 LDFLAGS = "-lssh ${LDFLAGS}"
+CFLAGS="-I /usr/local/include ${CFLAGS}"
 
 [tool.local.cythonize.flags]
 # This section can contain the following booleans:


docs/changelog-fragments/477.bugfix.rst
```diff
@@ -0,0 +1 @@
+- Fix missing includes under FreeBSD -- by :user:`donnerhacke`
```
webknjaz: It's already dashed in Towncrier's Jinja2 template + it makes sense to use the past tense with more details here:
```suggestion
Extended CFLAGS to include header locations typical to FreeBSD
(:file:`/usr/local/include`) -- by :user:`donnerhacke`
```

pyproject.toml
```diff
@@ -22,6 +22,7 @@ src = ["src/**/*.pyx"]
 #ANSIBLE_PYLIBSSH_TRACING = "1"
 #CFLAGS = "-DCYTHON_TRACE=1 ${CFLAGS}"
 LDFLAGS = "-lssh ${LDFLAGS}"
+CFLAGS="-I /usr/local/include ${CFLAGS}"
```
webknjaz, line 25: I'm wondering this should be after the existing flags and not before :thinking:
Thoughts @Qalthos?

donnerhacke: I'm sure, you know better, where to include this setting.
Maybe there is a platform depended possibility?

docs/changelog-fragments/477.bugfix.rst
donnerhacke: Thank you for the idea. Took it.

pyproject.toml
donnerhacke: I've tried to find a more or less simple way to define the compile environment for a single platform (i.e. freebsd), but this quickly becomes complex:

- `CFLAGS` are overwritten in `tox.ini`, so a pass-through from the calling environment would require changes there.
- Another possibility is to define an `envlist` for the `tox.ini` target, which would multiply the size of `tox.ini`. 
- Providing platform based `extra_compile_args` in `[build_env]` sections of the `pyproject.toml` simply does not work (due to the more specific cythonize definitions.

Hence I'd would keep the proposed solution. `/usr/local/include` is not that uncommon, so that other distributions would profit, while the current distributions are not harmed.

docs/changelog-fragments/477.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Extended CFLAGS to include header locations typical to FreeBSD
+(:file:/usr/local/include) -- by :user:`donnerhacke`
```
webknjaz:
```suggestion
(:file:`/usr/local/include`) -- by :user:`donnerhacke`
```

docs/changelog-fragments/477.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Extended CFLAGS to include header locations typical to FreeBSD
```
webknjaz:
```suggestion
Extended ``CFLAGS`` to include header locations typical to FreeBSD
```

docs/changelog-fragments/477.bugfix.rst
webknjaz: (missing such bits of syntax is exactly why it's best to accept the suggestions from GH UI)

pyproject.toml
webknjaz: FWIW this is designed to allow passing though env vars from the calling shell. So yes, changing the `passenv` in `tox.ini` is reasonable in this case. If your system has env vars set properly already, tox shouldn't isolate from that. Although, if your use-case is just building the dists, you probably shouldn't use tox and do so via `pip install`/`pip wheel` or `python -m build`.

Anyway, I'll let @Qalthos decide on this, I think. I don't have a strong preference.

webknjaz: Thinking about this more, maybe we shouldn't get into the business of hardcoding values that aren't directly related to our packaging efforts... Not on the native Python packaging level — the ability to pass those through seems like a good option. @donnerhacke perhaps try moving this to `tox`? Just `passenv` or `setenv` seem better suited.

webknjaz: @donnerhacke so do you think using `passenv` would be helpful?

donnerhacke:
I'm fine with a passthrough of environment variables.
So, the request can be closed without merging.

PR #479, opened 2023-01-04
Allow processing of a specific configuration file for the connection

donnerhacke:
##### SUMMARY

Depending on the destination, libssh might need to be configured in a way, that the algorithms and methods need to be adjusted. Instead of adding each possible configuration option into the wrapper, including a config file (as it exists for ordinary OpenSSH) is much easier. This allows ansible to connect to devices with less well supported algorithms.

This is a required step to fix connection issues with ansible.netcommon.libssh.  The calling ansible.netcommon.libssh need to provide the new optional argument config_file.  But this is a separate patch, which needs to be applied there.

##### ISSUE TYPE
- Feature Pull Request


donnerhacke:
Link to the patch on the other side: https://github.com/ansible-collections/ansible.netcommon/pull/498

docs/changelog-fragments/479.feature.rst
```diff
@@ -0,0 +1 @@
+Allowed processing of a specific configuration file for the connection -- by :user:`donnerhacke`
```
webknjaz: I'd rephrase this through something "Implemented reading connection details from a config file"

webknjaz:
```suggestion
Implemented reading connection details from a config file -- by :user:`donnerhacke`.
```

webknjaz:
@donnerhacke it would be really helpful to have test for this. Currently, there's no coverage for the new lines, which is... suboptimal.

src/pylibsshext/session.pyx
```diff
@@ -243,6 +243,16 @@ cdef class Session(object):
             if (key in OPTS_MAP or key in OPTS_DIR_MAP) and (kwargs[key] is not None):
                 self.set_ssh_options(key, kwargs[key])
 
+        if 'config_file' in kwargs:
+            file_name = kwargs['config_file']
+            rc = libssh.ssh_options_parse_config(self._libssh_session, file_name.encode())
+            if rc != libssh.SSH_OK or self._get_session_error_str() != "":
+                libssh.ssh_disconnect(self._libssh_session)
```
Jakuje, line 268: I think the disconnect here is not needed as we did not connect yet.

src/pylibsshext/session.pyx
```diff
@@ -243,6 +243,16 @@ cdef class Session(object):
             if (key in OPTS_MAP or key in OPTS_DIR_MAP) and (kwargs[key] is not None):
                 self.set_ssh_options(key, kwargs[key])
 
+        if 'config_file' in kwargs:
```
Jakuje, line 264: This new key needs to be documented in the doc text above with the description of how the corner cases are handled.

The libssh itself reads the OpenSSH configuration files by default and the openssh configuration files have awkward precedence, so this documentation needs to clarify if this configuration file is parsed before the openssh ones, after them and if the same option is specified in both config and direct option, which one is effective.

@donnerhacke Please, let me know if you plan to get back to make this working or you abandoned this PR in favor of different solution (such as using the OpenSSH config directly).

alehaa:
Is there some way to assist in this PR? Would love to see this merged.

Elektordi:
This feature is already shown in the ansible documentation as "added in ansible.netcommon 5.1.0" but currently not working without any warning due to this PR not being merged... :(
@donnerhacke do you need any help to finish it?

PR #502, opened 2023-05-11
🎨Only use setuptools-scm-git-archive under < py37

webknjaz:
`setuptools-scm` supports Git archives natively starting v7 but it only supports Python 3.7 or higher. This patch makes up a set of conditional build dependencies that add `setuptools-scm-git-archive` only below Python 3.7 and in combination with an older version of `setuptools-scm`.

##### ISSUE TYPE

- Bugfix Pull Request
- Maintenance Pull Request
- Packaging Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1161,6 +1161,7 @@ jobs:
   build-src:
     name: >-
       👷 an sdist 📦 ${{ needs.pre-setup.outputs.git-tag }}
+      under ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


serhii73 (approved):

ssbarnea (approved):

webknjaz:
FTR the centos stream 8 job in CI took 49m 24s, instead of the usual ~2min. Looks like there's currently some network issues. Not sure if it's within the GitHub platform or on the distro infra side.

PR #507, opened 2023-06-15
I think this should fix packit?

Qalthos:
##### SUMMARY

see https://github.com/packit/packit/issues/1821

##### ISSUE TYPE

- Bugfix Pull Request

### DIFF
--- a/packaging/rpm/ansible-pylibssh.spec
+++ b/packaging/rpm/ansible-pylibssh.spec
@@ -223,7 +223,6 @@ export PYTHONPATH="%{buildroot_site_packages}:${PYTHONPATH}"
 # Fedora:
 %if "%{?fedora:SET}" == "SET"
 %tox -e just-pytest -- \
-  -vv \
   --installpkg '%{_builddir}/%{pypi_name}-%{upstream_version}/pyproject-wheeldir/%{whl_glob}' \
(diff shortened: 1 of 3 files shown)


packaging/rpm/ansible-pylibssh.spec
```diff
@@ -223,8 +223,6 @@ export PYTHONPATH="%{buildroot_site_packages}:${PYTHONPATH}"
 # Fedora:
 %if "%{?fedora:SET}" == "SET"
 %tox -e just-pytest -- \
-  -vv \
-  --installpkg '%{_builddir}/%{pypi_name}-%{upstream_version}/pyproject-wheeldir/%{whl_glob}' \
```
webknjaz, line 227: @Qalthos why? This would cause a separate wheel build which means that the tests would run against something else..

packaging/rpm/ansible-pylibssh.spec
```diff
@@ -184,7 +184,7 @@ PYTHONPATH="$(pwd)/bin" \
 
 # Fedora:
 %if 0%{?fedora}
-%pyproject_wheel %{python_importable_name}
```
webknjaz, line 187: What is the effect of this?

PR #533, opened 2023-11-07
Add manylinux_2_28 support

Qalthos:
##### SUMMARY

Start the process of supporting 3.12

~~Also drop 3.6/3.7 support as it complicates the setuptools change~~

##### ISSUE TYPE

- Feature Pull Request


build-scripts/manylinux-container-image/Dockerfile
```diff
@@ -5,8 +5,8 @@ ARG LIBSSH_VERSION=0.9.6
 MAINTAINER Python Cryptographic Authority
 WORKDIR /root
 
-ADD update-packages.sh /root/update-packages.sh
-RUN ./update-packages.sh
+# ADD update-packages.sh /root/update-packages.sh
```
webknjaz: why?

docs/changelog-fragments/533.misc.rst
```diff
@@ -0,0 +1 @@
+Start building manylinux_2_28 base images -- by :user:`Qalthos`
```
webknjaz:
```suggestion
Started building ``manylinux_2_28`` base images -- by :user:`Qalthos`
```

.github/workflows/ci-cd.yml
```diff
@@ -1318,21 +1292,27 @@ jobs:
...
-          registry: registry.access.redhat.com
-        - tag: ubi8/ubi:8.4
-          registry: registry.access.redhat.com
         - tag: ubi8/ubi:8.5
           registry: registry.access.redhat.com
         - tag: ubi8/ubi:8.6
           registry: registry.access.redhat.com
+        - tag: ubi8/ubi:8.7
+          registry: registry.access.redhat.com
+        - tag: ubi8/ubi:8.8
```
webknjaz, line 1335: Looks like UBI 8 only ships with Python 3.6:
https://github.com/ansible/pylibssh/actions/runs/6788921335/job/18455029172?pr=533#step:20:622

Qalthos: Yeah, that does seem to be a problem. I can try to walk back the Python version raise or drop ubi8, not sure if there are other options available.

build-scripts/manylinux-container-image/Dockerfile
Qalthos: This was me trying to work around the _2_28 images being AlmaLinux which doesn't have `prelink` available to install, and also the debian repos for the older images having been moved to archive (and thus 404ing without more changes).

I wanted to see if we could get away without calling out to the package manager at all, and it looks like we can, so the next step to me seems to be removing the script entirely. I just didn't want to go through all of that and then have to fish it back out of the history if the images depended further on the package manager.

.github/workflows/ci-cd.yml
Qalthos: Okay, we can punt the 3.6 support issue for now. I didn't really want that in this PR anyway.

Some discussion of what we are testing here and how many versions to keep might not be amiss at this point.

build-scripts/manylinux-container-image/Dockerfile
webknjaz: Historically, the manylinux image generation has been `git merge`d from the upstream, which is why I was aiming at keeping the differences minimal, so there are less conflicts on sync.

.github/workflows/build-manylinux-container-images.yml
```diff
@@ -44,6 +44,7 @@ jobs:
         YEAR:
         - 2014
         - _2_24  # PEP 600
+        - _2_28
```
webknjaz, line 47: Looks like the same tag is needed in the CI/CD workflow. This one only builds images, not the project itself.

webknjaz: Ah, nevermind. #532 does that.

PR #534, opened 2023-11-08
Drop Python 3.6/3.7 support

Qalthos:
##### SUMMARY

This presently breaks ubi8

##### ISSUE TYPE

- Breaking Change Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/docs/changelog-fragments/532.breaking.rst
+++ b/docs/changelog-fragments/532.breaking.rst
@@ -1,0 +1,2 @@
+Dropped support for Python 3.6 and Python 3.7
+-- by :user:`Qalthos` and :user:`webknjaz`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,5 @@
 # pylint: disable=invalid-name
-# Requires Python 3.6+
+# Requires Python 3.8+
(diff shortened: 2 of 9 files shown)


webknjaz:
@Qalthos let's merge this after making a release. That way, there would be one last release with the wheels for those old Python runtimes. It doesn't cost us anything to click the merge button a bit later. The `devel` is green already.

docs/conf.py
```diff
@@ -1,5 +1,5 @@
 # pylint: disable=invalid-name
-# Requires Python 3.6+
+# Requires Python 3.8+
```
webknjaz, line 2: Technically, this patch doesn't upgrade anything in the Sphinx setup. So I'd leave it as is. OTOH, RTD is running Python 3.10. And GHA uses Python 3.11.

docs/changelog-fragments/532.breaking.rst
```diff
@@ -0,0 +1 @@
+Dropped support for Python 3.6/3.7 -- by :user:`Qalthos`
```
webknjaz:
```suggestion
Dropped support for Python 3.6/3.7 -- by :user:`Qalthos`.
```

PR #501, opened 2023-05-10
🎨 Migrate PEP 517 backend to `tomllib`+`tomli`

webknjaz:
`tomllib` is a part of stdlib since Python 3.11 and `tomli` is used as a fallback for the older Python versions. The latter is API-compatible with the former.

Fixes #504.

##### SUMMARY

$sbj.

##### ISSUE TYPE

- Maintenance Pull Request
- Packaging Pull Request

##### ADDITIONAL INFORMATION

N/A

PR #546, opened 2023-12-13
Fix rpm building

Qalthos:
##### SUMMARY

It probably isn't this easy, but let's try it anyway

##### ISSUE TYPE

- Misc Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1335,14 +1335,14 @@ jobs:
     strategy:
       matrix:
         target-container:
-        - tag: fedora:37
         - tag: fedora:38
         - tag: fedora:39
+        - tag: fedora:40
         - tag: centos/centos:stream8
           registry: quay.io
         # No matching package to install: 'python3dist(wheel)'
-        # - tag: centos/centos:stream9
-        #   registry: quay.io
+        - tag: centos/centos:stream9
+          registry: quay.io
         - tag: ubi8/ubi:8.5
           registry: registry.access.redhat.com
         - tag: ubi8/ubi:8.6
@@ -1377,12 +1377,13 @@ jobs:
           matrix.target-container.tag
         }}
 
-    continue-on-error: >-
-      ${{
-        contains(matrix.target-container.tag, 'ubi9')
-        && true
-        || false
-      }}
+    continue-on-error: true
+    # >-
+    # ${{
+    #   contains(matrix.target-container.tag, 'ubi9')
+    #   && true
+    #   || false
+    # }}
 
     steps:
     - name: Produce artifact name
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1343,13 +1343,13 @@ jobs:
         # No matching package to install: 'python3dist(wheel)'
         - tag: centos/centos:stream9
           registry: quay.io
-        - tag: ubi8/ubi:8.5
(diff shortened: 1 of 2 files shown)


webknjaz:
@Qalthos there's `sed -i` removing `expandvars` from `pyproject.toml` on Fedora specifically. We should probably check if that's still necessary and could be dropped, maybe.

webknjaz:
@Qalthos also, check out the new Towncrier category types: https://github.com/ansible/pylibssh/blob/b210a2df2f858ed24e0bc331fa14af5ea8ca4ed2/pyproject.toml#L91-L141 — they are supposed to make the way we track release notes more granular. I've been testing these categories in @aio-libs and they work great!

.github/workflows/ci-cd.yml
```diff
@@ -1377,12 +1377,13 @@ jobs:
           matrix.target-container.tag
         }}
 
-    continue-on-error: >-
-      ${{
-        contains(matrix.target-container.tag, 'ubi9')
-        && true
-        || false
-      }}
+    continue-on-error: true
```
webknjaz, line 1380: With this, failures won't contribute to the gating. Make sure to re-enable the important ones before merging.

webknjaz:
I've taken a look into the UBI 9.3 and I'm a bit confused. The current spec file installs a bunch of hardcoded lib versions from the PyPI that were meant to be used in distro versions that lack them in the RPM repos. But that was long ago and there's a chance some part of them isn't needed on any of the tested envs.

For RHEL, there used to be no `pyproject-rpm-macros` so one of the dists from PyPI was pypa/build of a version that supports in-tree build backends with its dependency `pep517`. Both have quite old versions as of today. Over time, the upstream `build` project replaced `pep517` with `pyproject_hooks` (which is mostly the same library with more stable APIs and the CLI bits removed).
And under this env, we also called `python -m build` manually in the spec file that used to work, especially since we injected a bunch of PyPI dists into the `bin/` dir and put that into `PYTHONPATH`.

And what puzzles me is that it used to work, plus the dists are pinned to the same versions they were in the past.

I experimented with what we have in a local container and removing `--no-isolation` does make the command work. Though, I still don't see why `build` behaves like this — it's supposed to inject the `packaging/` directory into `sys.path` / `PYTHONPATH` internally, when it sees that `pyproject.toml` declares an in-tree build backend. And now I observe that it doesn't do this with no isolation. I don't see any other libs called `build` that could leak into the imports.

FWIW, I hope that upgrading to using `pyproject-rpm-macros` in those envs will magically solve things. I verified interactively that the following upgrade does cause it to inject the in-tree build backend into the import path properly:
```console
[root@44fca2cd2e2c ansible-pylibssh-1.1.1.dev100+g2897acd.d20240103]# /usr/bin/python3 -m pip install -t bin/ 'build > 0.4'
...
Successfully installed build-1.0.3 importlib-metadata-7.0.1 packaging-23.2 pyproject-hooks-1.0.0 tomli-2.0.1 zipp-3.17.0
```
And the build command doesn't error out with a missing backend with that.

webknjaz:
So this appears to be my fault: https://github.com/ansible/pylibssh/pull/501/files replaced the TOML library with the stdlib's `tomllib` and it's compatible PyPI counterpart — `tomli` under older Pythons. The spec file still vendors the `toml` lib that used to be needed by our build backend.

webknjaz:
I fixed this stuff right on devel.

PR #562, opened 2024-01-22
Build Linux platform wheels with `cibuildwheel`

### LINKED ISSUE #324: Stop building manylinux1 wheels (eventually)
It is increasingly awkward to build manylinux1 wheels

* They are not supported by platforms other than x86_64 (and i686 which we do not build currently)
* The images we rely on to build them are deprecated
* Those same images do not have versions of Python > 3.9
* manylinux1 itself is based on CentOS 5 which reached EoL in 2017

We can and will continue to build them for now, but we will have to drop support for it at some point. This issue is for tracking that.

webknjaz:
##### SUMMARY

$sbj.

##### ISSUE TYPE

- Maintenance Pull Request
- Packaging Pull Request

##### ADDITIONAL INFORMATION

N/A


webknjaz:
Self note: cibuildwheel doesn't clean the build dir in between building different wheels in the same container.

Seems to be something mentioned in https://github.com/pypa/cibuildwheel/issues/1693.

.github/workflows/ci-cd.yml
```diff
@@ -649,6 +649,54 @@ jobs:
...
+          && !fromJSON(needs.pre-setup.outputs.release-requested)
+        ) && 'test' || ''
+      }}] manylinux
+    needs:
+    - build-src
+    - pre-setup  # transitive, for accessing settings
+    strategy:
+      matrix:
+        os:
+        - ubuntu
```
webknjaz:
```suggestion
        - ubuntu-latest
```

PR #565, opened 2024-01-23
Fix the workflow location in the release doc

Qalthos:
##### SUMMARY

The Release Guide documentation points to a workflow that no longer exists. This fixes that link.

##### ISSUE TYPE

- Docs Pull Request

### DIFF
--- a/docs/changelog-fragments/565.contrib.rst
+++ b/docs/changelog-fragments/565.contrib.rst
@@ -1,0 +1,2 @@
+Fixed the location of release workflow in the
+:ref:`Release Guide` document -- by :user:`Qalthos`.
--- a/docs/contributing/release_guide.rst
+++ b/docs/contributing/release_guide.rst
@@ -86,4 +86,4 @@ The release stage
 
 
 .. _GitHub Actions CI/CD workflow:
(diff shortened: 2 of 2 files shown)


docs/changelog-fragments/565.doc.rst
```diff
@@ -0,0 +1,2 @@
+Fix location of release workflow in the :ref:`Release Guide`
+-- :user:`Qalthos`.
```
webknjaz: Use the past tense to have the changelog consistency, plz.
```suggestion
Fixed the location of release workflow in the
:ref:`Release Guide` document -- by :user:`Qalthos`.
```
Overall the changelog should read as what changed since the previous release. Hence it's either the past tense or the present tense (non-imperative, unlike Git commits that have contributors as the target audience) with structures like "this is now X", referring to the difference. This will target the audience of the end users.

docs/changelog-fragments/565.doc.rst
webknjaz, line 1: I think this might be more relevant under the `contrib` section, rather than `doc`. (could be displayed both, if you symlink an additional filename, but this doesn't seem to be end-user-facing).

webknjaz (approved):

Issue #57, opened 2020-06-01
[BUG] Flaky SEGFAULT while testing `exec_command`


webknjaz:
_https://github.com/ansible/pylibssh/runs/725098816?check_suite_focus=true#step:10:146_:

```python
[gw1] [ 16%] PASSED tests/unit/channel_test.py::test_exec_command Fatal Python error: Segmentation fault

Thread 0x00007f13f32a4700 (most recent call first):
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/execnet/gateway_base.py", line 400 in read
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/execnet/gateway_base.py", line 432 in from_io
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/execnet/gateway_base.py", line 967 in _thread_receiver
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/execnet/gateway_base.py", line 220 in run
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/execnet/gateway_base.py", line 285 in _perform_spawn

Current thread 0x00007f13f58cf680 (most recent call first):
  File "/home/runner/work/pylibssh/pylibssh/tests/unit/channel_test.py", line 22 in ssh_channel
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/_pytest/fixtures.py", line 800 in _teardown_yield_fixture
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/_pytest/fixtures.py", line 871 in finish
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/_pytest/runner.py", line 318 in _callfinalizers
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/_pytest/runner.py", line 328 in _teardown_with_finalization
  File "/home/runner/work/pylibssh/pylibssh/.tox/test-binary-dists/lib/python3.7/site-packages/_pytest/runner.py", line 310 in _pop_and_teardown
[...]

PR #576, opened 2024-02-02
Allow request_exec with async I/O and more

pbrezina:
This PR add calls to ``libssh`` ``request_exec()`` and relevant functions in order
to allow non-blocking async I/O. I need this to replace ``parallel-ssh`` which
unfortunately became a dead project, incompatible with Python 3.12+.

Additionally, it adds call to ``libssh`` ``request_send_signal()`` which allows to
send signals to the remote process if SSH server supports it.

I also fixed few bugs that I stumbled upon.

##### SUMMARY

$sbj.

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/docs/changelog-fragments/576.bugfix.rst
+++ b/docs/changelog-fragments/576.bugfix.rst
@@ -1,0 +1,7 @@
+|project| no longer crashes when received EOF or when channel is not explicitly
+closed -- by :user:`pbrezina`.
+
+Previously, |project| crashed if ``channel.recv`` was called and ``libssh``
+returned ``SSH_EOF`` error. It also crashed on some special occasions where
+channel was not explicitly closed and the session object was garbage-collected
+first.
--- a/docs/changelog-fragments/576.doc.rst
+++ b/docs/changelog-fragments/576.doc.rst
@@ -1,0 +1,4 @@
+Fixed the example of invoking remote commands by using
+``Channel.exec_command()`` in snippets -- by :user:`pbrezina`.
+
+Its previously showcased version wasn't functional.
--- a/docs/changelog-fragments/576.feature.rst
+++ b/docs/changelog-fragments/576.feature.rst
@@ -1,0 +1,7 @@
+The ``request_exec()`` method was added to the ``Channel`` class. It exposes an
+interface for calling the respective low-level C-API of the underlying
+``libssh`` library -- by :user:`pbrezina`.
+
+Additionally, the following calls to ``libssh`` are now available in the same
+class: ``request_exec()``, ``send_eof()``, ``request_send_signal()`` and
+``is_eof`` which is exposed as a :py:class:`property`.
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -204,6 +204,12 @@ cdef class Channel:
     def get_channel_exit_status(self):
         return libssh.ssh_channel_get_exit_status(self._libssh_channel)
 
+    @property
+    def is_eof(self):
+        """True if remote has sent an EOF."""
+        rc = libssh.ssh_channel_is_eof(self._libssh_channel)
+        return rc != 0
+
     def close(self):
         if self._libssh_channel is not NULL:
             libssh.ssh_channel_close(self._libssh_channel)
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -15,6 +15,7 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 #
+import signal
 import time
 from io import BytesIO
 
@@ -181,6 +182,21 @@ cdef class Channel:
         rc = libssh.ssh_channel_send_eof(self._libssh_channel)
         if rc != libssh.SSH_OK:
             raise LibsshChannelException("Failed to ssh_channel_send_eof: [%d]" % rc)
+
+    def send_signal(self, sig):
+        """
+        Send signal to the remote process.
+
+        :param sig: a signal constant from ``signal``, e.g. ``signal.SIGUSR1``.
+        :type sig: signal.Signals
+        """
+        if not isinstance(sig, signal.Signals):
+            raise TypeError(f"Expecting signal.Signals not {type(sig)}")
+
+        sshsig = sig.name.replace("SIG", "")  # FIXME: replace w/ `str.removeprefix()` once Python 3.8 support is dropped
+        rc = libssh.ssh_channel_request_send_signal(self._libssh_channel, sshsig.encode("utf-8"))
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to ssh_channel_request_send_signal: [%d]" % rc)
 
     def get_channel_exit_status(self):
         return libssh.ssh_channel_get_exit_status(self._libssh_channel)
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -47,6 +47,7 @@ cdef int _process_outputs(libssh.ssh_session session,
 
 cdef class Channel:
     def __cinit__(self, session):
+        self._session = session
         self._libssh_session = get_libssh_session(session)
         self._libssh_channel = libssh.ssh_channel_new(self._libssh_session)
 
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -70,6 +70,12 @@ cdef class Channel:
         if rc != libssh.SSH_OK:
             raise LibsshChannelException("Failed to request_shell: [%d]" % rc)
 
+    def request_exec(self, command):
+        """Run a shell command without an interactive shell."""
+        rc = libssh.ssh_channel_request_exec(self._libssh_channel, command.encode("utf-8"))
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to request_exec: [%d]" % rc)
+
     def request_pty(self):
         rc = libssh.ssh_channel_request_pty(self._libssh_channel)
         if rc != libssh.SSH_OK:
@@ -169,6 +175,12 @@ cdef class Channel:
             libssh.ssh_channel_free(channel)
 
         return result
+
+    def send_eof(self):
+        """Send EOF to the channel, this will close stdin."""
+        rc = libssh.ssh_channel_send_eof(self._libssh_channel)
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to ssh_channel_send_eof: [%d]" % rc)
 
     def get_channel_exit_status(self):
         return libssh.ssh_channel_get_exit_status(self._libssh_channel)
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -110,6 +110,9 @@ cdef class Channel:
             # This is what Session._get_session_error_str() does, but we don't have the Python object
             error = libssh.ssh_get_error(<void*>self._libssh_session).decode()
             raise LibsshChannelReadFailure(error)
+        elif nbytes == libssh.SSH_EOF:
+            return None
+
         return <bytes>buffer[:nbytes]
 
     def recv(self, size=1024, stderr=0):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,6 +6,7 @@
 import shutil
 import socket
 import subprocess
+from functools import partial
 
 import pytest
 from _service_utils import (  # noqa: WPS436
@@ -105,7 +106,7 @@ def ssh_clientkey_path(sshd_path):
 
 
 @pytest.fixture
-def ssh_client_session(sshd_addr, ssh_clientkey_path):
+def ssh_client_session(ssh_session_connect):
     """Authenticate against SSHD with a private SSH key.
 
     :yields: Pre-authenticated SSH session.
@@ -114,12 +115,29 @@ def ssh_client_session(sshd_addr, ssh_clientkey_path):
     # noqa: DAR101
     """
     ssh_session = Session()
-    ensure_ssh_session_connected(ssh_session, sshd_addr, ssh_clientkey_path)
+    ssh_session_connect(ssh_session)
     try:  # noqa: WPS501
         yield ssh_session
     finally:
         ssh_session.close()
         del ssh_session  # noqa: WPS420
+
+
+@pytest.fixture
+def ssh_session_connect(sshd_addr, ssh_clientkey_path):
+    """
+    Authenticate existing session object against SSHD with a private SSH key.
+
+    It returns a function that takes session as parameter.
+
+    :returns: Function that will connect the session.
+    :rtype: Callback
+    """
+    return partial(
+        ensure_ssh_session_connected,
+        sshd_addr=sshd_addr,
+        ssh_clientkey_path=ssh_clientkey_path,
+    )
 
 
 @pytest.fixture
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -157,7 +157,7 @@ def ssh_authorized_keys_path(sshd_path, ssh_clientkey_path):
 
 
 @pytest.fixture
-def sshd_addr(free_port_num, ssh_authorized_keys_path, sshd_hostkey_path, sshd_path):
+def sshd_addr(free_port_num, ssh_authorized_keys_path, sshd_hostkey_path, sshd_path, ssh_clientkey_path):
     """Spawn an instance of sshd on a free port.
 
     :raises RuntimeError: If spawning SSHD failed.
@@ -198,7 +198,7 @@ def sshd_addr(free_port_num, ssh_authorized_keys_path, sshd_hostkey_path, sshd_p
     )
     proc = subprocess.Popen(cmd)
 
-    wait_for_svc_ready_state(hostname, free_port_num, b'SSH-2.0-OpenSSH_')
+    wait_for_svc_ready_state(hostname, free_port_num, ssh_clientkey_path)
 
     if proc.returncode:
         raise RuntimeError('sshd boom 💣')
--- a/tests/unit/channel_test.py
+++ b/tests/unit/channel_test.py
@@ -2,6 +2,7 @@
 
 """Tests suite for channel."""
 
+import signal
 import time
 
 import pytest
@@ -9,6 +10,7 @@ import pytest
 
 COMMAND_TIMEOUT = 30
 POLL_EXIT_CODE_TIMEOUT = 5
+POLL_TIMEOUT = 5000
 
 
 @pytest.fixture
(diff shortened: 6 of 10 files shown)


docs/changelog-fragments/576.bugfix.rst
```diff
@@ -0,0 +1,6 @@
+pylibssh no longer crashes when received EOF or when channel is not explicitly
+closed -- by :user:`pbrezina`
```
webknjaz:
```suggestion
closed -- by :user:`pbrezina`.
```

docs/changelog-fragments/576.bugfix.rst
```diff
@@ -0,0 +1,6 @@
...
+
+Previously, pylibssh crashed if channel.recv was called and libssh returned
```
webknjaz:
```suggestion
Previously, pylibssh crashed if ``channel.recv()`` was called and ``libssh`` returned
```

docs/changelog-fragments/576.bugfix.rst
```diff
@@ -0,0 +1,6 @@
...
+SSH_EOF error. It also crashed on some special occasions where channel was not
```
webknjaz:
```suggestion
the ``SSH_EOF`` error. It also crashed on some special occasions where channel was not
```

tests/unit/channel_test.py
```diff
@@ -2,9 +2,13 @@
 
 """Tests suite for channel."""
 
+import gc
 import time
 
 import pytest
+from _service_utils import ensure_ssh_session_connected  # noqa: WPS436
```
webknjaz: Those utils aren't meant to be imported directly. They only serve as helpers to the fixtures defined in `conftest.py`.
Use `ssh_client_session` and you'll get the same thing with less code.

tests/unit/channel_test.py
```diff
@@ -68,3 +72,82 @@ def test_read_bulk_response(ssh_client_session):
...
+    """
+    def _do_not_crash():  # noqa: WPS430
+        ssh_session = Session()
+        ensure_ssh_session_connected(ssh_session, sshd_addr, ssh_clientkey_path)
+        ssh_channel = ssh_session.new_channel()  # noqa: F841
+
+    # Without fix, garbage collector first deletes session and we segfault
+    # in channel destructor when trying to access low-level C session object.
+    _do_not_crash()
+    gc.collect(0)
```
webknjaz: It should be more stable if you first disable GC, then collect everything, and only then run the test helper, finally performing the collection call that checks for crashes. You could also assert the number of objects, maybe. Before and after. Then, re-enable GC upon completion of the test.

You could integrate `pypytools` and make a helper like https://github.com/cherrypy/cheroot/blob/0fd16f0c3f111c3709a4d14e6400cabf2c094239/cheroot/test/test_server.py#L412-L416 but that's not mandatory. Manual invocations in just one place is fine.

tests/unit/channel_test.py
```diff
@@ -68,3 +72,82 @@ def test_read_bulk_response(ssh_client_session):
...
+    """
+    ssh_channel.request_exec('exit 0')
+    ssh_channel.poll(timeout=1000)
+    ssh_channel.recv()
+
+
+def test_is_eof(ssh_channel):
+    """Test that EOF-state is correctly obtained with is_eof."""
+    ssh_channel.request_exec('exit 0')
+    ssh_channel.poll(timeout=5000)  # noqa: WPS432
```
webknjaz: Disabling this linting rule is not justified here. Assign a name to that arbitrary number of `5_000` that labels it with some context.

src/pylibsshext/channel.pyx
```diff
@@ -170,9 +180,23 @@ cdef class Channel:
...
+
+    def send_signal(self, sig):
+        rc = libssh.ssh_channel_request_send_signal(self._libssh_channel, sig.encode("utf-8"))
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to ssh_channel_request_send_signal: [%d]" % rc)
+
     def get_channel_exit_status(self):
         return libssh.ssh_channel_get_exit_status(self._libssh_channel)
 
+    def is_eof(self):
```
webknjaz, line 209: This method doesn't depend on any arguments and its name doesn't suggest it causes a side effect (it doesn't contain a verb). So it's actually a property, semantically:
```suggestion
    @property
    def is_eof(self):
```

webknjaz: Could you also add a docstring for this?

src/pylibsshext/channel.pyx
```diff
@@ -170,9 +180,23 @@ cdef class Channel:
 
         return result
 
+    def send_eof(self):
+        rc = libssh.ssh_channel_send_eof(self._libssh_channel)
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to ssh_channel_send_eof: [%d]" % rc)
+
+    def send_signal(self, sig):
```
webknjaz, line 190: Could you make this accept constants from the standard library module called `signal`?
Also, add a docstring, please.

src/pylibsshext/channel.pyx
```diff
@@ -170,9 +180,23 @@ cdef class Channel:
 
         return result
 
+    def send_eof(self):
```
webknjaz, line 184: Add a docstring here, please.

src/pylibsshext/channel.pyx
```diff
@@ -103,6 +109,10 @@ cdef class Channel:
             # This is what Session._get_session_error_str() does, but we don't have the Python object
             error = libssh.ssh_get_error(<void*>self._libssh_session).decode()
             raise LibsshChannelReadFailure(error)
+
+        if nbytes == libssh.SSH_EOF:
```
webknjaz: Perhaps, it's an `elif`?
```suggestion
        elif nbytes == libssh.SSH_EOF:
```

src/pylibsshext/channel.pyx
```diff
@@ -70,6 +71,11 @@ cdef class Channel:
         if rc != libssh.SSH_OK:
             raise LibsshChannelException("Failed to request_shell: [%d]" % rc)
 
+    def request_exec(self, command):
```
webknjaz, line 75: Could you add a docstring here?

src/pylibsshext/channel.pyx
```diff
@@ -46,6 +46,7 @@ cdef int _process_outputs(libssh.ssh_session session,
 
 cdef class Channel:
     def __cinit__(self, session):
+        self._session = session
```
webknjaz, line 50: @pbrezina is this the place with the actual fix for the SEGFAULT?

docs/changelog-fragments/576.feature.rst
```diff
@@ -0,0 +1,4 @@
+Published call to request_exec and other low level API -- by :user:`pbrezina`
```
webknjaz:
```suggestion
The ``request_exec()`` method was added to the ``Channel``
class. It exposes an interface for calling the respective
low-level C-API of the underlying ``libssh`` library
-- by :user:`pbrezina`.
```

docs/changelog-fragments/576.feature.rst
```diff
@@ -0,0 +1,4 @@
...
+
+The following calls to libssh are now available: request_exec, send_eof, is_eof,
+request_send_signal.
```
webknjaz:
```suggestion
Additionally, the following calls to ``libssh`` are now
available in the same class: ``request_exec()``, ``send_eof()``, ``request_send_signal()`` and ``is_eof``
which is exposed as a :py:class:`property`.
```

docs/changelog-fragments/576.doc.rst
```diff
@@ -0,0 +1,3 @@
+Fixed example for exec_command -- by :user:`pbrezina`
+
+The example now contains correct method code and is fully functional.
```
webknjaz:
```suggestion
Fixed the example of invoking remote commands by using
``Channel.exec_command()`` in snippets -- by :user:`pbrezina`.

Its previously showcased version wasn't as robust.
```

src/pylibsshext/channel.pyx
webknjaz: Is this going to fix https://github.com/ansible/pylibssh/issues/57? If yes, then plz symlink `57.bugfix.rst` to `576.bugfix.rst`.

tests/unit/channel_test.py
```diff
@@ -68,3 +72,82 @@ def test_read_bulk_response(ssh_client_session):
...
+def test_destructor(sshd_addr, ssh_clientkey_path):
+    """
+    Garbage collector can destroy session before channel.
+
+    Test that this event does not cause a segfault in channels destructor.
+    """
+    def _do_not_crash():  # noqa: WPS430
+        ssh_session = Session()
+        ensure_ssh_session_connected(ssh_session, sshd_addr, ssh_clientkey_path)
+        ssh_channel = ssh_session.new_channel()  # noqa: F841
```
webknjaz: Would it be easier to trigger if this is linked to an external object?
```suggestion
        _do_not_crash.ssh_channel = ssh_session.new_channel()  # noqa: F841  # setting a non-accessed var is needed for testing GC
```

tests/unit/channel_test.py
```diff
@@ -68,3 +72,82 @@ def test_read_bulk_response(ssh_client_session):
...
+    assert ssh_channel.is_eof()
+
+
+def test_destructor(sshd_addr, ssh_clientkey_path):
+    """
+    Garbage collector can destroy session before channel.
+
+    Test that this event does not cause a segfault in channels destructor.
+    """
+    def _do_not_crash():  # noqa: WPS430
```
webknjaz:
```suggestion
    def _do_not_crash():  # noqa: WPS430  # required to create a garbage-collection scope
```

src/pylibsshext/channel.pyx
pbrezina: It's likely, yes. But I can't tell without a coredump.

This is a fix for segfault I seen, test for it is in the test_destructor. The problem is that you open channel in cinit, then close it in destructor. To close it, you need to have valid session object. But if you rely on the destructor to close the channel, then it may happen that Session is gc'd first, therefore libssh_session is freed before you try to close the channel.

This patch makes sure that we keep reference to the python session object (not only to libssh session object) and therefore Channel is always gc'd before Session.

docs/changelog-fragments/576.doc.rst
pbrezina: I wouldn't say robust, it simply was not functional.

tests/unit/channel_test.py
pbrezina: I don't understand what you mean, can you ellaborate? But it is very easy to trigger it. Really just open session and channel and then let the program exit.

I added your suggestion.

tests/unit/channel_test.py
pbrezina: I can't, using `ssh_client_session` will not trigger the condition in Garbage Collector.

tests/unit/channel_test.py
pbrezina: I can't assert number of objects since it can be easily changed. Even `gc.is_tracked` can not be asserted as for some reason, it returns False for ssh_channel, but True for ssh_session. But I added `gc.disable` to have a better control over it.

Anyway, the crash is always triggered by the test code. Feel free to suggest improvements if you see any.

tests/unit/channel_test.py
pbrezina: I converted it to a fixture on its own.

tests/unit/channel_test.py
webknjaz: @pbrezina but in tests you have a separate nested function in hopes the variables in its scope would be garbage-collected in a breaking matter, right?

docs/changelog-fragments/576.doc.rst
webknjaz: @pbrezina was it nonfunctional unconditionally, then?

pbrezina:
It is ready for next round of review. I addresses all comments, leaving few as "unresolved" in case you want to discuss it further.

docs/changelog-fragments/576.doc.rst
pbrezina: Yes.

tests/unit/channel_test.py
pbrezina: Right. It needs to be garbage collected, so it needs to go out of scope.

src/pylibsshext/channel.pyx
```diff
@@ -15,14 +15,17 @@
...
 #
+import signal
 import time
 from io import BytesIO
 
 from cpython.bytes cimport PyBytes_AS_STRING
 from libc.string cimport memset
 
 from pylibsshext.errors cimport LibsshChannelException
+
```
webknjaz: Did some linter want these formatting changes?

docs/changelog-fragments/576.bugfix.rst
```diff
@@ -0,0 +1,7 @@
+pylibssh no longer crashes when received EOF or when channel is not explicitly
+closed -- by :user:`pbrezina`.
+
+Previously, pylibssh crashed if ``channel.recv`` was called and ``libssh``
```
webknjaz: I just realized that this refers to `pylibssh` which is not an identifier used anywhere in this project. The PyPI project is `ansible-pylibssh` and the corresponding GitHub slug is `ansible/pylibssh`. The importable is `pylibsshext`. Let's try using the project name here. I hope the `|project|` RST substitution declared in `docs/conf.py` will just work:

```suggestion
|project| no longer crashes when received EOF or when channel is not explicitly
closed -- by :user:`pbrezina`.

Previously, |project| crashed if ``channel.recv`` was called and ``libssh``
```

src/pylibsshext/channel.pyx
```diff
@@ -170,9 +183,35 @@ cdef class Channel:
...
+    def send_signal(self, sig):
+        """
+        Send signal to the remote process.
+
+        @sig is a signal constant from ``signal``, e.g. ``signal.SIGUSR1``.
+        """
+        if not isinstance(sig, signal.Signals):
+            raise TypeError(f"Expecting signal.Signals not {type(sig)}")
+
+        sshsig = sig.name.replace("SIG", "")
```
webknjaz:
```suggestion
        sshsig = sig.name.replace("SIG", "")  # FIXME: replace w/ `str.removeprefix()` once Python 3.8 support is dropped
```

src/pylibsshext/channel.pyx
```diff
@@ -170,9 +183,35 @@ cdef class Channel:
...
+        """Send EOF to the channel, this will close stdin."""
+        rc = libssh.ssh_channel_send_eof(self._libssh_channel)
+        if rc != libssh.SSH_OK:
+            raise LibsshChannelException("Failed to ssh_channel_send_eof: [%d]" % rc)
+
+    def send_signal(self, sig):
+        """
+        Send signal to the remote process.
+
+        @sig is a signal constant from ``signal``, e.g. ``signal.SIGUSR1``.
```
webknjaz: Let's follow the native Sphinx param list style just like in other places:
```suggestion
        :param sig: a signal constant from ``signal``, e.g. ``signal.SIGUSR1``.
        :type sig: 
```

Alternatively, you could've used double backticks for `sig`. But the `@sig` syntax is not something that is typical for Python (I believe it comes from Java or JS?).

tests/unit/channel_test.py
```diff
@@ -68,3 +74,87 @@ def test_read_bulk_response(ssh_client_session):
...
+
+def test_recv_eof(ssh_channel):
+    """
+    Test that reading EOF does not raise error.
+
+    SystemError: Negative size passed to PyBytes_FromStringAndSize
+    """
+    ssh_channel.request_exec('exit 0')
+    ssh_channel.poll(timeout=POLL_TIMEOUT)
+    ssh_channel.recv()
```
webknjaz, line 132: Does this test rely on the session reaching EOF before invoking `recv()`? If yes, I suggest being explicit about that in the test:
```suggestion
    assert ssh_channel.is_eof
    ssh_channel.recv()
```

tests/conftest.py
```diff
@@ -122,6 +123,23 @@ def ssh_client_session(sshd_addr, ssh_clientkey_path):
         del ssh_session  # noqa: WPS420
 
 
+@pytest.fixture
+def ssh_session_connect(sshd_addr, ssh_clientkey_path):
```
webknjaz, line 127: Could you inject this fixture into `ssh_client_session` so there's no duplication?

src/pylibsshext/channel.pyx
pbrezina: It must have been isort.

src/pylibsshext/channel.pyx
pbrezina: doxygen, fixed

src/pylibsshext/channel.pyx
pbrezina: Removed.

pbrezina:
Thank you. It is ready for next round.

src/pylibsshext/channel.pyx
webknjaz: Alright, thank you for the explanation! I'm marking this thread as resolved.

webknjaz (approved):
Thank you for the well-explained patch!

tests/unit/channel_test.py
```diff
@@ -68,3 +74,88 @@ def test_read_bulk_response(ssh_client_session):
...
+    ssh_channel.send_eof()
+
+    rc = -1
+    while rc == -1:
+        ssh_channel.poll(timeout=POLL_EXIT_CODE_TIMEOUT)
+        rc = ssh_channel.get_channel_exit_status()
+    assert rc == 0
+
+
+def test_send_signal(ssh_channel):
```
webknjaz, line 102: @pbrezina any idea why this test might be SEGFAULTing sometimes? See the CI log: https://github.com/ansible/pylibssh/actions/runs/7798061967/job/21271285995?pr=576#step:15:319.

pbrezina: If I read it correctly this test timed out when connecting to ssh server:
```
 >       raise TimeoutError('Timed out waiting for a successful connection')
E       TimeoutError: Timed out waiting for a successful connection
```

I saw this from time to time when running tests locally (also with the tests not related to my changes).

Segfault comes from
```
XFAIL tests/unit/channel_test.py::test_exec_command - reason: This test causes SEGFAULT, flakily. Ref: https://github.com/ansible/pylibssh/issues/57; pytest-forked reason: :-1: running the test CRASHED with signal 11
```

pbrezina: Which also means that #57 is not fixed by this.

pbrezina:
```
        libssh.ssh_channel_send_eof(channel)
        result.returncode = libssh.ssh_channel_get_exit_status(channel)
        if channel is not NULL:
            libssh.ssh_channel_close(channel)
            libssh.ssh_channel_free(channel)
```
I think you need to poll for exit status since it may still return -1 and you close the channel. Perhaps there is some race condition when the channel is being closed but callback is still called. We could be sure if we have coredump available but this is my wild guess.

webknjaz: Would you be able to take a look? I kinda don't want to make the CI even more flaky with this patch..

pbrezina: It's not getting more flaky since the problems are already there :-)

Anyway, I believe I have fixed the timeout issue. See the last commit.

For the segfault... is it possible to get a coredump and full backtrace out of the tests? I never hit it locally and I don't think we can fix it without it. We can try to implement my proposal, but then again, if it was true, it should be possible to write a solid reproducer and I was not able to find one.

webknjaz: Well, I googled and couldn't understand how to retrieve that core. The faulthandler is enabled and prints out some trace through `catchsegv`. I wish there was a way to put it into a GHA artifact, though.

Thanks for the patch!

PR #602, opened 2024-06-07
🧪 Make post-gate job chain run on gate success

webknjaz:
Previously, jobs following the `check` job didn't have `if` conditionals
or the existing expressions didn't use the status check functions. Those
jobs are responsible for the finishing steps of the release automation.
They used to work when the entire parent job tree was successful. The
release automation is triggered by the `workflow_dispatch` event, which
has a boolean flag/checkbox for disregarding the testing stage status.
Using it makes the `check` gate pass even when some test matrix jobs
fail. It was not obvious that jobs the depend on `check` don't run if
any of the parent chain jobs fail. Not by default, so they were being
skipped for no obvious reason.

This patch updates said `if` conditionals to force GitHub Actions CI/CD
to only take into account the `check` job result and not its
dependencies by prepending the trick of
`always() && needs.check.result == 'success' &&` in front of each `if`
conditional in publishing jobs.

Refs:
* actions/runner#2205 (comment)
* https://github.com/orgs/community/discussions/45058#discussioncomment-7465690
* https://docs.github.com/en/actions/learn-github-actions/expressions#status-check-functions

Co-Authored-By: Pavel Březina <pbrezina@redhat.com>

##### ISSUE TYPE

- Bugfix Pull Request
- Maintenance Pull Request
- Testing Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1540,7 +1540,9 @@ jobs:
     - check
     - pre-setup  # transitive, for accessing settings
     if: >-
-      fromJSON(needs.pre-setup.outputs.release-requested)
+      always()
+      && needs.check.result == 'success'
+      && fromJSON(needs.pre-setup.outputs.release-requested)
     runs-on: ubuntu-latest
 
     environment:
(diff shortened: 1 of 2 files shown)


webknjaz:
I reverted 30a9da5c3fa8bb8728d03f725aabad864f552d53 because https://github.com/ansible/pylibssh/actions/runs/9414210260 revealed that the workflow seems to go into the canceled state after ignored failures anyway so it's not possible to rely on this check. `always()` will work better, as originally suggested.

webknjaz:
I also reverted f5597a02424594e34aa9242cc6d64c46cce23992 for the same reason — I probably missed something in the generalization example, so I will have to get back to trying it out later.

webknjaz:
The workflow is now functional in YOLO mode: https://github.com/ansible/pylibssh/actions/runs/9419572873. Finally.

Issue #557, opened 2024-01-12
[CI] Sometimes the tests hang and are killed after a 6-hour timeout


webknjaz:
Example: https://github.com/ansible/pylibssh/actions/runs/7493462482/job/20399328667#step:15:1.

Applying `timeout-minutes` on GHA is the least that should be done.

webknjaz:
Fixed by #706

PR #621, opened 2024-06-21
Fix reading large files over SCP

Jakuje:
##### SUMMARY
The current SCP code does not handle large reads. Instead just allocates the large buffer fills up to 64k bytes and then writes the whole to the file (most likely the rest filled with junk).

The libssh returns the number of read bytes from `ssh_scp_read()`  and for practical reasons does not allocate the whole file in memory. The reads are capped to 64k B.

This PR fixes the read by both not allocating the entire file size in memory and reading it by chunks (that much max libssh chunk atm) and writing that to the local file.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION

The attached test case is failing without the change, with errors like this:
```console
_______________________________________________________________________ test_get_large _______________________________________________________________________
[gw0] linux -- Python 3.12.3 /home/jjelen/devel/pylibssh/.tox/python/bin/python

dst_path = PosixPath('/tmp/pytest-of-jjelen/pytest-35/popen-gw0/test_get_large0/dst-file.txt')
src_path_large = PosixPath('/tmp/pytest-of-jjelen/pytest-35/popen-gw0/test_get_large0/large.txt'), ssh_scp = <pylibsshext.scp.SCP object at 0x7f3f83cffac0>
large_payload = b'&PE~~,!=1@F7RaUccg#)q-ru^~f8Zv/-{~tFE-\\>v2_b}i@`N/]_%*~{oD`"F`*4Ns]HK^lewILvC1O[]gFC9(27%MuX$s&24et<dRJ$"& zf_XIqm7...**<J_;?uI47vm\'/A"xi5a!kQXoL=M}ne;$E[B=EpasVbRO&\\zy7t,GK yu;l*$s<#P:sJLi@od#:&:m?8]9KFN4@QAF8hD, ~T@WNP;)%+>)TV\'ODi9'

    def test_get_large(dst_path, src_path_large, ssh_scp, large_payload):
        """Check that SCP file download works also for large files."""
>       ssh_scp.get(str(src_path_large), str(dst_path))

dst_path   = PosixPath('/tmp/pytest-of-jjelen/pytest-35/popen-gw0/test_get_large0/dst-file.txt')
large_payload = (b'&PE~~,!=1@F7RaUccg#)q-ru^~f8Zv/-{~tFE-\\>v2_b}i@`N/]_%*~{oD`"F`*4Ns]HK^le'
[...]
 b"?8]9KFN4@QAF8hD, ~T@WNP;)%+>)TV'ODi9")
src_path_large = PosixPath('/tmp/pytest-of-jjelen/pytest-35/popen-gw0/test_get_large0/large.txt')
ssh_scp    = <pylibsshext.scp.SCP object at 0x7f3f83cffac0>


tests/unit/scp_test.py:111: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   raise LibsshSCPException("Unexpected request: %s" % self._get_ssh_error_str())
E   pylibsshext.errors.LibsshSCPException: Unexpected request: b'ssh_scp_pull_request called under invalid state'


src/pylibsshext/scp.pyx:148: LibsshSCPException
```
(lines do not match)

Additionally, there is one more test case, that I suspected will be problematic, but it turned out it works just ok, so leaving for the coverage.

### DIFF
--- a/docs/changelog-fragments/621.bugfix.rst
+++ b/docs/changelog-fragments/621.bugfix.rst
@@ -1,0 +1,1 @@
+Downloading files larger than 64kB over SCP no longer fails -- by :user:`Jakuje`.
--- a/src/pylibsshext/scp.pyx
+++ b/src/pylibsshext/scp.pyx
@@ -24,6 +24,9 @@ from pylibsshext.errors cimport LibsshSCPException
 from pylibsshext.session cimport get_libssh_session
 
 
+SCP_MAX_CHUNK = 65536
+
+
 cdef class SCP:
     def __cinit__(self, session):
         self.session = session
@@ -122,7 +125,9 @@ cdef class SCP:
             size = libssh.ssh_scp_request_get_size(scp)
             mode = libssh.ssh_scp_request_get_permissions(scp)
 
-            read_buffer = <char *>PyMem_Malloc(size)
+            # cap the buffer size to reasonable number -- libssh will not return the whole data at once anyway
+            read_buffer_size = min(size, SCP_MAX_CHUNK)
+            read_buffer = <char *>PyMem_Malloc(read_buffer_size)
             if read_buffer is NULL:
                 raise LibsshSCPException("Memory allocation error")
 
@@ -131,14 +136,17 @@ cdef class SCP:
             if rc == libssh.SSH_ERROR:
                 raise LibsshSCPException("Failed to start read request: %s" % self._get_ssh_error_str())
 
-            # Read the file
-            rc = libssh.ssh_scp_read(scp, read_buffer, size)
-            if rc == libssh.SSH_ERROR:
-                raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
-
-            py_file_bytes = read_buffer[:size]
+            remaining_bytes_to_read = size
             with open(local_file, "wb") as f:
-                f.write(py_file_bytes)
+                while remaining_bytes_to_read > 0:
+                    requested_read_bytes = min(remaining_bytes_to_read, read_buffer_size)
+                    read_bytes = libssh.ssh_scp_read(scp, read_buffer, requested_read_bytes)
+                    if read_bytes == libssh.SSH_ERROR:
+                        raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
+
+                    py_file_bytes = read_buffer[:read_bytes]
+                    f.write(py_file_bytes)
+                    remaining_bytes_to_read -= read_bytes
             if mode >= 0:
                 os.chmod(local_file, mode)
 
--- a/tests/unit/scp_test.py
+++ b/tests/unit/scp_test.py
@@ -75,3 +75,17 @@ def test_copy_from_non_existent_remote_path(path_to_non_existent_src_file, ssh_s
     error_msg = '^Error receiving information about file:'
     with pytest.raises(LibsshSCPException, match=error_msg):
         ssh_scp.get(str(path_to_non_existent_src_file), os.devnull)
+
+
+@pytest.fixture
+def pre_existing_file_path(tmp_path):
+    """Return local path for a pre-populated file."""
+    path = tmp_path / 'pre-existing-file.txt'
+    path.write_bytes(b'whatever')
+    return path
+
+
+def test_get_existing_local(pre_existing_file_path, src_path, ssh_scp, transmit_payload):
+    """Check that SCP file download works and overwrites local file if it exists."""
+    ssh_scp.get(str(src_path), str(pre_existing_file_path))
+    assert pre_existing_file_path.read_bytes() == transmit_payload
--- a/tests/unit/scp_test.py
+++ b/tests/unit/scp_test.py
@@ -105,7 +105,12 @@ def large_payload():
 
 @pytest.fixture
 def src_path_large(tmp_path, large_payload):
-    """Return a remote path that is a file larger than 64k B, which is the most libssh can read at once."""
+    """Return a remote path that to a 65537 byte-sized file.
+
+    Typical single-read chunk size is 64kB in ``libssh`` so
+    the test needs a file that would overflow that to trigger
+    the read loop.
+    """
     path = tmp_path / 'large.txt'
     path.write_bytes(large_payload)
     return path
--- a/tests/unit/scp_test.py
+++ b/tests/unit/scp_test.py
@@ -4,6 +4,7 @@
 
 import os
 import random
+import string
 import uuid
 
 import pytest
@@ -94,16 +95,12 @@ def test_get_existing_local(pre_existing_file_path, src_path, ssh_scp, transmit_
 
 @pytest.fixture
 def large_payload():
-    """Generate a large test payload."""
-    # buffer of random 1024 bytes -- just printable ones for easier debugging
-    rands = []
-    for _ in range(1024):
-        rnd = chr(random.randint(ord(' '), ord('~')))
-        rands.append(rnd)
-    rand = ''.join(rands)
-    # .. repeated 65 times
-    repeat = 65
-    return ''.join(rand for _ in range(repeat)).encode()
+    """Generate a large 65537 byte (64kB+1B) test payload."""
+    random_char_kilobyte = [ord(random.choice(string.printable)) for _ in range(1024)]
+    full_bytes_number = 64
+    a_64kB_chunk = bytes(random_char_kilobyte * full_bytes_number)
+    the_last_byte = random.choice(random_char_kilobyte).to_bytes(length=1, byteorder='big')
+    return a_64kB_chunk + the_last_byte
 
 
 @pytest.fixture
--- a/tests/unit/scp_test.py
+++ b/tests/unit/scp_test.py
@@ -3,6 +3,7 @@
 """Tests suite for scp."""
 
 import os
+import random
 import uuid
 
 import pytest
@@ -89,3 +90,31 @@ def test_get_existing_local(pre_existing_file_path, src_path, ssh_scp, transmit_
     """Check that SCP file download works and overwrites local file if it exists."""
     ssh_scp.get(str(src_path), str(pre_existing_file_path))
     assert pre_existing_file_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large test payload."""
+    # buffer of random 1024 bytes -- just printable ones for easier debugging
+    rands = []
+    for _ in range(1024):
+        rnd = chr(random.randint(ord(' '), ord('~')))
+        rands.append(rnd)
+    rand = ''.join(rands)
+    # .. repeated 65 times
+    repeat = 65
+    return ''.join(rand for _ in range(repeat)).encode()
+
+
+@pytest.fixture
+def src_path_large(tmp_path, large_payload):
+    """Return a remote path that is a file larger than 64k B, which is the most libssh can read at once."""
+    path = tmp_path / 'large.txt'
+    path.write_bytes(large_payload)
+    return path
+
+
+def test_get_large(dst_path, src_path_large, ssh_scp, large_payload):
+    """Check that SCP file download gets over 64kB of data."""
+    ssh_scp.get(str(src_path_large), str(dst_path))
+    assert dst_path.read_bytes() == large_payload


webknjaz:
@Jakuje FYI the linting failures block this PR

docs/changelog-fragments/621.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed reading large files over SCP -- by :user:`Jakuje`
```
webknjaz: Please, add a trailing period to the sentence. Also, it'd be nice to elaborate on this in a way that it would prompt the users of what changed in the behavior compared to the previous release. High-level, not need to be very detailed.

src/pylibsshext/scp.pyx
```diff
@@ -122,6 +122,10 @@ cdef class SCP:
             size = libssh.ssh_scp_request_get_size(scp)
             mode = libssh.ssh_scp_request_get_permissions(scp)
 
+            # cap the buffer size to reasonable number -- libssh will not return the whole data at once anyway
+            read_buffer_size = size
+            if read_buffer_size > 65536:
+                read_buffer_size = 65536
```
webknjaz: There's an elegant helper for this:
```suggestion
            read_buffer_size = min(size, 65536)
```

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +135,17 @@ cdef class SCP:
...
-            # Read the file
-            rc = libssh.ssh_scp_read(scp, read_buffer, size)
-            if rc == libssh.SSH_ERROR:
-                raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
-
-            py_file_bytes = read_buffer[:size]
             with open(local_file, "wb") as f:
-                f.write(py_file_bytes)
+                left = size
+                while left > 0:
```
webknjaz: Let's polish the vars to assign better meaning:
```suggestion
                remaining_bytes_to_read = size
                while remaining_bytes_to_read > 0:
```

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +135,17 @@ cdef class SCP:
...
+                    # Read the file
+                    read_bytes = libssh.ssh_scp_read(scp, read_buffer, left)
```
webknjaz:
```suggestion
                    read_bytes = libssh.ssh_scp_read(scp, read_buffer, remaining_bytes_to_read)
```

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +135,17 @@ cdef class SCP:
...
+                    if read_bytes == libssh.SSH_ERROR:
+                        raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
+
+                    py_file_bytes = read_buffer[:read_bytes]
+                    f.write(py_file_bytes)
+                    left = left - read_bytes
```
webknjaz:
```suggestion
                    remaining_bytes_to_read -= read_bytes
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,39 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+@pytest.fixture
+def src_path_large(tmp_path, large_payload):
+    """Return a remote path that is a file larger than 64k B, which is the most libssh can read at once."""
+    path = tmp_path / 'large.txt'
+    path.write_bytes(large_payload)
+    return path
+
+
+def test_get_large(dst_path, src_path_large, ssh_scp, large_payload):
+    """Check that SCP file download works also for large files."""
```
webknjaz: I think this would be a more accurate high-level description:
```suggestion
    """Check that SCP file download gets over 64kB of data."""
```

Jakuje:
The suggestions accepted, the linting should pass now.

docs/changelog-fragments/621.bugfix.rst
```diff
@@ -0,0 +1 @@
+Downloading files larger thank 64kB over SCP no longer fails. -- by :user:`Jakuje`
```
webknjaz: The trailing period should be after the byline, if it continues a sentence. Otherwise, a byline w/o a trailing period should go to a separate line with an empty-line separator.
```suggestion
Downloading files larger thank 64kB over SCP no longer fails -- by :user:`Jakuje`.
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,45 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+
+def test_get_existing_local(dst_path_exists, src_path, ssh_scp, transmit_payload):
+    """Check that SCP file download works and overrides local file if it exists."""
+    ssh_scp.get(str(src_path), str(dst_path_exists))
+    assert dst_path_exists.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large test payload."""
```
webknjaz: Let's be a bit more informative here:
```suggestion
    """Generate a large 65kB test payload."""
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,45 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    # buffer of random 1024 bytes -- just printable ones for easier debugging
+    rands = []
+    for _ in range(1024):
+        rnd = chr(random.randint(ord(' '), ord('~')))
+        rands.append(rnd)
+    rand = ''.join(rands)
+    # .. repeated 65 times
```
webknjaz: This is already present in the variable name. Code comments shouldn't be used to repeat what the code does — just to explain motivation and unobvious considerations. For everything else — there's proper naming of vars and other identifiers.
```suggestion
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,45 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
     """Check that SCP file download works."""
     ssh_scp.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def dst_path_exists(tmp_path):
+    """Return a local path that exists."""
+    path = tmp_path / 'non-existing.txt'
+    path.write_bytes('whatever'.encode())
```
webknjaz: just use a byte string right away
```suggestion
    path.write_bytes(b'whatever')
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,45 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+        rands.append(rnd)
+    rand = ''.join(rands)
+    # .. repeated 65 times
+    repeat = 65
+    return ''.join(rand for _ in range(repeat)).encode()
+
+
+@pytest.fixture
+def src_path_large(tmp_path, large_payload):
+    """Return a remote path that is a file larger than 64k B, which is the most libssh can read at once."""
```
webknjaz:
```suggestion
    """Return a remote path that to a 65537 byte-sized file.
    
    Typical single-read chunk size is 64kB in ``libssh`` so
    the test needs a file that would overflow that to trigger
    the read loop.
    """
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,45 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    """Generate a large test payload."""
+    # buffer of random 1024 bytes -- just printable ones for easier debugging
+    rands = []
+    for _ in range(1024):
+        rnd = chr(random.randint(ord(' '), ord('~')))
+        rands.append(rnd)
+    rand = ''.join(rands)
+    # .. repeated 65 times
+    repeat = 65
+    return ''.join(rand for _ in range(repeat)).encode()
```
webknjaz: I think, we could do better and use just one byte over the limit. It would make a better edge-case test.

webknjaz: I realized there's a lot of conversions between integers, bytes and text strings back and forth here. So here's some refactoring to make it better. I'll post a series of comments with iterations of my refactoring and then, will commit it myself. Stay tuned!

This is the first improvement:

```suggestion
@pytest.fixture
def large_payload():
    """Generate a large 64kB+1B test payload."""
    # buffer of random 1024 bytes -- just printable ones for easier debugging
    rands = []
    for _ in range(1024):
        rnd = chr(random.randint(ord(' '), ord('~')))
        rands.append(rnd)
    rand = ''.join(rands)

    full_bytes_number = 64
    return ''.join(rand for _ in range(full_bytes_number)).encode() + rand[0]
```

webknjaz:
```suggestion
@pytest.fixture
def large_payload():
    """Generate a large 64kB+1B test payload."""
    # buffer of random 1024 bytes -- just printable ones for easier debugging
    whitespace_position = ord(' ')
    tilda_position = ord('~')
    rands = []
    for _ in range(1024):
        rnd = chr(random.randint(whitespace_position, tilda_position))
        rands.append(rnd)
    rand = ''.join(rands)

    full_bytes_number = 64
    return ''.join(rand for _ in range(full_bytes_number)).encode() + rand[0]
```

webknjaz:
```suggestion
from itertools import chain


@pytest.fixture
def large_payload():
    """Generate a large 65537 byte (64kB+1B) test payload."""
    # buffer of random 1024 bytes -- just printable ones for easier debugging
    whitespace_position = ord(' ')
    tilda_position = ord('~')
    rands = [random.randint(whitespace_position, tilda_position) for _ in range(1024)]

    full_bytes_number = 64
    return bytes(chain.from_iterable(rands for _ in range(full_bytes_number))) + rands[0].to_bytes(1)
```

webknjaz:
```suggestion
@pytest.fixture
def large_payload():
    """Generate a large 65537 byte (64kB+1B) test payload."""
    # buffer of random 1024 bytes -- just printable ones for easier debugging
    whitespace_position = ord(' ')
    tilda_position = ord('~')
    rands = [random.randint(whitespace_position, tilda_position) for _ in range(1024)]

    full_bytes_number = 64
    return bytes(rands * full_bytes_number) + rands[0].to_bytes(1)
```

webknjaz:
```suggestion
import string


@pytest.fixture
def large_payload():
    """Generate a large 65537 byte (64kB+1B) test payload."""
    random_char_kilobyte = [ord(random.choice(string.printable)) for _ in range(1024)]
    full_bytes_number = 64
    return bytes(random_char_kilobyte * full_bytes_number) + random_char_kilobyte[0].to_bytes(1)
```

tests/unit/scp_test.py
```diff
@@ -57,3 +58,42 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+
+def test_get_existing_local(dst_path_exists, src_path, ssh_scp, transmit_payload):
+    """Check that SCP file download works and overrides local file if it exists."""
+    ssh_scp.get(str(src_path), str(dst_path_exists))
+    assert dst_path_exists.read_bytes() == transmit_payload
+
+
+import string
+
+
```
webknjaz:
```suggestion
```

tests/unit/scp_test.py
```diff
@@ -2,6 +2,7 @@
 
 """Tests suite for scp."""
 
+import random
```
webknjaz, line 6:
```suggestion
import random
import string
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
     """Check that SCP file download works."""
     ssh_scp.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def dst_path_exists(tmp_path):
```
webknjaz: This name sounds like a check, but it's not. It should be renamed to reflect what it actually does.

webknjaz: Moreover, it mentions that something exists, but the code uses `non-existing` in the name. Which one is it? I'm confused.

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    """Return a local path that exists."""
+    path = tmp_path / 'non-existing.txt'
+    path.write_bytes(b'whatever')
+    return path
+
+
+def test_get_existing_local(dst_path_exists, src_path, ssh_scp, transmit_payload):
+    """Check that SCP file download works and overrides local file if it exists."""
```
webknjaz:
```suggestion
    """Check that SCP file download works and overwrites local file if it exists."""
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
     """Check that SCP file download works."""
     ssh_scp.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def dst_path_exists(tmp_path):
+    """Return a local path that exists."""
+    path = tmp_path / 'non-existing.txt'
```
webknjaz: oh...
```suggestion
    path = tmp_path / 'pre-existing-file.txt'
```

tests/unit/scp_test.py
webknjaz:
```suggestion
def pre_existing_file_path(tmp_path):
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    path.write_bytes(b'whatever')
+    return path
+
+
+def test_get_existing_local(dst_path_exists, src_path, ssh_scp, transmit_payload):
```
webknjaz:
```suggestion
def test_get_existing_local(pre_existing_file_path, src_path, ssh_scp, transmit_payload):
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    """Check that SCP file download works and overrides local file if it exists."""
+    ssh_scp.get(str(src_path), str(dst_path_exists))
```
webknjaz:
```suggestion
    ssh_scp.get(str(src_path), str(pre_existing_file_path))
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    assert dst_path_exists.read_bytes() == transmit_payload
```
webknjaz:
```suggestion
    assert pre_existing_file_path.read_bytes() == transmit_payload
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
     """Check that SCP file download works."""
     ssh_scp.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def pre_existing_file_path(tmp_path):
+    """Return a local path that exists."""
```
webknjaz:
```suggestion
    """Return local path for a pre-populated file."""
```

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +133,17 @@ cdef class SCP:
...
-            rc = libssh.ssh_scp_read(scp, read_buffer, size)
-            if rc == libssh.SSH_ERROR:
-                raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
-
-            py_file_bytes = read_buffer[:size]
             with open(local_file, "wb") as f:
-                f.write(py_file_bytes)
+                remaining_bytes_to_read = size
+                while remaining_bytes_to_read > 0:
+                    # Read the file
```
webknjaz: Comments repeating what the code does are pointless. Proper identifier names should be used instead.
```suggestion
```
Learning materials: https://www.youtube.com/watch?v=YklKUuDpX5c / https://www.youtube.com/watch?v=x-kB2o8sd5c / https://www.youtube.com/watch?v=1__lNTlj1_w.

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +133,17 @@ cdef class SCP:
...
+                    read_bytes = libssh.ssh_scp_read(scp, read_buffer, min(remaining_bytes_to_read, read_buffer_size))
+                    if read_bytes == libssh.SSH_ERROR:
+                        raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
+
+                    py_file_bytes = read_buffer[:read_bytes]
+                    f.write(py_file_bytes)
+                    remaining_bytes_to_read -= read_bytes
```
webknjaz, line 149: I believe that this would be more correct:
```suggestion
                    remaining_bytes_to_read -= len(py_file_bytes)
```
The reason is that the slice operator may return fewer bytes, then you think in case the actual iterable is smaller:
```pycon
>>> a_list = [2, 3, 4]
>>> a_list
[2, 3, 4]
>>> a_list[:0]
[]
>>> a_list[:1]
[2]
>>> a_list[:2]
[2, 3]
>>> a_list[:3]
[2, 3, 4]
>>> a_list[:4]
[2, 3, 4]
>>> a_list[:5]
[2, 3, 4]
>>> a_list[:100500]
[2, 3, 4]
```

src/pylibsshext/scp.pyx
```diff
@@ -131,14 +133,17 @@ cdef class SCP:
...
 
-            # Read the file
-            rc = libssh.ssh_scp_read(scp, read_buffer, size)
-            if rc == libssh.SSH_ERROR:
-                raise LibsshSCPException("Error receiving file data: %s" % self._get_ssh_error_str())
-
-            py_file_bytes = read_buffer[:size]
             with open(local_file, "wb") as f:
-                f.write(py_file_bytes)
+                remaining_bytes_to_read = size
```
webknjaz: While on it, let's move this var out to reduce the indentation:
```suggestion
            remaining_bytes_to_read = size
            with open(local_file, "wb") as f:
```

src/pylibsshext/scp.pyx
```diff
@@ -122,7 +122,9 @@ cdef class SCP:
             size = libssh.ssh_scp_request_get_size(scp)
             mode = libssh.ssh_scp_request_get_permissions(scp)
 
-            read_buffer = <char *>PyMem_Malloc(size)
+            # cap the buffer size to reasonable number -- libssh will not return the whole data at once anyway
+            read_buffer_size = min(size, 65536)
```
webknjaz: Let's name this piece of data (though, I'd probably even move it into a module-level constant):
```suggestion
            default_read_buffer_size = 65536
            read_buffer_size = min(size, default_read_buffer_size)
```

Is there a constant within libssh that we could reuse instead of re-hardcoding it here?

webknjaz: Is it possible to rely on this https://github.com/canonical/libssh/blob/6f4ec9be/src/connector.c#L33?

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    ssh_scp.get(str(src_path), str(pre_existing_file_path))
+    assert pre_existing_file_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large 65537 byte (64kB+1B) test payload."""
+    random_char_kilobyte = [ord(random.choice(string.printable)) for _ in range(1024)]
+    full_bytes_number = 64
+    return bytes(random_char_kilobyte * full_bytes_number) + random_char_kilobyte[0].to_bytes(1)
```
webknjaz:
```suggestion
    return bytes(random_char_kilobyte * full_bytes_number) + random_char_kilobyte[0].to_bytes(length=1, byteorder='big')
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,44 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    ssh_scp.get(str(src_path), str(pre_existing_file_path))
+    assert pre_existing_file_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large 65537 byte (64kB+1B) test payload."""
+    random_char_kilobyte = [ord(random.choice(string.printable)) for _ in range(1024)]
+    full_bytes_number = 64
+    return bytes(random_char_kilobyte * full_bytes_number) + random_char_kilobyte[0].to_bytes(length=1, byteorder='big')
```
webknjaz:
```suggestion
    a_64kB_chunk = bytes(random_char_kilobyte * full_bytes_number)
    the_last_byte = random_choice(random_char_kilobyte).to_bytes(length=1, byteorder='big')
    return a_64kB_chunk + the_last_byte
```

tests/unit/scp_test.py
```diff
@@ -57,3 +59,46 @@ def test_get(dst_path, src_path, ssh_scp, transmit_payload):
...
+    assert pre_existing_file_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large 65537 byte (64kB+1B) test payload."""
+    random_char_kilobyte = [ord(random.choice(string.printable)) for _ in range(1024)]
+    full_bytes_number = 64
+    a_64kB_chunk = bytes(random_char_kilobyte * full_bytes_number)
+    the_last_byte = random_choice(random_char_kilobyte).to_bytes(length=1, byteorder='big')
```
webknjaz:
```suggestion
    the_last_byte = random.choice(random_char_kilobyte).to_bytes(length=1, byteorder='big')
```

webknjaz:
@Jakuje hey, so I added a few improvements on top and posted a few more comments inline. I've now stopped pushing to the PR branch so feel free to clean it up/rebase/make more changes.

docs/changelog-fragments/621.bugfix.rst
```diff
@@ -0,0 +1 @@
+Downloading files larger thank 64kB over SCP no longer fails -- by :user:`Jakuje`.
```
webknjaz: just noticed a typo
```suggestion
Downloading files larger than 64kB over SCP no longer fails -- by :user:`Jakuje`.
```

src/pylibsshext/scp.pyx
Jakuje: There is no constant for this. This one is for reading different chunks in different part of the protocol. The SCP has this hardcoded and given that we are deprecating SCP I do not expect this will ever change:

https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/scp.c#L1025

Jakuje: I think we will have eventually the same problem with writing (but that is likely much less used in the context of ansible) so I will make a separate constant for this .

src/pylibsshext/scp.pyx
Jakuje: Not sure if this is an issue as if we could not trust the consistency of the returned value `read_bytes` and the amount of read bytes, we are screwed anyway.

Given that we asked to read maximum of `read_buffer_size` and the buffer should have the same size we should not end up in the situation you describe. And if we do, we would create a different file that was on remote anyway so I would rather propose some defensive sanity check where we would fail instead of trying to save it with something like this. Committed a sanity check for the returned size that it is in the bounds we expect.

src/pylibsshext/scp.pyx
Jakuje: Removed.

src/pylibsshext/scp.pyx
Jakuje: Applied. Thanks!

src/pylibsshext/scp.pyx
webknjaz: @Jakuje I see you added a check for getting *more* bytes which I suspect would be a protocol violation. But I was talking about the case when it returns *fewer* bytes, but we'd be removing more bytes than we actually got from the counter.

Imagine we have a file size of 5 and a buffer of 5 but the first invocation of `ssh_scp_read()` does not fill out the entire buffer and only receives 4 bytes. But we still set `read_buffer_size = read_buffer_size - buffer = 5 - 5 = 0` so after the next iteration never happens, and we never attempt getting that last byte.

webknjaz:
This needs conflict resolution now.

src/pylibsshext/scp.pyx
Jakuje: I got that, but I still do not see how this could happen with the current code nor how the proposed fix would fix that. We are subtracting the `read_bytes` from the `remaining_bytes_to_read` and that should always be equal to the `len(py_file_bytes)` unless the `read_bytes` would be larger than the `read_buffer_size`. Did you mistake the `read_buffer_size` and `read_bytes` variables?

Jakuje:
Resolved the conflict and dropped the check for too large replies.

src/pylibsshext/scp.pyx
webknjaz: Yeah, I got confused. I suppose `len(read_buffer[:read_bytes]) == read_bytes` is always true. Marking this thread as resolved.

webknjaz (approved):
Thanks!

webknjaz:
Looks like one of the CI jobs got stuck with https://github.com/ansible/pylibssh/issues/557 so I restarted it: https://github.com/ansible/pylibssh/actions/runs/9699687288/job/26769948869?pr=621#step:15:152.

webknjaz:
@Jakuje this has been released as a part of v1.2.2: https://github.com/ansible/pylibssh/releases/tag/v1.2.2 / https://github.com/ansible/pylibssh/discussions/626 / https://pypi.org/project/ansible-pylibssh/1.2.2/.

Enjoy! 🎉

PR #636, opened 2024-08-29
Bump libssh to v0.11.1 in `manylinux` build infra (includes #734)

Jakuje:
##### SUMMARY
This is an attempt to bump libssh version to 0.11.0, which is a prerequisite to improve the SFTP throughput.

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION
I am not sure if it would make sense to run the test matrix on some older libssh versions too (as libssh will not get updated in el9 for example) so keeping as a draft for now to address remaining questions before merging.

The version 0.11.1 was just released, which should be already quite stable for general use:

https://www.libssh.org/2024/08/30/libssh-0-11-1-release/


docs/changelog-fragments/636.feature.rst
```diff
@@ -0,0 +1 @@
+Bumped the libssh version to latest release 0.11.1 -- by :user:`Jakuje`.
```
webknjaz: Perhaps, mention that it's for manylinux? macOS is pulling the latest version available in Homebrew, currently: https://github.com/ansible/pylibssh/actions/runs/10801441579/job/29961681318#step:5:212. I've been meaning to figure out compiling it in a more controllable way, similar to manylinux but never got to it...

Jakuje: Thanks! Updated.

webknjaz:
Seeing that `manylinux1` cannot build this new version of `libssh`, it would make sense to deal with #324 first.

tox.ini
```diff
@@ -258,7 +258,7 @@ commands =
     -i --rm \
     -v {toxinidir}:/io \
     -e ANSIBLE_PYLIBSSH_CYTHON_TRACING \
-    ghcr.io/ansible/pylibssh-manylinux{env:MANYLINUX_VERSION_TAG}_{env:MANYLINUX_ARCH_TAG}:libssh-v{env:LIBSSH_VERSION:0.9.6} \
+    ghcr.io/ansible/pylibssh-manylinux{env:MANYLINUX_VERSION_TAG}_{env:MANYLINUX_ARCH_TAG}:libssh-v{env:LIBSSH_VERSION:0.11.1} \
```
webknjaz: @Jakuje the new place where this is defined is https://github.com/ansible/pylibssh/blob/80bccfe/pyproject.toml#L164-L167. Take it into account when rebasing.

Although, the CI will be failing since there are some unsolved issues with the base containers that I haven't had a chance to troubleshoot properly. Especially, with non-x86 arches (my local QEMU setup is broken).

Jakuje: Yes. I saw that already, but I did not see a lot of value in pushing it right now. But I can do that to keep it up to date.

Jakuje:
Looking at the manylinux builds, I see that you build OpenSSL from source, but version 1.1.1k, which is quite dated. Would it make sense to build some newer too?

Looking at what is failing, its the libffi build. There is version 3.3, which is also 5 years old. So I will try if the new version will do better.

webknjaz:
> Looking at the manylinux builds, I see that you build OpenSSL from source, but version 1.1.1k, which is quite dated. Would it make sense to build some newer too?

Maybe. That hasn't been updated in a while.

> Looking at what is failing, its the libffi build. There is version 3.3, which is also 5 years old. So I will try if the new version will do better.

That'd be a start. I'd try it myself if my local QEMU wasn't broken.

--------

For historic context, originally the `Dockerfile` and things it includes are coming from https://github.com/pyca/infra. We used to just inherit images that the PyCA folks were posting for their own CI/CD: https://github.com/orgs/pyca/packages?ecosystem=container&tab=packages&ecosystem=container&q=manylinux. And the sources for said images are in that repo. Over time, they dropped `manylinux1` and deleted the images, and they got rid of various bits of automation around that. So I've made a branch in my fork recovering some of their older commits while including their newer manylinux bits: https://github.com/webknjaz/pyca-infra/tree/bugfixes/manylinux1. That was merged into this repo's https://github.com/ansible/pylibssh/tree/baa2ab8/build-scripts/manylinux-container-image. And when the PyCA infra got new things, I'd update the branch in my fork preserving older hacks. And the fork would be merged into this repo (well, just the corresponding directory).

And on top of that, we had a few other things like pre-building libssh once so that normal CI runs wouldn't waste CPU on that. This is especially impactful in non-x86 jobs because QEMU is slow.
I think that over time @Qalthos attempted to simplify things around the container configuration. And in doing so, made it diverge from the upstreams. Probably not realizing that I'd originally wanted to be as close to the upstream as possible and only add minimal patches on top: https://github.com/ansible/pylibssh/tree/baa2ab8/build-scripts/manylinux-container-image#manylinux-images.
I haven't looked into synching with PyCA in a long time. They've also diverged noticeably.

It is possible that we may not need all of the things included, but somebody needs to be able to experiment with that to decide.

Since forking, our container directly inherits the PyPA-provided ones. And it's always `latest`. So something's updated in the upstream containers and made our container builds fail. Pinning to a hash from the times it worked is probably the easiest workaround short-term.

webknjaz:
The last successful build was https://github.com/ansible/pylibssh/actions/runs/12858475891 on 2025-01-20T01:17:01+01:00 / Jan 20, 2025, 1:17 AM GMT+1.

And the following failing build was https://github.com/ansible/pylibssh/actions/runs/12979659724 on 2025-01-27T01:17:12+01:00 / Jan 27, 2025, 1:17 AM GMT+1.

webknjaz:
So likely `sha256:3394f5f0684d7c22313ef435eba96cb37c11c9d1f6e5c4c8c7aff06983c02b23` (tags: `2025.01.24-1` / `2025-01-24-7eee2d4`) was what broke it. *(the version refs here are for `pypa/manylinux_2_28_aarch64`)*

webknjaz:
Opened https://github.com/ansible/pylibssh/pull/697 to test this theory.

webknjaz:
> Opened #697 to test this theory.

I tested that tag against the aarch64 job and it segfaulted on attempting to build openssl: https://github.com/ansible/pylibssh/actions/runs/14148431217/job/39638625616?pr=697#step:4:631. I wonder what else's changed...

webknjaz:
Good news is that we can actually retrieve digests of the last built containers from GHCR. For example, scrolling to the bottom of https://github.com/ansible/pylibssh/pkgs/container/pylibssh-manylinux_2_28_s390x/339542363?tag=latest

shows
```json
  "labels": {
    "org.opencontainers.image.base.digest": "sha256:af02a43280267238b7e21b929a5a0b5a091d0de7737d450a554907eb99d8332c",
    "org.opencontainers.image.base.name": "quay.io/pypa/manylinux_2_28_s390x:latest"
  }
```
at the very end, meaning that we can say that the base containter was exactly `quay.io/pypa/manylinux_2_28_s390x:af02a43280267238b7e21b929a5a0b5a091d0de7737d450a554907eb99d8332c`, at that was apparently `quay.io/pypa/manylinux_2_28_s390x:2025.01.19-2`

webknjaz:
Same on the earlier container version: https://github.com/ansible/pylibssh/actions/runs/14148529357/job/39638839231?pr=697#step:4:536.

I wonder if the version of buildah changed or the host VM updates somehow influenced QEMU.

webknjaz:
So `ubuntu-latest` completed transition from 22.04 to 24.04 a bit earlier, on Jan 17: https://github.com/actions/runner-images/issues/10636. Could this be related?

webknjaz:
Downgrading the runner VM to Ubuntu 20.04 hasn't helped. I'm not sure what else to try given the inability to troubleshoot locally...

build-scripts/manylinux-container-image/install_libffi.sh
```diff
@@ -9,8 +9,8 @@ MY_DIR=$(dirname "${BASH_SOURCE[0]}")
 # Get build utilities
 source $MY_DIR/build_utils.sh
 
-LIBFFI_SHA256="72fba7922703ddfa7a028d513ac15a85c8d54c8d67f55fa5a4802885dc652056"
-LIBFFI_VERSION="3.3"
+LIBFFI_SHA256="f07c08c9c14977eafb9b5f9277713d91358ec18fc8aaa5607d6790cde90cba12"
+LIBFFI_VERSION="3.4.7"
 
 fetch_source "libffi_${LIBFFI_VERSION}.orig.tar.gz" "https://mirrors.ocf.berkeley.edu/debian/pool/main/libf/libffi"
```
webknjaz: This should switch to https://github.com/libffi/libffi/releases/download/v3.4.8/libffi-3.4.8.tar.gz

webknjaz:
@Jakuje looks like you bumped the version but didn't switch the URL to point to GitHub Releases?

webknjaz:
Oh, I see. The old host has a new version, but not the previous one. We may still want to change that URL to be future-proof, but for now, it's best to declare it out of the scope of this PR.

Now, the failure seems to be missing `makeinfo` in the image:

```console
make[2]: Entering directory `/root/libffi-3.4.8/x86_64-pc-linux-gnu/doc'
restore=: && backupdir=".am$$" && \
	am__cwd=`pwd` && CDPATH="${ZSH_VERSION+.}:" && cd ../../doc && \
	rm -rf $backupdir && mkdir $backupdir && \
	if (/bin/sh '/root/libffi-3.4.8/missing' makeinfo --version) >/dev/null 2>&1; then \
	  for f in ../../doc/libffi.info ../../doc/libffi.info-[0-9] ../../doc/libffi.info-[0-9][0-9] ../../doc/libffi.i[0-9] ../../doc/libffi.i[0-9][0-9]; do \
	    if test -f $f; then mv $f $backupdir; restore=mv; else :; fi; \
	  done; \
	else :; fi && \
	cd "$am__cwd"; \
	if /bin/sh '/root/libffi-3.4.8/missing' makeinfo   -I ../../doc \
	 -o ../../doc/libffi.info ../../doc/libffi.texi; \
	then \
	  rc=0; \
	  CDPATH="${ZSH_VERSION+.}:" && cd ../../doc; \
	else \
	  rc=$?; \
	  CDPATH="${ZSH_VERSION+.}:" && cd ../../doc && \
	  $restore $backupdir/* `echo "./../../doc/libffi.info" | sed 's|[^/]*$||'`; \
	fi; \
	rm -rf $backupdir; exit $rc
/root/libffi-3.4.8/missing: line 81: makeinfo: command not found
WARNING: 'makeinfo' is missing on your system.
         You should only need it if you modified a '.texi' file, or
         any other file indirectly affecting the aspect of the manual.
         You might want to install the Texinfo package:
         <https://www.gnu.org/software/texinfo/>
         The spurious makeinfo call might also be the consequence of
         using a buggy 'make' (AIX, DU, IRIX), in which case you might
         want to install GNU make:
         <https://www.gnu.org/software/make/>
make[2]: *** [../../doc/libffi.info] Error 127
make[2]: Leaving directory `/root/libffi-3.4.8/x86_64-pc-linux-gnu/doc'
make[1]: *** [install-recursive] Error 1
make[1]: Leaving directory `/root/libffi-3.4.8/x86_64-pc-linux-gnu'
make: *** [install] Error 2
Error: building at STEP "RUN ./install_libffi.sh "${RELEASE}"": while running runtime: exit status 2
Error: Error: buildah exited with code 2
```

Jakuje:
> @Jakuje looks like you bumped the version but didn't switch the URL to point to GitHub Releases?

The URL still downloads from the mirror where it downloaded from previously. Is there an advantage to use github releases instead of the current mirror?

Jakuje:
> Now, the failure seems to be missing `makeinfo` in the image:

Missed that part. What would be a good place to install it? Directly in build-scripts/manylinux-container-image/install_libffi.sh or in Dockerfile?

docs/changelog-fragments/636.contrib.rst
```diff
@@ -0,0 +1 @@
+Bumped the libssh version to latest release 0.11.1 in manylinux -- by :user:`Jakuje`.
```
webknjaz: Let's try this
```suggestion
Updated the version of ``libssh`` to the latest release v0.11.1
in the cached ``manylinux`` build environment container images
-- by :user:`Jakuje`.
```

Jakuje: Adjusted.

build-scripts/manylinux-container-image/install_libffi.sh
Jakuje: Switched.

webknjaz:
@Jakuje another rebase is needed?

Jakuje:
> @Jakuje another rebase is needed?

yeah. my bad. Fixed now.

PR #638, opened 2024-08-30
Fix large SFTP reads

### LINKED ISSUE #341: sftp.get downloads only the last chunk of file
##### SUMMARY

I'm trying to download configuration file from remote router and getting only last chunk of the file. 

##### ISSUE TYPE
- Bug Report

```paste below
0.4.0
```

##### OS / ENVIRONMENT

Ubuntu 22.04 LTS
Python 3.10.4

##### STEPS TO REPRODUCE

Narrowing the problem I wrote a piece of code below.
You need a file on remote host to reproduce it.

```paste below
from pylibsshext.session import Session, AutoAddPolicy

hssh = Session()
p = AutoAddPolicy()
hssh.set_missing_host_key_policy(p)

HOST = 'hostname'
USER = 'username'
PASSWORD = 'password'
TIMEOUT = 30
PORT = 22

hssh.connect(
    host=HOST,
    user=USER,
    password=PASSWORD,
    timeout=TIMEOUT,
    port=PORT,
)
sftp = hssh.sftp()
sftp.get('m8.rsc', 'm8.rsc')
```

##### EXPECTED RESULTS

File 'm8.rsc' in current directory, size of 16K

##### ACTUAL RESULTS

File 'm8.rsc', containing last 223b of source file.

```paste below

```
As I found out, the problem lays in line #103 of sftp.pyx
```
           with open(local_file, 'wb+') as f:
```
In python3, "w" parameter truncates the file, so during reading the source file new data overwrites old.
Please check and consider changing open mode to 'ab'

Jakuje:
##### SUMMARY
Fixes and reproducers for large SFTP reads that truncate file with the second chunk.

Fixes: #341

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/638.bugfix.rst
+++ b/docs/changelog-fragments/638.bugfix.rst
@@ -1,0 +1,5 @@
+Fixed reading files over SFTP that go over the pre-defined chunk size.
+
+Prior to this change, the files could end up being corrupted, ending up with the last read chunk written to the file instead of the entire payload.
+
+-- by :user:`Jakuje`
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -100,7 +100,7 @@ cdef class SFTP:
                 raise LibsshSFTPException("Reading data from remote file [%s] failed with error [%s]"
                                           % (remote_file, self._get_sftp_error_str()))
 
-            with open(local_file, 'wb+') as f:
+            with open(local_file, 'ab') as f:
                 bytes_written = f.write(read_buffer[:file_data])
                 if bytes_written and file_data != bytes_written:
                     sftp.sftp_close(rf)
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -91,16 +91,16 @@ cdef class SFTP:
         if rf is NULL:
             raise LibsshSFTPException("Opening remote file [%s] for read failed with error [%s]" % (remote_file, self._get_sftp_error_str()))
 
-        while True:
-            file_data = sftp.sftp_read(rf, <void *>read_buffer, sizeof(char) * 1024)
-            if file_data == 0:
-                break
-            elif file_data < 0:
-                sftp.sftp_close(rf)
-                raise LibsshSFTPException("Reading data from remote file [%s] failed with error [%s]"
-                                          % (remote_file, self._get_sftp_error_str()))
+        with open(local_file, 'wb') as f:
+            while True:
+                file_data = sftp.sftp_read(rf, <void *>read_buffer, sizeof(char) * 1024)
+                if file_data == 0:
+                    break
+                elif file_data < 0:
+                    sftp.sftp_close(rf)
+                    raise LibsshSFTPException("Reading data from remote file [%s] failed with error [%s]"
+                                              % (remote_file, self._get_sftp_error_str()))
 
-            with open(local_file, 'ab') as f:
                 bytes_written = f.write(read_buffer[:file_data])
                 if bytes_written and file_data != bytes_written:
                     sftp.sftp_close(rf)
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -87,6 +87,12 @@ def src_path_large(tmp_path, large_payload):
     return path
 
 
+def test_put_large(dst_path, src_path_large, sftp_session, large_payload):
+    """Check that SFTP can upload large file."""
+    sftp_session.put(str(src_path_large), str(dst_path))
+    assert dst_path.read_bytes() == large_payload
+
+
 def test_get_large(dst_path, src_path_large, sftp_session, large_payload):
     """Check that SFTP can download large file."""
     sftp_session.get(str(src_path_large), str(dst_path))
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -2,6 +2,8 @@
 
 """Tests suite for sftp."""
 
+import random
+import string
 import uuid
 
 import pytest
@@ -63,3 +65,29 @@ def test_get(dst_path, src_path, sftp_session, transmit_payload):
     """Check that SFTP file download works."""
     sftp_session.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
+    """Generate a large 1025 byte (1024 + 1B) test payload."""
+    payload_len = 1024 + 1
+    random_bytes = [ord(random.choice(string.printable)) for _ in range(payload_len)]
+    return bytes(random_bytes)
+
+
+@pytest.fixture
+def src_path_large(tmp_path, large_payload):
+    """Return a remote path to a 1025 byte-sized file.
+
+    The pylibssh chunk size is 1024 so the test needs a file that would
+    execute at least two loops.
+    """
+    path = tmp_path / 'large.txt'
+    path.write_bytes(large_payload)
+    return path
+
+
+def test_get_large(dst_path, src_path_large, sftp_session, large_payload):
+    """Check that SFTP can download large file."""
+    sftp_session.get(str(src_path_large), str(dst_path))
+    assert dst_path.read_bytes() == large_payload
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -88,6 +88,12 @@ def test_get_existing(pre_existing_dst_path, src_path, sftp_session, transmit_pa
     assert pre_existing_dst_path.read_bytes() == transmit_payload
 
 
+def test_put_existing(pre_existing_dst_path, src_path, sftp_session, transmit_payload):
+    """Check that SFTP file upload works when target file exists."""
+    sftp_session.put(str(src_path), str(pre_existing_dst_path))
+    assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
 @pytest.fixture
 def large_payload():
     """Generate a large 1025 byte (1024 + 1B) test payload."""
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -50,6 +50,21 @@ def dst_path(file_paths_pair):
     return path
 
 
+@pytest.fixture
+def other_payload():
+    """Generate a binary test payload."""
+    uuid_name = uuid.uuid4()
+    return 'Original content: {name!s}'.format(name=uuid_name).encode()
+
+
+@pytest.fixture
+def pre_existing_dst_path(dst_path, other_payload):
+    """Return a data destination path."""
+    dst_path.write_bytes(other_payload)
+    assert dst_path.exists()
+    return dst_path
+
+
 def test_make_sftp(sftp_session):
     """Smoke-test SFTP instance creation."""
     assert sftp_session
@@ -65,6 +80,12 @@ def test_get(dst_path, src_path, sftp_session, transmit_payload):
     """Check that SFTP file download works."""
     sftp_session.get(str(src_path), str(dst_path))
     assert dst_path.read_bytes() == transmit_payload
+
+
+def test_get_existing(pre_existing_dst_path, src_path, sftp_session, transmit_payload):
+    """Check that SFTP file download works when target file exists."""
+    sftp_session.get(str(src_path), str(pre_existing_dst_path))
+    assert pre_existing_dst_path.read_bytes() == transmit_payload
 
 
 @pytest.fixture
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -20,11 +20,21 @@ def sftp_session(ssh_client_session):
         del sftp_sess  # noqa: WPS420
 
 
-@pytest.fixture
-def transmit_payload():
-    """Generate a binary test payload."""
-    uuid_name = uuid.uuid4()
-    return 'Hello, {name!s}'.format(name=uuid_name).encode()
+@pytest.fixture(
+    params=(32, 1024 + 1),
+    ids=('small-payload', 'large-payload'),
+)
+def transmit_payload(request: pytest.FixtureRequest) -> bytes:
+    """Generate binary test payloads of assorted sizes.
+
+    The choice 32 is arbitrary small value.
+
+    The choice 1024 + 1 is meant to be 1B larger than the chunk size used in
+    :file:`sftp.pyx` to make sure we excercise at least two rounds of reading/writing.
+    """
+    payload_len = request.param
+    random_bytes = [ord(random.choice(string.printable)) for _ in range(payload_len)]
+    return bytes(random_bytes)
 
 
 @pytest.fixture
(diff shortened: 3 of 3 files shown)


src/pylibsshext/sftp.pyx
```diff
@@ -100,7 +100,7 @@ cdef class SFTP:
                 raise LibsshSFTPException("Reading data from remote file [%s] failed with error [%s]"
                                           % (remote_file, self._get_sftp_error_str()))
 
-            with open(local_file, 'wb+') as f:
+            with open(local_file, 'ab') as f:
```
kucharskim: This will append content to the file which already exists. I've tested this diff with `ansible-libssh v1.2.2` and only a60a6cbf0ea4a7ef0e25299a1b3d01509488c6b5 applied. I see multiple executions on the same destination file growing above original size:

```
remote-machine# ls -l /tmp/foo?
-rw-r--r--  1 root  wheel  1048576 Nov 15 12:31 /tmp/foo1
-rwx------  1 root  wheel  3145728 Nov 15 12:37 /tmp/foo3
```

```
local-machine$ ls -l /tmp/foo?
-rw-r--r--  1 mkucharski  wheel  3145728 Nov 15 12:35 /tmp/foo2
```

Above we see size 3x after 3 executions of:
```
t1 = sftp_session.get(remote_file, local_file)
t2 = sftp_session.put(local_file, new_remote_file)
```

I think this needs more work. Also mode doesn't match after `put()`, but that is another issue.

Jakuje: Thank you for throughful testing! Indeed, you are right. I thought this small change will make it working, but it looks like its more complicated. Updated the PR.

For the modes, I think this would be a different issue or rfe. Right now, the code opens the file with normal python open, which will probably set modes based on your umask, which will result in sane permissions. We could update the permissions based on what is on the server (if the server uses unix/linux ACL), but given that they will likely have different usernames and groups, it might result in unwanted permissions.

kucharskim:
This version works, but it is much slower comparing to SCP version.

```
$ time python3 test_sftp1.py
ssh.is_connected=1
<pylibsshext.sftp.SFTP object at 0xd2daf8eb910>
get
None
put
None
Closing sftp session...
Closing connection...
    1m19.09s real     0m00.16s user     0m00.21s system
```

```
$ time python3 test_scp1.py
ssh.is_connected=1
<pylibsshext.scp.SCP object at 0x7bec33ef820>
get
0
put
0
Closing connection...
    0m03.59s real     0m00.17s user     0m00.13s system
```

Above we see approx 80 seconds vs 4 seconds.

I've bumped all `1024` instances in `sftp.pyx` to `64 * 1024` like here (in `put()` and in `get()`):
```
-            buffer = f.read(1024)
+            buffer = f.read(1024 * 64)
```

and this improved time significantly:
```
$ time python3 test_sftp1.py
ssh.is_connected=1
<pylibsshext.sftp.SFTP object at 0x44fda57910>
get
None
put
None
Closing sftp session...
Closing connection...
    0m05.36s real     0m00.17s user     0m00.11s system
```

Jakuje:
Yes, your observations are right. Buffer of 16k is recommended by the the libssh documentation to have some reasonable performance:

https://api.libssh.org/master/libssh_tutor_sftp.html

I can change it as part of this PR or separate, but the PRs are starting to depend on each other, making the changes hard.

Jakuje:
Using 64k chunks might be problematic as the SSH specification suggests maximum packet size. Most of the implementation work with larger ones, but if they do not, it causes hard to debug issues, such as https://github.com/curl/curl/pull/11804. I would go ahead with 32k as libssh will split large chunks anyway to this size.

kucharskim:
Sure, whatever works. Just to make sftp faster.

Jakuje:
> Sure, whatever works. Just to make sftp faster.

If you want to get even better performance, you can try #641. Its still in draft as it requires libssh 0.11 and I am not sure how to support at the same time also old version, but it should get even larger speeds.

Jakuje:
Filled #664 for increasing the chunk size (on top of this PR).

tests/unit/sftp_test.py
```diff
@@ -63,3 +81,47 @@ def test_get(dst_path, src_path, sftp_session, transmit_payload):
...
+
+
+def test_get_existing(dst_exists_path, src_path, sftp_session, transmit_payload):
+    """Check that SFTP file download works when target file exists."""
+    sftp_session.get(str(src_path), str(dst_exists_path))
+    assert dst_exists_path.read_bytes() == transmit_payload
+
+
+def test_put_existing(dst_exists_path, src_path, sftp_session, transmit_payload):
+    """Check that SFTP file download works when target file exists."""
```
webknjaz, line 96:
```suggestion
    """Check that SFTP file upload works when target file exists."""
```

tests/unit/sftp_test.py
```diff
@@ -63,3 +81,47 @@ def test_get(dst_path, src_path, sftp_session, transmit_payload):
...
+    sftp_session.put(str(src_path), str(dst_exists_path))
+    assert dst_exists_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def large_payload():
```
webknjaz: I think `other_payload` and `large_payload` could be turned into a single parametrized fixture.

tests/unit/sftp_test.py
Jakuje: Fixed, thank you!

tests/unit/sftp_test.py
Jakuje: Updated to parametrized fixture. Hopefully the way you had in your mind.

tests/unit/sftp_test.py
```diff
@@ -48,6 +50,23 @@ def dst_path(file_paths_pair):
     return path
 
 
+@pytest.fixture(params=[32, 1025])
+def sized_payload():
+    """Generate a binary test payload of size given as a parameter."""
+    payload_len = 1024 + 1
```
webknjaz: It's incomplete. 1024 is still hardcoded. It needs to be retrieved from the `request` fixture. Additionally, it's a good idea to always set IDs so that the semantic meaning would be clear.

Also, renaming it to `transmit_payload` will turn all of the tests using it into parametrized. So there's no need for a separate `*_large` copies of the same logic:
```suggestion
@pytest.fixture(
    params=(32, 1024 + 1),
    ids=('small-payload', 'large-payload'),
)
def transmit_payload(request: pytest.FixtureRequest) -> None:
    """Generate binary test payloads of assorted sizes."""
    payload_len = request.param
```

tests/unit/sftp_test.py
```diff
@@ -63,3 +82,39 @@ def test_get(dst_path, src_path, sftp_session, transmit_payload):
...
+def test_put_large(dst_path, src_path_large, sftp_session, sized_payload):
+    """Check that SFTP can upload large file."""
+    sftp_session.put(str(src_path_large), str(dst_path))
+    assert dst_path.read_bytes() == sized_payload
+
+
+def test_get_large(dst_path, src_path_large, sftp_session, sized_payload):
+    """Check that SFTP can download large file."""
+    sftp_session.get(str(src_path_large), str(dst_path))
+    assert dst_path.read_bytes() == sized_payload
```
webknjaz: With https://github.com/ansible/pylibssh/pull/638/files#r1927502223, this won't be necessary:
```suggestion
```

tests/unit/sftp_test.py
```diff
@@ -2,6 +2,8 @@
 
 """Tests suite for sftp."""
 
+import random
+import string
 import uuid
```
webknjaz, line 7: With https://github.com/ansible/pylibssh/pull/638/files#r1927502223, this import is no longer needed:
```suggestion
```

tests/unit/sftp_test.py
webknjaz, line 1: Need to drop the old `transmit_payload` when implementing the new one per https://github.com/ansible/pylibssh/pull/638/files#r1927502223.

src/pylibsshext/sftp.pyx
```diff
@@ -89,18 +89,19 @@ cdef class SFTP:
...
-            if file_data == 0:
-                break
-            elif file_data < 0:
-                sftp.sftp_close(rf)
-                raise LibsshSFTPException("Reading data from remote file [%s] failed with error [%s]"
-                                          % (remote_file, self._get_sftp_error_str()))
-
-            with open(local_file, 'wb+') as f:
+            raise LibsshSFTPException("Opening remote file [%s] for read failed with error [%s]"
+                                      % (remote_file, self._get_sftp_error_str()))
```
webknjaz: Is there any difference here other than formatting? I can't spot it...

docs/changelog-fragments/638.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed large sftp reads and proper overriding existing files -- by :user:`Jakuje`.
```
webknjaz: I'm trying to come up with something that could be understood by the end-users easily in terms of the impact on how they use the project.

Would this be more accurate? (The pre-existing file reference is unclear — I tried to infer this from the PR description and the linked issue, but it doesn't seem like there's a direct influence here)
```suggestion
Fixed reading files over SFTP that go over the pre-defined chunk size.

Prior to this change, the files could end up being corrupted, ending up with the last read chunk written to the file instead of the entire payload.

-- by :user:`Jakuje`
```

tests/unit/sftp_test.py
Jakuje: Oh, my bad. Missed that part. Thanks for guidance!

src/pylibsshext/sftp.pyx
Jakuje: No difference except formatting. I think the formatter complained about line length. I can revert this if needed.

docs/changelog-fragments/638.bugfix.rst
Jakuje: This looks good. I used your wording in updated PR.

tests/unit/sftp_test.py
Jakuje: I kept the uuid in the `other_payload()` as I would need the payload for existing files.

docs/changelog-fragments/638.bugfix.rst
```diff
@@ -0,0 +1,5 @@
+Fixed reading files over SFTP that go over the pre-defined chunk size.
+
+Prior to this change, the files could end up being corrupted, ending up with the last read chunk written to the file instead of the entire payload.
+
+-- by :user:`Jakuje`.
```
webknjaz: Detached bylines don't need a trailing period.
```suggestion
-- by :user:`Jakuje`
```

tests/unit/sftp_test.py
```diff
@@ -18,11 +20,21 @@ def sftp_session(ssh_client_session):
...
+    params=(32, 1024 + 1),
+    ids=('small-payload', 'large-payload'),
+)
+def transmit_payload(request: pytest.FixtureRequest):
+    """Generate binary test payloads of assorted sizes.
+
+    The choice 32 is arbitrary small value.
+
+    The choice 1024 + 1 is meant to be 1B larger than the chunk size used in
+    sftp.pyx to make sure we excercise at least two rounds of reading/writing.
```
webknjaz: It's usually a good idea to use RST syntax in docstrings.
```suggestion
    :file:`sftp.pyx` to make sure we excercise at least two rounds of reading/writing.
```

tests/unit/sftp_test.py
```diff
@@ -48,6 +60,22 @@ def dst_path(file_paths_pair):
     return path
 
 
+@pytest.fixture
+def other_payload():
+    """Generate a binary test payload."""
+    uuid_name = uuid.uuid4()
+    return 'Original content: {name!s}'.format(name=uuid_name).encode()
```
webknjaz, line 67: The range of supported Pythons is modern enough to let us just use f-strings here:
```suggestion
    return f'Original content: {uuid_name !s}'.encode()
```

tests/unit/sftp_test.py
```diff
@@ -48,6 +60,22 @@ def dst_path(file_paths_pair):
...
+
+
+@pytest.fixture
+def dst_exists_path(file_paths_pair, other_payload):
+    """Return a data destination path."""
+    path = file_paths_pair[1]
+    path.write_bytes(other_payload)
```
webknjaz: Why can't we just use `transmit_payload` here?
```suggestion
def dst_exists_path(file_paths_pair, transmit_payload):
    """Return a data destination path."""
    path = file_paths_pair[1]
    path.write_bytes(transmit_payload)
```

Also, I'd rename this fixture to `pre_existing_dst_path` to communicate the intent better.

Jakuje: Because we want to be sure that the existing file has different content than the transmitted one to make sure it will be correctly overridden with the new data. Renaming is a good idea though.

tests/unit/sftp_test.py
Jakuje: The lint does not like the f-strings:
```
tests/unit/sftp_test.py

  67:12    WPS305 Found `f` string
  return f'Original content: {uuid_name!s}'.encode()
         ^

WPS305: Found `f` string
  1     tests/unit/sftp_test.py
```

webknjaz: Oh, we just need to disable that rule. But fine, this is out of the scope right now.

webknjaz: Marking this as resolved.

tests/unit/sftp_test.py
```diff
@@ -48,6 +60,22 @@ def dst_path(file_paths_pair):
...
+    return 'Original content: {name!s}'.format(name=uuid_name).encode()
+
+
+@pytest.fixture
+def pre_existing_dst_path(file_paths_pair, other_payload):
+    """Return a data destination path."""
+    path = file_paths_pair[1]
+    path.write_bytes(other_payload)
+    assert path.exists()
+    return path
```
webknjaz: Small refactor to deduplicate logic
```suggestion
def pre_existing_dst_path(dst_path, other_payload):
    """Return a data destination path."""
    dst_path.write_bytes(other_payload)
    assert dst_path.exists()
    return dst_path
```

tests/unit/sftp_test.py
```diff
@@ -18,11 +20,21 @@ def sftp_session(ssh_client_session):
...
+def transmit_payload(request: pytest.FixtureRequest):
+    """Generate binary test payloads of assorted sizes.
+
+    The choice 32 is arbitrary small value.
+
+    The choice 1024 + 1 is meant to be 1B larger than the chunk size used in
+    :file:`sftp.pyx` to make sure we excercise at least two rounds of reading/writing.
+    """
+    payload_len = request.param
+    random_bytes = [ord(random.choice(string.printable)) for _ in range(payload_len)]  # NOSONAR
```
webknjaz: It's unclear what this is ignoring. Let's not add suppressions w/o justification in a comment.

tests/unit/sftp_test.py
webknjaz: ack

tests/unit/sftp_test.py
Jakuje: Added a comment. I did not figure out any other way to skip this warning.

src/pylibsshext/sftp.pyx
webknjaz: In the future, it's best to exclude such things so it's easier to study the diff.

tests/unit/sftp_test.py
webknjaz: Let's revert it, then, and I'll make sonar shut up on the web UI.

tests/unit/sftp_test.py
```diff
@@ -18,11 +20,22 @@ def sftp_session(ssh_client_session):
...
-@pytest.fixture
-def transmit_payload():
-    """Generate a binary test payload."""
-    uuid_name = uuid.uuid4()
-    return 'Hello, {name!s}'.format(name=uuid_name).encode()
+@pytest.fixture(
+    params=(32, 1024 + 1),
+    ids=('small-payload', 'large-payload'),
+)
+def transmit_payload(request: pytest.FixtureRequest):
```
webknjaz: Since you're already adding typing to the args here, let's also annotate the return type.
```suggestion
def transmit_payload(request: pytest.FixtureRequest) -> bytes:
```

tests/unit/sftp_test.py
Jakuje: Reverted

webknjaz:
@Jakuje I haven't yet fixed the CI, but if this is ready, I guess I could just force-merge. Otherwise, this will have to hang for some while.

Jakuje:
> @Jakuje I haven't yet fixed the CI, but if this is ready, I guess I could just force-merge. Otherwise, this will have to hang for some while.

I do not have any huge urgency to get this merged personally. But this change is quite trivial and people in Ansible are hitting this in the wild so if the CI can be bypassed and you are ok with this, I think users would welcome to have this fixed.

webknjaz:
@Jakuje I'll bypass the CI, but I don't think we'll be releasing it until it's fixed.

PR #641, opened 2024-09-11
Implement asynchronous SFTP file transfer

Jakuje:
##### SUMMARY
The SFTP transfers done the way they work now are very slow especially on connections with large latency because each chunk (now 1024B) needs to travel all the way to remote as well as the confirmation (or request + data for the other transfer), before another chunk is sent. This can be improved by two things: 
 * sending larger chunks (up to the size reported by SFTP limits extension)
 * sending more requests while waiting for the confirmation/data

This depends on #636 (libssh 0.11), but also on other platforms having the updated libssh. It will likely need some conditional compilation based on the libssh version, but I did not find a good solution for cython after it was deprecated (cython/cython#4310) so any help/suggestions welcomed.

This PR also reproduces an issue with the new API that it currently hangs waiting for the second message, while it already happily sits in the queue, which is being worked on upstream.

Therefore marking as a draft to collect initial inputs.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/src/pylibsshext/includes/sftp.pxd
+++ b/src/pylibsshext/includes/sftp.pxd
@@ -22,7 +22,10 @@ from libc cimport stdint
 from pylibsshext.includes.libssh cimport ssh_channel, ssh_session, ssh_string
 
 
-cdef extern from "libssh/sftp.h" nogil:
+cdef extern from "includes/sftp_compat.h" nogil:
+
+    enum:
+        HAVE_SFTP_AIO
 
     struct sftp_session_struct:
         pass
@@ -84,6 +87,25 @@ cdef extern from "libssh/sftp.h" nogil:
 
     sftp_attributes sftp_stat(sftp_session session, const char *path)
 
+    void sftp_attributes_free(sftp_attributes file)
+
+    struct sftp_aio_struct:
+        pass
+    ctypedef sftp_aio_struct * sftp_aio
+    ssize_t sftp_aio_begin_read(sftp_file file, size_t len, sftp_aio *aio)
+    ssize_t sftp_aio_wait_read(sftp_aio *aio, void *buf, size_t buf_size)
+    ssize_t sftp_aio_begin_write(sftp_file file, const void *buf, size_t len, sftp_aio *aio)
+    ssize_t sftp_aio_wait_write(sftp_aio *aio)
+    void sftp_aio_free(sftp_aio aio)
+
+    struct sftp_limits_struct:
+        stdint.uint64_t max_packet_length
+        stdint.uint64_t max_read_length
+        stdint.uint64_t max_write_length
+        stdint.uint64_t max_open_handles
+    ctypedef sftp_limits_struct * sftp_limits_t
+    sftp_limits_t sftp_limits(sftp_session sftp)
+    void sftp_limits_free(sftp_limits_t limits)
 
 cdef extern from "sys/stat.h" nogil:
     cdef int S_IRWXU
--- a/src/pylibsshext/sftp.pxd
+++ b/src/pylibsshext/sftp.pxd
@@ -16,6 +16,8 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 #
+from libc cimport stdint
+
 from pylibsshext.includes cimport libssh, sftp
 from pylibsshext.session cimport Session
 
@@ -23,3 +25,19 @@ from pylibsshext.session cimport Session
 cdef class SFTP:
     cdef Session session
     cdef sftp.sftp_session _libssh_sftp_session
+
+cdef class _RemoteFile:
+    cdef sftp.sftp_file _fd
+
+cdef class SFTP_AIO:
+    cdef SFTP _sftp_obj
+    cdef _aio_queue
+    cdef _remote_file
+    cdef _local_file
+    cdef stdint.uint64_t _file_size
+    cdef stdint.uint64_t _total_bytes_requested
+    cdef sftp.sftp_session _sftp
+    cdef sftp.sftp_limits_t _limits
+
+cdef class C_AIO:
+    cdef sftp.sftp_aio aio
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -15,6 +15,10 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 
+import os
+from collections import deque
+from typing import BinaryIO
+
 from posix.fcntl cimport O_CREAT, O_RDONLY, O_TRUNC, O_WRONLY
 
 from cpython.mem cimport PyMem_Free, PyMem_Malloc
@@ -27,6 +31,13 @@ from pylibsshext.session cimport get_libssh_session
 # The value 32kB is a safe fallback when we cannot determine better value from
 # the server, for example using limits@openssh.com (since libssh 0.11.0).
 SFTP_MAX_CHUNK = 32_768
+
+# 255kB is absolute maximum we want to send over the SSH channel. Sending
+# more will likely not work and capping at some value will prevent DoS.
+SFTP_MAX_CHUNK_LIMIT = 261_120
+
+# Default number of requests for asynchronous/overlapping uploads and downloads.
+SFTP_MAX_REQUESTS = 10
 
 
 MSG_MAP = {
@@ -45,6 +56,40 @@ MSG_MAP = {
     sftp.SSH_FX_WRITE_PROTECT: "We are trying to write on a write-protected filesystem",
     sftp.SSH_FX_NO_MEDIA: "No media in remote drive"
 }
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
+    def __cinit__(self, sftp_obj: SFTP, path: str | bytes, int flags):
+        mode = "write" if flags & O_WRONLY else "read"
+        path_b = path
+        if isinstance(path_b, str):
+            path_b = path.encode("utf-8")
+
+        self._fd = sftp.sftp_open(
+            sftp_obj._libssh_sftp_session,
+            path_b,
+            flags,
+            sftp.S_IRWXU)
+        if self._fd is NULL:
+            raise LibsshSFTPException(
+                "Opening remote file [%s] for %s failed with error [%s]"
+                % (str(path), mode, sftp_obj._get_sftp_error_str()))
+
+    def __dealloc__(self):
+        if self._fd is not NULL:
+            sftp.sftp_close(self._fd)
+            self._fd = NULL
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if self._fd is not NULL:
+            sftp.sftp_close(self._fd)
+            self._fd = NULL
+        return False
+
+
 cdef class SFTP:
     def __cinit__(self, session):
         self.session = session
@@ -59,16 +104,18 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
-    def put(self, local_file, remote_file):
+    def put(self, local_file: str | os.PathLike, remote_file: str | bytes):
         """
         Upload local file to remote server.
 
         :param local_file: The file name on the local file system to upload
-        :type local_file: str or os.PathLike
 
         :param remote_file: The path to upload the file on the remote system
-        :type remote_file: str or bytes
         """
+        if sftp.HAVE_SFTP_AIO:
+            return SFTP_AIO(self).put(local_file, remote_file)
+
+        # Synchronous code compatible with libssh < 0.11.0
         cdef sftp.sftp_file rf
         cdef const char* c_buf
         with open(local_file, "rb") as f:
@@ -108,7 +155,7 @@ cdef class SFTP:
                 read_buffer = f.read(SFTP_MAX_CHUNK)
             sftp.sftp_close(rf)
 
-    def get(self, remote_file, local_file):
+    def get(self, remote_file: str | bytes, local_file: str | os.PathLike):
         """
         Download remote file to local path.
 
@@ -118,12 +165,15 @@ cdef class SFTP:
         :param local_file: The path on the local file system to place the downloaded file
         :type local_file: str or os.PathLike
         """
-        cdef sftp.sftp_file rf
+        if sftp.HAVE_SFTP_AIO:
+            return SFTP_AIO(self).get(remote_file, local_file)
+
+        # Synchronous code compatible with libssh < 0.11.0
         cdef char *read_buffer = NULL
         cdef sftp.sftp_attributes attrs
 
         remote_file_b = remote_file
-        if isinstance(remote_file_b, unicode):
+        if isinstance(remote_file_b, str):
             remote_file_b = remote_file.encode("utf-8")
 
         attrs = sftp.sftp_stat(self._libssh_sftp_session, remote_file_b)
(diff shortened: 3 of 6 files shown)


webknjaz:
I applied some CI fixes and now this PR needs to be rebased to pick them up. I'm clicking the rebase button now.

src/pylibsshext/sftp.pyx
```diff
@@ -54,69 +57,186 @@ cdef class SFTP:
...
+        if rf is NULL:
+            raise LibsshSFTPException("Opening remote file [%s] for write failed with error [%s]" % (remote_file, self._get_sftp_error_str()))
+        self._rf = rf
+
+        with open(local_file, "rb") as f:
+            f.seek(0, os.SEEK_END)
+            self._file_size = f.tell()
+            f.seek(0, os.SEEK_SET)
+
+            # open up to 10 parallel transfers
```
galorithm: I am not sure whether the word "parallel" is the most appropriate choice here.

In libssh code, we avoided using the word parallel / concurrent to make it clear to the reader that there is no parallelism or concurrency (e.g thread based) in the picture. I think the same should be applicable here.

Would "outstanding requests" be a better choice ?

src/pylibsshext/sftp.pyx
```diff
@@ -54,69 +57,186 @@ cdef class SFTP:
...
+            i = 0
+            while i < 10 and self._total_bytes_requested < self._file_size:
+                self._put_chunk(f)
+                i += 1
+
+            while len(self._aio_queue):
+                aio = self._aio_queue.pop()
```
galorithm: From my understanding, pop() will pop the last appended request out of the list (based on the following example performed using python interpreter).
 
```
>>> q = []
>>> q.append(10)
>>> q.append(20)
>>> q.append(30)
>>> q
[10, 20, 30]
>>> q.pop()
30
>>> 

```

So if we issue requests Rq_1, Rq_2 ... Rq_10 and append each of them one by one by the queue. The pop() will pop out Rq_10 and will wait for it.

Is this intended ?

I think we should instead wait for Rq_1's response first because:
- That is expected to arrive sooner than Rq_10's response.
- After we get the response, we may issue more requests. So, waiting for a response that comes earlier would help the transfer to proceed faster.

 Hence I think, the code should be changed to:
```suggestion
                aio = self._aio_queue.pop(0)
```

(Unless I am missing something here, I think this is why the bug with libssh was being encountered because the code waited for later request's response first before waiting for responses of earlier issued outstanding requests.)

galorithm: Additionally, reading the python documentation (https://docs.python.org/3/tutorial/datastructures.html#using-lists-as-queues) which suggests to not use lists (for our purpose of queue) as it isn't very efficient in case of pops from the beginning of the list.

So we could also go ahead and use `collections.deque` as they suggest, if we care about efficiency here and if its available.

(Hopefully I am not missing something that  justifies the use of list and pop() as its currently done)

src/pylibsshext/sftp.pyx
```diff
@@ -54,69 +57,186 @@ cdef class SFTP:
...
+
+            with open(local_file, 'wb') as f:
+                # open up to 10 parallel transfers
+                i = 0
+                while i < 10 and self._total_bytes_requested < self._file_size:
+                    self._get_chunk()
+                    i += 1
+
+                while len(self._aio_queue):
+                    aio = self._aio_queue.pop()
```
galorithm: Similar to the previous comment for pop() in case of write.

From what I understand, in case of pop, assuming requests Rq_1, Rq_2 ... Rq_10 are issued:

In case of download, waiting for Rq_10's response first (due to pop) would write the data of response 10 first in the local file before writing the data that comes as request 1's response, which isn't the desired order for file contents.

Hence, I think the usage of pop here should also be changed in a similar way as suggested for write.

src/pylibsshext/sftp.pyx
```diff
@@ -54,69 +57,186 @@ cdef class SFTP:
...
-        if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
-            return MSG_MAP[error]
-        return "Generic failure: %s" % self.session._get_session_error_str()
+            rf = sftp.sftp_open(self._sftp, remote_file_b, O_RDONLY, sftp.S_IRWXU)
+            if rf is NULL:
+                raise LibsshSFTPException("Opening remote file [%s] for reading failed with error [%s]" % (remote_file, self._get_sftp_error_str()))
+            self._rf = rf
+
+            with open(local_file, 'wb') as f:
+                # open up to 10 parallel transfers
```
galorithm: Same comment regarding changing the word "parallel" as commented for upload.

galorithm (changes requested):

src/pylibsshext/sftp.pyx
webknjaz: Yes, this is a good suggestion: https://wiki.python.org/moin/TimeComplexity

src/pylibsshext/sftp.pyx
webknjaz: It would be interesting to look into threading in the context of Python 3.13t. But that's obviously out of the scope here. I do agree that this word should be replaced, though.

src/pylibsshext/sftp.pyx
Jakuje: Thanks for double-checking! This is obviously overlook. I obviously wanted to use the list as a queue and messed it up.

Given that we really just store there up to 10 references to AIO structures, the performance should not really matter, but I changed it to deque anyway.

src/pylibsshext/sftp.pyx
Jakuje: Thanks. I reworded this to avoid `parallel` and described it as `start up to 10 requests before waiting for responses`.

src/pylibsshext/sftp.pyx
Jakuje: Thanks! Fixed now!

src/pylibsshext/sftp.pyx
Jakuje: Right. I do not expect treading would help the sftp performance thoughput as it needs to go through a single tcp connection and ssh session (and sharing ssh session among different threads is not supported by libssh).

Jakuje:
The testing farm build are failing because the libssh 0.11 did not make it to the stable fedora versions.

NilashishC:
Looks like v0.11.0 should be available now in F41? https://koji.fedoraproject.org/koji/buildinfo?buildID=2524005

Jakuje:
> Looks like v0.11.0 should be available now in F41? https://koji.fedoraproject.org/koji/buildinfo?buildID=2524005

Correct, but not in RHEL9 and other target architectures, which means this one still need some work to keep this code working also with the older libssh. Either with some conditional compilation or some other tricks...

NilashishC:
@Jakuje So this is what I understood. The async SFTP is only available in `libssh >=0.11.0`, and this PR updates the existing SFTP implementation in pylibssh to use the libssh async SFTP API. This means that if this PR is merged, the minimum required libssh version for pylibssh SFTP to work will be >=0.11.0. 

Now, the challenge is that `libssh >=0.11.0` is not _yet_ available on all platforms (and target arch). Furthermore, it isn't clear to me if `libssh >=0.11.0` will be made available on other RHEL versions (+target arch) that AAP currently supports OR if the async SFTP support will be backported to older libssh version(s), which would also require a bump in the min libssh version required by this binding.

An alternative that I can think of is inspect the libssh version and invoke the async implementation is we're running a libssh version that supports the async SFTP API and fallback to the existing implementation if not. That way, older pylibssh SFTP would still work but it'll be terribly slow like it has been so far.

Please let me know if what I said above is correct here.

Jakuje:
All correct. We do not plan to backport this API to older RHELs so if there is a way to implement some fallback, that would be preferred. Hints or pointers welcomed if you know how to approach. I am keeping this as a draft as this is started more like a proof of concept that the new API does what it is supposed to do.

To increase the SFTP throughput everywhere, the #664 should get you basically to the same performance for the most common read sizes (up until 32kB. But as you can see in #638, the large (>1kB) reads over SFTP were not working at all so right now, I would rather focus on these PRs. I think the changes there are ready to be merged (or at least reviewed).

src/pylibsshext/sftp.pyx
```diff
@@ -58,85 +61,186 @@ cdef class SFTP:
...
+    return sftp_obj._libssh_sftp_session
+
+cdef class SFTP_AIO:
+    def __cinit__(self, SFTP sftp_obj):
+        self._sftp = get_sftp_session(sftp_obj)
+
+        self._limits = sftp.sftp_limits(self._sftp)
+        if self._limits is NULL:
+            raise LibsshSFTPException("Failed to get remote SFTP limits [%s]" % (self._get_sftp_error_str()))
+
```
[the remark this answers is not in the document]
Jakuje: oh, this should be called on `sftp_obj` instead of `self`. Good catch!

src/pylibsshext/sftp.pyx
```diff
@@ -58,85 +61,186 @@ cdef class SFTP:
...
+                    self._get_chunk()
+
         finally:
-            if read_buffer is not NULL:
-                PyMem_Free(read_buffer)
-        sftp.sftp_close(rf)
+            if buffer is not NULL:
+                PyMem_Free(buffer)
+            sftp.sftp_close(rf)
+            self._rf = NULL
```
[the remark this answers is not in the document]
Jakuje: `sftp_close()` handles NULL argument

src/pylibsshext/sftp.pyx
```diff
@@ -58,85 +61,189 @@ cdef class SFTP:
...
-            return MSG_MAP[error]
-        return "Generic failure: %s" % self.session._get_session_error_str()
+
+cdef class C_AIO:
+    def __cinit__(self):
+        self.aio = NULL
+
+    def __dealloc__(self):
+        sftp.sftp_aio_free(self.aio)
+        self.aio = NULL
```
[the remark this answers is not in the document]
Jakuje: This is safe. Libssh takes NULL arguments to free function.

src/pylibsshext/sftp.pyx
```diff
@@ -58,6 +66,10 @@ cdef class SFTP:
             self._libssh_sftp_session = NULL
 
     def put(self, local_file, remote_file):
+        if sftp.HAVE_SFTP_AIO:
+            return SFTP_AIO(self).put(local_file, remote_file)
```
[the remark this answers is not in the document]
Jakuje: The sftp_limits() returns sane fallback limits if the server does not support the extension. It can fail only on memory issues, which result in hard failures.

src/pylibsshext/sftp.pyx
```diff
@@ -80,10 +92,14 @@ cdef class SFTP:
                             self._get_sftp_error_str(),
                         )
                     )
-                buffer = f.read(SFTP_MAX_CHUNK)
+                buffer = f.read(SFTP_MAX_CHUNK_BASE)
             sftp.sftp_close(rf)
 
     def get(self, remote_file, local_file):
+        if sftp.HAVE_SFTP_AIO:
+            return SFTP_AIO(self).get(remote_file, local_file)
```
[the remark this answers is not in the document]
Jakuje: The sftp_limits() returns sane fallback limits if the server does not support the extension. It can fail only on memory issues, which result in hard failures.

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +157,206 @@ cdef class SFTP:
...
+
+        cdef sftp.sftp_aio aio = NULL
+        bytes_requested = sftp.sftp_aio_begin_write(self._rf, PyBytes_AS_STRING(buffer), to_write, &aio)
+        if bytes_requested != to_write:
+            raise LibsshSFTPException("Failed to write chunk of size [%d] of file [%s] with error [%s]"
+                                      % (to_write, self._remote_file, self._sftp_obj._get_sftp_error_str()))
+        self._total_bytes_requested += bytes_requested
+        c_aio = C_AIO()
+        c_aio.aio = aio
+        self._aio_queue.append(c_aio)
```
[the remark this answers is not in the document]
Jakuje: Libssh encrypts data and sends the packet before returning. The wait is just waiting for the server reply.

src/pylibsshext/includes/sftp.pxd
```diff
@@ -22,7 +22,9 @@ from libc cimport stdint
 from pylibsshext.includes.libssh cimport ssh_channel, ssh_session, ssh_string
 
 
-cdef extern from "libssh/sftp.h" nogil:
+cdef extern from "includes/sftp_compat.h" nogil:
+
+    bint HAVE_SFTP_AIO
```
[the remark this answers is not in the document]
Jakuje: Good catch! Gemini acknowledged it was wrong in this approach!

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +157,206 @@ cdef class SFTP:
...
+            self._rf = NULL
+
+    def _get_chunk(self):
+        to_read = self._cap_read_size(self._file_size - self._total_bytes_requested)
+        cdef sftp.sftp_aio aio = NULL
+        bytes_requested = sftp.sftp_aio_begin_read(self._rf, to_read, &aio)
+        if bytes_requested != to_read:
+            raise LibsshSFTPException("Failed to request to read chunk of size [%d] of file [%s] with error [%s]"
+                                      % (to_read, self._remote_file, self._sftp_obj._get_sftp_error_str()))
+        self._total_bytes_requested += bytes_requested
```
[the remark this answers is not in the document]
Jakuje: Can you reference where the SFTP specification says it can return less bytes than requested?

If we receive less bytes than requested, we fail somewhere down the road.

src/pylibsshext/sftp.pxd
```diff
@@ -23,3 +23,17 @@ from pylibsshext.session cimport Session
...
+cdef class SFTP_AIO:
+    cdef SFTP _sftp_obj
+    cdef _aio_queue
+    cdef _remote_file
+    cdef _local_file
+    cdef _file_size
+    cdef _total_bytes_requested
+    cdef sftp.sftp_session _sftp
+    cdef sftp.sftp_limits_t _limits
+    cdef sftp.sftp_file _rf
```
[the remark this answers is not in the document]
Jakuje: Good point. Added types

Jakuje:
Rebased on top of #848 to unbreak the Packit CI.

Jakuje:
> In the async download path, completion is determined by _total_bytes_requested (bytes queued/requested), but the code never tracks bytes actually received/written. If sftp_aio_wait_read() returns fewer bytes than requested (including 0 at EOF), the loop can exit successfully with a truncated local file because _total_bytes_requested may already equal _file_size and no further reads are issued.

The `aio` structure internally contains the requested length and the function `sftp_aio_wait_read()` already checks the server returned us the requested amount of data so no need to check here:

https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/sftp_aio.c?ref_type=heads#L198

> Docstring typo: "its capped" should be "it's capped" (contraction for "it is").

I will fix this.

Jakuje:
The copilot is annoyingly finding issues in the old code so I extracted them to separate commit. I can move that one into separate PR if you prefer. But the happy path should not change function-wise from the previous versions.

src/pylibsshext/sftp.pyx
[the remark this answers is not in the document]
webknjaz: It could be a good idea to record this in some sort of a memory file or a `THREAT_MODEL.md` — a commonly used document that communicates security boundaries and other assumptions of a project. We don't have one yet but should start having one at some point.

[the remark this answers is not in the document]
Jakuje: Not sure if this is something for a threat modeling, rather than API assumptions. I find this logic very common in many C projects, because it makes cleanup handling significantly easier than checking for null before each of the call (and making a good habit from initializing variables at declaration).

Jakuje:
Moved all the fixes for synchronous SFTP to #862 and #869 - #875. This now contains only the self-contained async SFTP implementation and could theoretically go in even without the other 8 PRs if this would be the goal.

src/pylibsshext/sftp.pyx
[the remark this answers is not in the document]
webknjaz: This is not really a C project, though. So people around tend to have different habits. And these differences in perception are usually where security problems happen 🤷‍♂️

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+            remote_file_b = remote_file.encode("utf-8")
+
+        rf = sftp.sftp_open(self._sftp, remote_file_b, O_WRONLY | O_CREAT | O_TRUNC, sftp.S_IRWXU)
+        if rf is NULL:
+            raise LibsshSFTPException("Opening remote file [%s] for write failed with error [%s]"
+                                      % (remote_file, self._sftp_obj._get_sftp_error_str()))
+        self._rf = rf
+
+        try:
+            with open(local_file, "rb") as f:
```
webknjaz: Can we plz avoid using 1-2 letter var names and use something more representative?

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+                f.seek(0, os.SEEK_END)
+                self._file_size = f.tell()
+                f.seek(0, os.SEEK_SET)
```
webknjaz: Looks like this could be abstracted into a helper function.

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+        self._aio_queue = deque()
+        self._total_bytes_requested = 0
+
+        cdef C_AIO aio
+        cdef sftp.sftp_file rf
+        self._remote_file = remote_file
+        self._local_file = local_file
+
+        remote_file_b = remote_file
+        if isinstance(remote_file_b, unicode):
```
```suggestion
        if isinstance(remote_file_b, str):
```

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+
+    def __dealloc__(self):
+        if self._limits is not NULL:
+            sftp.sftp_limits_free(self._limits)
+            self._limits = NULL
+        if self._rf is not NULL:
+            sftp.sftp_close(self._rf)
+            self._rf = NULL
+
+    def put(self, local_file, remote_file):
```
webknjaz: type annotations plz

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+                    # whole file written
+                    if self._total_bytes_requested == self._file_size:
+                        continue
+
+                    # else issue more write requests
+                    self._put_chunk(f)
+
+        finally:
+            sftp.sftp_close(rf)
+            self._rf = NULL
```
webknjaz: This entire thing should be abstracted via a `contextlib.contextmanager`.

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+        attrs = sftp.sftp_stat(self._sftp, remote_file_b)
+        if attrs is NULL:
+            raise LibsshSFTPException("Failed to stat the remote file [%s] with error [%s]"
+                                      % (remote_file, self._sftp_obj._get_sftp_error_str()))
+        self._file_size = attrs.size
+        sftp.sftp_attributes_free(attrs)
+
+        buffer_size = self._cap_read_size(self._file_size)
+        try:
+            buffer = <char *>PyMem_Malloc(buffer_size)
```
webknjaz: `buffer` is a name of the builtin function in Python.

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +179,227 @@ cdef class SFTP:
...
+                                                  % (bytes_read, self._remote_file, bytes_written, local_file))
+
+                    # whole file read
+                    if self._total_bytes_requested == self._file_size:
+                        continue
+
+                    # else issue more read requests
+                    self._get_chunk()
+
+        finally:
```
webknjaz, line 472: This tells me the resource management block around here has a case for being abstracted w/ `contextlib.contextmanager`.

Jakuje: Good idea. I did this separately in #879 as it involves also the synchronous SFTP.

Jakuje: Did also here among other changes

src/pylibsshext/sftp.pyx
Jakuje: fixed. Forgot about that in this PR

src/pylibsshext/sftp.pyx
Jakuje: Done in #879 for synchronous SFTP and also here.

src/pylibsshext/sftp.pyx
Jakuje: added

src/pylibsshext/sftp.pyx
Jakuje: changed. Thanks

src/pylibsshext/sftp.pyx
Jakuje: moved to helper function.

src/pylibsshext/sftp.pyx
Jakuje: Used longer variable name

PR #675, opened 2025-02-06
Add support for specifying Kex Algos

NilashishC:
##### SUMMARY
- Allow setting key exchange methods to be used.

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION
None

PR #676, opened 2025-02-10
🧪 Bump `actions/{upload,download}-artifact` to v4

NilashishC:
##### SUMMARY

They are updated across all GitHub Actions CI/CD workflow definitions
in the repository. It is necessary due to Artifacts API v3 having been
deprecated on Jan 30, 2025.

As a part of the change, the uploaded artifact names needed to be made
unique and their download counterparts had to be invoked against a
name pattern, merging multiple artifacts together.

Making the names unique came with challenges that this patch overcame
by appending SHA-256 hashes of reusable workflow inputs to the
artifact names.

This change also bumps `re-actors/checkout-python-sdist` to v2 as it
uses `actions/download-artifact` internally and so we
needed to get the version that uses non-deprecated APIs.

[1]: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

##### ISSUE TYPE

- Maintenance Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1124,7 +1124,7 @@ jobs:
       run: mkdir -pv ~/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}
 
     - name: Retrieve the project source from an sdist inside the GHA artifact
-      uses: re-actors/checkout-python-sdist@release/v1
+      uses: re-actors/checkout-python-sdist@release/v2
       with:
         source-tarball-name: ${{ needs.pre-setup.outputs.sdist-artifact-name }}
         workflow-artifact-name: >-
@@ -1404,7 +1404,7 @@ jobs:
         python-version: 3.11
 
     - name: Retrieve the project source from an sdist inside the GHA artifact
-      uses: re-actors/checkout-python-sdist@release/v1
+      uses: re-actors/checkout-python-sdist@release/v2
       with:
         source-tarball-name: ${{ needs.pre-setup.outputs.sdist-artifact-name }}
         workflow-artifact-name: >-
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -597,7 +597,7 @@ jobs:
     - name: Verify that expected patch got created
       run: ls -1 '${{ needs.pre-setup.outputs.changelog-patch-name }}'
     - name: Save the package bump patch as a GHA artifact
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: changelog
         path: |
@@ -876,7 +876,7 @@ jobs:
         ref: ${{ github.event.inputs.release-committish }}
 
     - name: Fetch the GHA artifact with the version patch
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
         name: changelog
     - name: Apply the changelog patch but do not commit it
@@ -1017,7 +1017,7 @@ jobs:
         'dist/${{ needs.pre-setup.outputs.sdist-artifact-name }}'
     - name: Store the source distribution package
       if: fromJSON(matrix.store-sdist-to-artifact)
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: ${{ needs.pre-setup.outputs.dists-artifact-name }}
         # NOTE: Exact expected file names are specified here
@@ -1137,10 +1137,11 @@ jobs:
         packaging/rpm/ansible-pylibssh.spec
 
     - name: Download all the dists  # because `rpmlint` executes the spec file
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
-        name: ${{ needs.pre-setup.outputs.dists-artifact-name }}
+        pattern: ${{ needs.pre-setup.outputs.dists-artifact-name }}*
         path: dist/
+        merge-multiple: true
 
     - name: Lint the RPM spec file
       if: >-
@@ -1303,7 +1304,7 @@ jobs:
         );
         echo "artifact-id=${normalized_container_id}" >> "${GITHUB_OUTPUT}"
     - name: Store RPM and SRPM as artifacts
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: ${{ steps.artifact-name.outputs.artifact-id }}--srpm-n-rpm
         path: |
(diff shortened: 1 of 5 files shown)


NilashishC:
@webknjaz Hi. Could you please have a look at this PR? This needs to be merged before https://github.com/ansible/pylibssh/pull/675.

Thank you.

webknjaz (commented):
@NilashishC this is a move in the right direction. It is, however, incomplete.

You have to upgrade `re-actors/checkout-python-sdist@release/v1` to `re-actors/checkout-python-sdist@release/v2` (because v1 uses an older version of the download action).

And additionally, the wheel uploads must change their artifact names to be unique (example: https://github.com/aio-libs/yarl/blob/07c1b4f/.github/workflows/reusable-build-wheel.yml#L92C15-L95C28).

Finally, the download step in the publishing jobs must change to merge+pattern (example: https://github.com/aio-libs/yarl/blob/07c1b4f/.github/workflows/ci-cd.yml#L532C9-L533C29). The last one will not be testable in the PR CI runs and will only fail later if you don't get it right.

NilashishC:
@webknjaz thanks for the review.

The variables required to make the wheel name unique in `reusable-build-wheel.yml` seems to be commented out in `inputs` - https://github.com/ansible/pylibssh/blob/devel/.github/workflows/reusable-build-wheel.yml#L17-L26. Should I uncomment those and add it to the wheel name? Or should I choose other inputs.

NilashishC:
> @webknjaz thanks for the review.
> 
> The variables required to make the wheel name unique in `reusable-build-wheel.yml` seems to be commented out in `inputs` - https://github.com/ansible/pylibssh/blob/devel/.github/workflows/reusable-build-wheel.yml#L17-L26. Should I uncomment those and add it to the wheel name? Or should I choose other inputs.

Okay, I did try something to make it unique - https://github.com/ansible/pylibssh/pull/676/files#diff-3826dffee70012c1e07ee3d63a4cf5ba7ed82c17bbac68145432a97e71dbade7R187-R192.

However, now some of the build jobs in the matrix fail.

webknjaz:
@NilashishC your commit message says

> use merge+pattern for download artifact

But that commit only makes use of the merge input and doesn't replace name with pattern. So these jobs only download the sdist while they expect the wheels to be present.

There might be other problems (I saw some weirdness in the logs) but I would imagine that those are unrelated. Fix this first and we'll see how it goes then.

NilashishC:
@webknjaz Apologies for my lack of understanding here. But I'm unsure what the pattern would be . The patterns differ in `reusable-cibuildwheel`  vs `reusable-build-wheel` (because the `inputs` are different). How do you recommend we handle this in the download steps?

webknjaz:
Wouldn't just adding an asterisk at the end do it? This should work for as long as the prefix remains the same.

NilashishC:
So, all x86 ones are passing, but not the other architectures.

.github/workflows/ci-cd.yml
```diff
@@ -275,7 +275,7 @@ jobs:
     - name: Set up pip cache
       if: >-
         steps.request-check.outputs.release-requested != 'true'
-      uses: actions/cache@v4.0.2
```
webknjaz: Let's move this to a separate PR.

.github/workflows/reusable-cibuildwheel.yml
```diff
@@ -81,9 +81,11 @@ jobs:
           with-cython-tracing=${{ inputs.cython-tracing }}
 
     - name: Upload built artifacts for testing and publishing
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
-        name: ${{ inputs.dists-artifact-name }}
+        name: ${{ inputs.dists-artifact-name }}-
+          ${{ inputs.os }}-
+          ${{ inputs.qemu }}
```
webknjaz: It might be useful to take this input into account, as #562 revealed the need to run this workflow several times to split wheel building into smaller portions.
```suggestion
          ${{ inputs.qemu }}-
          ${{ inputs.wheel-tags-to-skip }}
```

In the future, though, I'd look into rewriting this to use something like a hash of the entire `${{ inputs }}` object. But that's out of the scope right now.

webknjaz:
@NilashishC will you be cleaning up the commits?

webknjaz:
> So, all x86 ones are passing, but not the other architectures.

As for https://github.com/ansible/pylibssh/actions/runs/13268100337/job/37040748168?pr=676#step:9:50, it's probably a separate issue that I'd look into separately from this PR.

.github/workflows/reusable-tests.yml
```diff
@@ -190,7 +190,7 @@ jobs:
         '${{ env.TOX_VERSION }}'
 
     - name: Download all the dists
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
         name: ${{ inputs.dists-artifact-name }}
         path: dist/
```
webknjaz: The tests jobs are failing because this is missing:
```suggestion
        merge-multiple: true
        path: dist/
        pattern: ${{ inputs.dists-artifact-name }}*
```

(I checked the entire PR and this should be the only place needing the change)

NilashishC:
@webknjaz I _think_ I have resolved all your comments now. Here's the PR for `actions/cache` - https://github.com/ansible/pylibssh/pull/677.

NilashishC:
> @NilashishC will you be cleaning up the commits?

Do you want me to squash them into a single one?

webknjaz:
> Do you want me to squash them into a single one?

It's up to you now. I like a series of atomic commits. Not a blind squash. Currently, the repo still allows the fake merge methods like “squash” or “rebase” but I'm eventually going to enable the merge queues feature which will take those away.
So if you want this to be accepted via the squash method, it's fine for now. Just make sure there's nothing unrelated to the title of the PR in the diff.

.github/workflows/reusable-build-wheel.yml
```diff
@@ -184,9 +184,12 @@ jobs:
         ls -1
         dist/${{ inputs.wheel-artifact-name }}
     - name: Store ${{ inputs.manylinux-python-target }} binary wheel
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
-        name: ${{ inputs.dists-artifact-name }}
+        name: ${{ inputs.dists-artifact-name }}-
+          ${{ inputs.manylinux-year-target }}-
+          ${{ inputs.manylinux-image-target-arch }}-
```
webknjaz, line 210: Looks like one more factor is missing here:
```suggestion
          ${{ inputs.manylinux-image-target-arch }}-
          ${{ inputs.manylinux-image-target-qemu-arch }}-
```

It's not breaking the CI in PR builds because a portion of jobs are being skipped. However, it'll likely explode on release.

webknjaz:
Urgh.. https://github.com/ansible/pylibssh/actions/runs/13306998592/job/37160303882?pr=676#step:6:15 fails due to asterisks. Looks like that inputs hashing idea will probably have to make its way into this PR's scope. I'll think about it and maybe make a direct commit into your branch.

webknjaz:
@NilashishC I merged this, but the CI still needs love to be able to function and make releases. I'm planning to keep looking into it. Stay tuned!

PR #693, opened 2025-03-21
tests: Fix signal name in the test comment

Jakuje:
##### SUMMARY
Fixes just a comment in the test that I accidentally noticed.

##### ISSUE TYPE
- Test Bugfix Pull Request


Jakuje:
@webknjaz do we need changelog fragments also for such simple user invisible changes?

PR #658, opened 2024-11-14
Prevent `SEGFAULT`s on consecutive `exec_command()` invocations

Jakuje:
##### SUMMARY
The function `exec_command()` keeps the callbacks as a local variable before assigning them to the created channel. The channel is not guaranteed to be completely freed when `ssh_channel_free()` is called because there might be some leftover messages or responses to process (close confirmation, exit code ...).

Calling the `exec_command()` as done previously in the test from the same function without anything in between (except assert) will likely map the second function call to the same memory on the call stack so it was working most of the time. But calling it from different functions or contexts will likely change the call stack and processing of outstanding callbacks is more likely to result in addressing wrong memory location.

Likely fixes #57, #645 and #657

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
I was not able to reproduce the issue locally so pushing to see if the CI will be able to crash.

This is also introducing memory leaks as the callback structure is never freed. We should probably store it somewhere in the python code before returning to make sure it is not garbage collected (or can the python GC track the callback pointer is still stored on the libssh side?).

PR #694, opened 2025-03-21
Fix minor issues of handling channels

Jakuje:
##### SUMMARY
This is part of #658 to see what parts is really making the rpm tests time out.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -178,8 +178,8 @@ def sshd_addr(free_port_num, ssh_authorized_keys_path, sshd_hostkey_path, sshd_p
         opt, 'PidFile={pid!s}'.format(pid=sshd_path / 'sshd.pid'),
 
         # NOTE: 'UsePAM no' is not supported on Fedora.
-        # Ref: https://bugzilla.redhat.com/show_bug.cgi?id=770756#c1
-        opt, 'UsePAM=yes',
+        # But it is ok for testing as it simplifies everything
+        opt, 'UsePAM=no',
         opt, 'PasswordAuthentication=no',
         opt, 'ChallengeResponseAuthentication=no',
         opt, 'GSSAPIAuthentication=no',


Jakuje:
This should be the leftover change from #658 that makes the rpm builds fail. No idea why only in the rpmbuild case though. Need to figure out so keeping open as draft for now

webknjaz:
@Jakuje evidently, there were problems with Fedora years ago: https://github.com/ansible/pylibssh/commit/cdd8555312adfaa3509537657a587a3270791917. Can it be that RHEL caught up with those over time? Also, note that it was set together with a number of other settings.

Jakuje:
> @Jakuje evidently, there were problems with Fedora years ago: [cdd8555](https://github.com/ansible/pylibssh/commit/cdd8555312adfaa3509537657a587a3270791917). Can it be that RHEL caught up with those over time? Also, note that it was set together with a number of other settings.

This warning is about production systems where the PAM really takes care of setting up the user session through systemd and many other modules in play. But if you want to run tests, the PAM should not be needed for anything. If the tests run under non-root user, it does not even have access to set things like other users sessions.

In libssh, we run all the tests (that do not explicitly require PAM) with `UsePAM no`. But libssh has a bit more controlled environment with LD_PRELOAD wrappers (uid_wrapper, socket_wrapper ...) making sure the sshd believes it runs under root.

webknjaz:
@Jakuje is it possible to make the env more controlled here as well? I'd be in favor of making more things pinned/predictable.

Jakuje:
> @Jakuje is it possible to make the env more controlled here as well? I'd be in favor of making more things pinned/predictable.

Likely yes. The https://cwrap.org/ is really just setting some environment variables and LD_PRELOAD so if python would be able to handle the this, it should work the same way here. The ugly thing is that running the openssh sshd needs all of them.

webknjaz:
@Jakuje should this PR be renamed?

Jakuje:
> @Jakuje should this PR be renamed?

It should be probably closed as currently there is just one-line change that breaks half of the testing. I wanted to get back to that or to see if the wrappers could be used here, but it should be probably separate PR anyway.

PR #697, opened 2025-03-29
🐳 Pin base manylinux containers to 2025.01.19-2

webknjaz:
##### SUMMARY

SSIA.

##### ISSUE TYPE

- Maintenance Pull Request
- Packaging Pull Request

##### ADDITIONAL INFORMATION

The base manylinux container is currently unpinned, which means that the upstream can break us any time. This is an attempt to pinpoint when that happened.

### DIFF
--- a/.github/workflows/build-manylinux-container-images.yml
+++ b/.github/workflows/build-manylinux-container-images.yml
@@ -25,6 +25,7 @@ jobs:
     timeout-minutes: 120
 
     strategy:
+      fail-fast: false
       matrix:
         IMAGE:
         # Build containers for x86_64
--- a/.github/workflows/build-manylinux-container-images.yml
+++ b/.github/workflows/build-manylinux-container-images.yml
@@ -51,17 +51,6 @@ jobs:
         - VALUE: _2_24  # PEP 600
         - VALUE: _2_28
           TAG: 2025.01.19-2
-        include:
-        - IMAGE:
-            ARCH: x86_64
-            QEMU_ARCH: amd64
-          YEAR:
-            VALUE: 1
-        - IMAGE:
-            ARCH: x86_64
-            QEMU_ARCH: amd64
-          YEAR:
-            VALUE: 2010
 
     env:
       LIBSSH_VERSION: 0.9.6
--- a/.github/workflows/build-manylinux-container-images.yml
+++ b/.github/workflows/build-manylinux-container-images.yml
@@ -45,28 +45,32 @@ jobs:
         # Build containers for ppc64
         # - ARCH: ppc64
         YEAR:
-        - 2014
-        - _2_24  # PEP 600
-        - _2_28
+        - VALUE: 2014
+          TAG: 2025.01.19-2
+        - VALUE: _2_24  # PEP 600
+        - VALUE: _2_28
+          TAG: 2025.01.19-2
         include:
         - IMAGE:
             ARCH: x86_64
             QEMU_ARCH: amd64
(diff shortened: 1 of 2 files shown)


.github/workflows/build-manylinux-container-images.yml
```diff
@@ -20,7 +20,7 @@ on:  # yamllint disable-line rule:truthy
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
webknjaz: @Qalthos

This won't even start the jobs anymore because GH has fully sunset 20.04 a few weeks ago.

*(hence https://github.com/ansible/pylibssh/pull/708)*

But I can tell you that when I last updated the PR and was able to see the logs, they didn't look any different and were hitting segfaults in those QEMU jobs (cancelling the rest).

Qalthos: I am successfully building the containers (or, most of them at least) locally, and I'm trying to get this in a shape where GH is at least running them so I can see what the differences are

webknjaz: Okay, feel free to push to this PR. I'll avoid updating it meanwhile.

.github/workflows/build-manylinux-container-images.yml
```diff
@@ -45,28 +45,32 @@ jobs:
...
+        - VALUE: _2_24  # PEP 600
+        - VALUE: _2_28
+          TAG: 2025.01.19-2
         include:
         - IMAGE:
             ARCH: x86_64
             QEMU_ARCH: amd64
-          YEAR: 1
+          YEAR:
+            VALUE: 1
```
webknjaz: @Qalthos FYI, I dropped using manylinux1/2010/2014 in favor of PEP 600 since https://github.com/ansible/pylibssh/pull/562/files#diff-7eab675a08c9bdcc37864c516c651215c9f52401564e65c18e201b933c635890.

I assessed they aren't really needed anymore with the modern support matrix + the fact that downstreams would do their own thing anyway.

So you don't really need to worry about those. If we really want to build something with those, we still can use the old images that already exist on GHCR.

Hope this will reduce the number of things that need fixing/checking.

In fact, `cibuildwheel` only points to `2_28` so technically even `2_24` might not be necessary, but maybe we should keep building them. Modern enough pip (shipped in supported CPythons) will recognize these tags.

.github/workflows/build-manylinux-container-images.yml
```diff
@@ -45,28 +45,32 @@ jobs:
...
         - IMAGE:
             ARCH: x86_64
             QEMU_ARCH: amd64
-          YEAR: 2010
+          YEAR:
+            VALUE: 2010
 
     env:
       LIBSSH_VERSION: 0.9.6
```
webknjaz, line 56: Note to myself: we should eventually read this value from `pyproject.toml` or some other centralized location..

webknjaz:
@Qalthos I recommend adding `strategy.fail-fast: false` to the workflow for troubleshooting. This will prevent all matrix jobs from being cancelled if one fails.

Qalthos:
> @Qalthos I recommend adding `strategy.fail-fast: false` to the workflow for troubleshooting. This will prevent all matrix jobs from being cancelled if one fails.

Yeah I was looking for that but couldn't find it last night. Thanks for the reminder

webknjaz:
@Qalthos any getting to the bottom of this?

PR #705, opened 2025-04-23
Update libssh version to unblock container builds

Qalthos:
##### SUMMARY

Pre-0.11 releases have a too-low minimum cmake version. This is likely fixable, but let's see if this works first.

##### ISSUE TYPE

- Feature Pull Request


webknjaz (changes requested):
This is incomplete — another place to update would be `pyproject.toml` — we now route all the platform-specific builds through `cibuildwheel`.

Also, this would require a change note.

webknjaz:
@Qalthos now sure if you saw but I tried downgrading to older containers in https://github.com/ansible/pylibssh/pull/697 and it still didn't work for the non-x86 builds. And that's puzzling/why I wanted help with QEMU.

webknjaz:
@Qalthos here's a few more context comments in another PR trying to upgrade libssh: https://github.com/ansible/pylibssh/pull/636#issuecomment-2763708441

webknjaz:
So this is actually a duplicate of #636 that is more complete but is blocked by the QEMU breakage I wasn't able to figure out.

Qalthos:
Yeah I was just trying to build the containers locally and figure out where things were failing. I can close this for #636

PR #709, opened 2025-04-24
Fix Session.close() to prevent freeing underlying session

### LINKED ISSUE #702: SIGSEGV when running session.connect() for closed session
##### SUMMARY
Script fails with error  SIGSEGV "Segmentation fault" that cannot be handled

##### ISSUE TYPE
- Bug Report

```
ansible-pylibssh        1.2.2
libssh2-1:amd64         1.10.0-3+b1
```

##### OS / ENVIRONMENT
```
Debian GNU/Linux 12
Python 3.11.2
```

##### STEPS TO REPRODUCE
```console
# Step 1: create file with content
$ cat > segfault.py << "EOF"

session = Session()
session.close()
session.connect()
EOF

# Step 2: run script
$ python3 segfault.py
```

##### EXPECTED RESULTS
Script fails with exception
```python-traceback
pylibsshext.errors.LibsshSessionException
```

##### ACTUAL RESULTS
Script fails with "Segmentation fault"

Image

Jakuje:
##### SUMMARY
Call to connect() after close() crashed, likely because the NULL pointer was passed somewhere it should not.

The session should really be freed only in the destructor so removing the offending code.

Fixes: #702

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/709.bugfix.rst
+++ b/docs/changelog-fragments/709.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed crash when more operations were called after ``session.close()`` -- by :user:`Jakuje`.
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -539,8 +539,6 @@ cdef class Session(object):
         if self._libssh_session is not NULL:
             if libssh.ssh_is_connected(self._libssh_session):
                 libssh.ssh_disconnect(self._libssh_session)
-            libssh.ssh_free(self._libssh_session)
-            self._libssh_session = NULL
 
     def set_missing_host_key_policy(self, policy):
         """The policy to use if the know host key is missing.
--- a/tests/unit/session_test.py
+++ b/tests/unit/session_test.py
@@ -13,6 +13,15 @@ def test_make_session():
     assert Session()
 
 
+def test_make_session_close_connect():
+    """Make sure the session is usable after call to close()."""
+    session = Session()
+    session.close()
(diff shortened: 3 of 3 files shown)


docs/changelog-fragments/709.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed crash when more operations were called after session.close() -- by :user:`Jakuje`.
```
webknjaz:
```suggestion
Fixed crash when more operations were called after ``session.close()`` -- by :user:`Jakuje`.
```

webknjaz (approved):
@Jakuje does this make it so the underlying session object could persist and would then be reused on reconnect? I guess that's useful. Another way could be having some flag on the session for whether it's still usable.

P.S. I dropped Ubuntu 20.04 on `devel` so it would be a good idea to rebase now.

Jakuje:
> @Jakuje does this make it so the underlying session object could persist and would then be reused on reconnect? I guess that's useful. Another way could be having some flag on the session for whether it's still usable.

I think the libssh session should be reusable. But I know we had some bugs in the past that were making some operations fail so if we want to be sure, we should add some more extensive tests.

> P.S. I dropped Ubuntu 20.04 on devel so it would be a good idea to rebase now.

Will do.

docs/changelog-fragments/709.bugfix.rst
Jakuje: Thanks! Fixed.

PR #710, opened 2025-04-24
🧪 Disable `fail-fast` in CI when in debug mode

webknjaz:
##### SUMMARY

$sbj. This will let us get logs of all matrix jobs without in-matrix cancellations.

##### ISSUE TYPE

- Maintenance Pull Request

##### ADDITIONAL INFORMATION

N/A


webknjaz:
It doesn't work. Perhaps, because it's a PR from fork. Dunno. Will have to get back to this later.

PR #713, opened 2025-04-30
🧪 Use `tonistiigi/binfmt` QEMU container image

webknjaz:
This replaces `multiarch/qemu-user-static` that hasn't gotten a release in over two years. And fixes the breakage that occurred as a result of GitHub Actions CI/CD updating their Ubuntu VMs [[1]].

[1]: https://github.com/actions/runner-images/issues/11662

##### SUMMARY

$sbj.

##### ISSUE TYPE

- Maintenance Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/.github/workflows/build-manylinux-container-images.yml
+++ b/.github/workflows/build-manylinux-container-images.yml
@@ -86,11 +86,17 @@ jobs:
         Set up QEMU ${{ env.QEMU_ARCH }} arch emulation
         with Podman
       if: env.QEMU_ARCH != 'amd64'
-      run: >-
+      run: >
         sudo podman run
         --rm --privileged
-        multiarch/qemu-user-static
-        --reset -p yes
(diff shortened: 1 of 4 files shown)


Qalthos (approved):

webknjaz:
Merging postponed until Monday due to new `2_28` images not having Python 3.8 (and `2_24` doesn't have 3.12 and 3.13). Might end up making use of https://cibuildwheel.pypa.io/en/stable/options/#linux-image if we decide not to drop Python 3.8.

Qalthos:
I don't have any problem with dropping 3.8, that was essentially expected to be part of the "adding 3.13" work and I don't see a point in building a version we won't use

Qalthos:
cc @KB-perByte

Qalthos (approved):

PR #741, opened 2025-06-14
Start building pre-cached `manylinux_2_34` images

KB-perByte:


.gitignore
```diff
@@ -396,3 +396,4 @@ $RECYCLE.BIN/
 
 # setuptools-scm write_to:
 /src/pylibsshext/_scm_version.py
+.ansible
```
webknjaz: Don't add arbitrary entries into the project gitignore. You can do this in your user-global gitignore if you need it but it is in no way related to the project so it shouldn't be dumped in here.
```suggestion
```

docs/changelog-fragments/741.packaging.rst
```diff
@@ -0,0 +1,2 @@
+Updated building of manylinux versions to use _2_32
```
webknjaz:
```suggestion
Started caching ``manylinux`` build images to be used for producing ``2_32`` tagged wheels
```

.github/workflows/build-manylinux-container-images.yml
```diff
@@ -48,6 +48,7 @@ jobs:
         YEAR:
         - _2_24
         - _2_28
+        - _2_34
```
webknjaz, line 51: Match the change note.
```suggestion
        - _2_32
```

docs/changelog-fragments/741.packaging.rst
webknjaz, line 1: This should be a contrib note because it only affects our internal build infra.

webknjaz (changes requested):

webknjaz (approved):
@KB-perByte LGTM but have you checked if any older base image versions support Python 3.13?

webknjaz:
@KB-perByte I just verified that the base `manylinux_2_28` images actually contain Python 3.13, so there's no rush to migrate to `manylinux_2_34`, which would unnecessarily reduce compatibility. We should still keep building this cache in the CI, though, so the process remains healthy, and the future migration is seamless.

PR #744, opened 2025-06-16
Set `MACOSX_DEPLOYMENT_TARGET` in `cibuildwheel` when calling Homebrew

webknjaz:
##### SUMMARY

SSIA

##### ISSUE TYPE

- Packaging Pull Request

##### ADDITIONAL INFORMATION

N/A


webknjaz:
UPD: v2.21.3 is the first one with the stable Python 3.13 in. And v2.21.2 was when its support was declared official.

So we should work through the macOS blocker (build libssh and OpenSSL on our own) and upgrade further.

webknjaz:
Covered by #648.

PR #745, opened 2025-06-17
WIP: test update cibuildwheel

KB-perByte:
##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


webknjaz (changes requested):
This is way too premature. We don't really need to bump this far at this stage. macOS must be fixed first.

webknjaz:
The specific thing to fix is visible @ https://github.com/ansible/pylibssh/actions/runs/15679025783/job/44166024735?pr=744#step:7:712

webknjaz:
@KB-perByte why are you tweaking this in the first place?

pyproject.toml
```diff
@@ -184,7 +184,7 @@ LD_LIBRARY_PATH = "$(cat /root/.static-deps-path)/lib64:$(cat /root/.static-deps
 
 [tool.cibuildwheel.macos]
 before-build = [
-  "brew install libssh",  # @0.9.4  # pinning the version does not work
+  "MACOSX_DEPLOYMENT_TARGET=10.9 brew install --build-from-source libssh",  # @0.9.4  # pinning the version does not work
```
webknjaz, line 187: I've identified that `brew` drops env vars like `MACOSX_DEPLOYMENT_TARGET` and they never make it to the underlying compiler.

Could you try pasting instructions from https://github.com/ansible/pylibssh/blob/51f99dd/build-scripts/manylinux-container-image/install_libssh.sh#L74-L90 into this place instead? I don't know if we can reuse this script, which would be good. So I'd start by simply copying stuff.

Technically, you should be able to break into GHA workers via https://github.com/marketplace/actions/debugging-with-tmate to experiment. But bear in mind that those sessions expire in like an hour.

webknjaz: Alternatively, extract this into a separate experimental workflow and temporarily remove all others to get quick CI runs that don't start unnecessary jobs while you play with it. You don't need the entire `cibuildwheel` machinery to see if something compiles...

.github/workflows/reusable-cibuildwheel.yml
```diff
@@ -101,12 +101,15 @@ jobs:
...
-      uses: pypa/cibuildwheel@v2.17.0
+      uses: pypa/cibuildwheel@v2.21.1
       env:
         # CIBW_ARCHS_MACOS: all x86_64 arm64 universal2
+        CIBW_PRERELEASE_PYTHONS: 1
         CIBW_ARCHS_MACOS: native
+        CIBW_SKIP: "*_i686 *-macosx_universal2 *-musllinux_* *-win32 pp*"  # Removed *_arm64
         CIBW_CONFIG_SETTINGS: >-  # Cython line tracing for coverage collection
           with-cython-tracing=${{ inputs.cython-tracing }}
+        CIBW_ENVIRONMENT_MACOS: "CFLAGS=-I$(brew --prefix libssh)/include LDFLAGS=-L$(brew --prefix libssh)/lib"
```
webknjaz, line 112: This already works with homebrew, even w/o these. We need to throw away brew for good.

pyproject.toml
KB-perByte: MACOSX_DEPLOYMENT_TARGET yes, this makes no sense. Would be taking it out.

.github/workflows/reusable-cibuildwheel.yml
KB-perByte: @webknjaz, any pointers on how we proceed?

webknjaz: have you followed these pointers? https://github.com/ansible/pylibssh/pull/745#discussion_r2161214775

webknjaz: Also, exploring the macOS builds does not need any of the changes in this PR.

KB-perByte:
This PR is not needed!

PR #751, opened 2025-06-24
Enable python-3.13 for tests

KB-perByte:


KB-perByte:
@webknjaz 3.13 passed here https://github.com/ansible/pylibssh/actions/runs/15821795682/job/44592806600
I am checking why it's failing here.

webknjaz:
@KB-perByte that's Ubuntu, not macOS. Nothing changed for macOS. We still need to start compiling `libssh` properly.

Issue #222, opened 2021-06-03
SFTP issue with libssh


ashwini-mhatre:
##### SUMMARY
sftp is not working with libssh on iosxr platform

##### ISSUE TYPE
- Bug Report

##### PYLISSH and LIBSSH VERSION

```console
(py3.6.10) amhatre@ashwinis-MacBook-Pro playbooks % pip show ansible-pylibssh
Name: ansible-pylibssh
Version: 0.2.0
Summary: Python bindings for libssh client specific to Ansible use case
Home-page: https://github.com/ansible/pylibssh
Author: Ansible, Inc.
Author-email: info+github/ansible/pylibssh@ansible.com
License: LGPLv2+
Location: /Users/amhatre/ansible_venvs/py3.6.10/lib/python3.6/site-packages
Requires: 
Required-by: 
```

##### OS / ENVIRONMENT

IOSXR 6.1.3
IOSXR 7.0.2

##### STEPS TO REPRODUCE

```yaml
- name: test iosxr with pylibssh
  hosts: iosxr
  connection: ansible.netcommon.network_cli
  gather_facts: no
  tasks:
    - name: replace config
      register: result
      cisco.iosxr.iosxr_config:
        src: 'running_cfg_iosxr.txt'
        replace: config

```

##### EXPECTED RESULTS

```console
changed: [127.0.0.1] => {
    "changed": true,
    "commands": [
        "load harddisk:/ansible_config.txt"
    ],
    "diff": {
        "prepared": "Building configuration...\n!! IOS XR Configuration 7.0.2\n+  interface preconfigure GigabitEthernet0/0/0/0\n   !\n+  interface preconfigure GigabitEthernet0/0/0/1\n   !\n+  interface preconfigure GigabitEthernet0/0/0/2\n   !\nend"
    },
    "invocation": {
        "module_args": {
            "admin": false,
            "after": null,
            "backup": false,
            "backup_options": null,
            "before": null,
            "comment": "configured by iosxr_config",
            "config": null,
            "exclusive": false,
            "force": false,
            "label": null,
            "lines": null,
            "match": "line",
            "parents": null,
            "provider": null,
            "replace": "config",
[...]

PR #752, opened 2025-06-24
Update to libssh 0.11.2 in manylinux

Jakuje:
##### SUMMARY
The libssh 0.11.2 landed and among other things, it should fix the #222.

##### ISSUE TYPE
- Bugfix Pull Request


webknjaz:
@Jakuje let's add that “fixes” comment in the follow-up that actually makes use of this.

Jakuje:
Sorry, the Fixes was already in the commits so the #222 got closed. Will include it in the next PR too.

PR #754, opened 2025-06-25
Sync `reusable-cibuildwheel.yml` with upstream

webknjaz:
This patch includes setting input defaults for sdist-related fields and passing the sdist directly into `cibuildwheel`. It also normalizes the workflow name.


KB-perByte:
What is our upstream here? @webknjaz

webknjaz:
@KB-perByte nothing official just yet. It's a part of my effort to have a shared externally hosted reusable workflow for various things. In this case, for `cibuildwheel`. I'm still working on it but trying to keep multiple repos in sync to ease future migration.

Issue #699, opened 2025-04-07
No binary wheel for py313 published on pypi


ssbarnea:
##### SUMMARY

Latest version of `ansible-pylibssh` published on pypi is missing py313, for both amd64 and arm64 architectures and this break ability to install any ansible component that is requiring this library, like `ansible.netcommon` collection on python 3.13

Apparently trying to install from source does create some kind of deadlock on these platforms:

- MacOS 
- ubuntu 24.24 

These failures can be seen at https://github.com/ansible/ansible-dev-environment/actions/runs/14306408049

I mention that the jobs timeout after ~40 minutes due to GHA default setup.

```
TRACE Comparing installed with source: Registry(InstalledRegistryDist { name: PackageName("setuptools"), version: "78.1.0", path: "/Users/ssbarnea/.cache/uv/builds-v0/.tmpc8o1Rt/lib/python3.13/site-packages/setuptools-78.1.0.dist-info", cache_info: None }) Registry { specifier: VersionSpecifiers([VersionSpecifier { operator: Equal, version: "78.1.0" }]), index: Some(IndexMetadata { url: Pypi(VerbatimUrl { url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("pypi.org")), port: None, path: "/simple", query: None, fragment: None }, given: None }), format: Simple }), conflict: None }
DEBUG Requirement already installed: setuptools==78.1.0
DEBUG Installing build requirement: cython==3.1.0b1
TRACE Extracting file name=PackageName("cython")
TRACE Cloning /Users/ssbarnea/.cache/uv/archive-v0/14SPFoExWhnhqJwWCSInL/cython.py to /Users/ssbarnea/.cache/uv/builds-v0/.tmpc8o1Rt/lib/python3.13/site-packages/cython.py
TRACE Cloning /Users/ssbarnea/.cache/uv/archive-v0/14SPFoExWhnhqJwWCSInL/cython-3.1.0b1.dist-info to /Users/ssbarnea/.cache/uv/builds-v0/.tmpc8o1Rt/lib/python3.13/site-packages/cython-3.1.0b1.dist-info
TRACE Cloning /Users/ssbarnea/.cache/uv/archive-v0/14SPFoExWhnhqJwWCSInL/Cython to /Users/ssbarnea/.cache/uv/builds-v0/.tmpc8o1Rt/lib/python3.13/site-packages/Cython
[...]

PR #597, opened 2024-05-23
Fix the log levels mapping

Jakuje:
##### SUMMARY
The libssh provides the most verbose logging with SSH_LOG_TRACE, which was mapped to `logging.CRITICAL`, causing the users being unable to get full debug logs. These are critical for libssh developers to be able to investigate issues.

##### ISSUE TYPE
- Bugfix Pull Request

Issue #73, opened 2020-06-17
[TODO] Build libssh under macOS instead of using brew


webknjaz:
This would allow us to control the version of libssh (it's currently what brew ships and that's it — they remove old versions when adding new ones, hence no control at all).

Also, having control over the building allows us to single-source libssh version for manulinux+macos instead of hoping that they are the same.

webknjaz:
With the sunset of `macos-13` in CI, this is going to be more important to solve soon: https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/#macos-13-is-closing-down.

komaldesai13:
created shell script to build libssh artifacts for a specific architecture and merge both architecture to produce unvirsal2 build   , testing it locally for x86 on macos


https://github.com/ansible/pylibssh/pull/795

webknjaz:
@komaldesai13 could you turn it into a draft PR? It's easier to leave comments inline.

komaldesai13:
created cibuildwheel-before-build.sh script to use in cibuldwheel
scripts can be found here in [PR](https://github.com/ansible/pylibssh/pull/795)

 

ran cibuildwheel locally for macos universal2 artifacts as below


` python3 -m cibuildwheel --only cp312-macosx_universal2`

logs for cibuidwheel run can be found here

 

https://docs.google.com/document/d/1bBbjm-NEwSNevce_YwmMb8iX3HM2eiLJ_U4HI4K0ztI/edit?usp=sharing

webknjaz:
FYI, Google Docs (or similar editors that can't store plain text) isn't the best place to put the logs. Something like https://pastebin.com would work better. Or a file attached to a comment, or even just a snippet wrapped with [`<details>`](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections).

PR #756, opened 2025-07-30
Retry on libssh `SSH_AGAIN` return code

justin-stephenson:
##### SUMMARY
When a low SSH options timeout value is set, we see sometimes that calls to `new_channel()` and `ssh_channel_open_session`  fail when libssh returns `SSH_AGAIN`. Currently, pylibssh returns an exception:
```

../../../../pytest-mh/pytest_mh/conn/ssh.py:285: in _run
    self.__channel = self.__conn.new_channel()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^
src/pylibsshext/session.pyx:514: in pylibsshext.session.Session.new_channel
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>   ???
E   pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
src/pylibsshext/channel.pyx:71: LibsshChannelException
```

SSH_AGAIN return code is documented https://api.libssh.org/master/group__libssh__channel.html#gaf051dd30d75bf6dc45d1a5088cf970bd 

It is not clearly stated this but SSH_AGAIN also happens due to timeout.

```
ssh_channel_open_session()

Returns
    SSH_OK on success, SSH_ERROR if an error occurred, SSH_AGAIN if in nonblocking mode and call has to be done again.
```

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
This issue happens in our https://github.com/next-actions/pytest-mh project.

CC @pbrezina

### DIFF
--- a/docs/changelog-fragments/756.feature.1.rst
+++ b/docs/changelog-fragments/756.feature.1.rst
@@ -1,0 +1,8 @@
+Added a ``pylibsshext.session.connect()`` parameter
+``open_session_retries`` -- by :user:`justin-stephenson`.
+
+The ``open_session_retries`` session ``connect()``
+parameter allows a configurable number of retries if
+libssh ``ssh_channel_open_session()`` returns ``SSH_AGAIN``.
+The default option value is 0, no retries will be
+attempted.
--- a/docs/changelog-fragments/756.feature.2.rst
+++ b/docs/changelog-fragments/756.feature.2.rst
@@ -1,0 +1,8 @@
+Added a ``pylibsshext.session.connect()`` parameter
+``timeout_usec`` to set ``SSH_OPTIONS_TIMEOUT_USEC``.
+
+This allows setting the ``SSH_OPTIONS_TIMEOUT_USEC``
+ssh option, though ``SSH_OPTIONS_TIMEOUT`` is a more
+practical option.
+
+-- by :user:`justin-stephenson`
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -24,7 +24,7 @@ from libc.string cimport memset
 
 from pylibsshext.errors cimport LibsshChannelException
 from pylibsshext.errors import LibsshChannelReadFailure
-from pylibsshext.session cimport get_libssh_session
+from pylibsshext.session cimport get_libssh_session, get_session_retries
 
 from subprocess import CompletedProcess
 
@@ -63,12 +63,8 @@ cdef class Channel:
 
         if self._libssh_channel is NULL:
             raise MemoryError
-        rc = libssh.ssh_channel_open_session(self._libssh_channel)
 
-        if rc != libssh.SSH_OK:
-            libssh.ssh_channel_free(self._libssh_channel)
-            self._libssh_channel = NULL
-            raise LibsshChannelException("Failed to open_session: [%d]" % rc)
+        self._open_session_with_retries(self._libssh_channel)
 
     def __dealloc__(self):
         if self._libssh_channel is not NULL:
@@ -158,16 +154,28 @@ cdef class Channel:
             response = recv_buff.getvalue()
         return response
 
+    cdef _open_session_with_retries(self, libssh.ssh_channel channel):
+        retry = get_session_retries(self._session)
+
+        for attempt in range(retry + 1):
+            rc = libssh.ssh_channel_open_session(channel)
+            if rc == libssh.SSH_OK:
+                break
+            if rc == libssh.SSH_AGAIN and attempt < retry:
+                continue
+            # either SSH_ERROR, or SSH_AGAIN with final attempt
+            if rc != libssh.SSH_OK:
+                self._libssh_channel = NULL
+                libssh.ssh_channel_free(channel)
+                raise LibsshChannelException(f"Failed to open_session: [{rc}]")
+
     def exec_command(self, command):
         # request_exec requires a fresh channel each run, so do not use the existing channel
         cdef libssh.ssh_channel channel = libssh.ssh_channel_new(self._libssh_session)
         if channel is NULL:
             raise MemoryError
 
-        rc = libssh.ssh_channel_open_session(channel)
-        if rc != libssh.SSH_OK:
-            libssh.ssh_channel_free(channel)
-            raise LibsshChannelException("Failed to open_session: [{0}]".format(rc))
+        self._open_session_with_retries(channel)
 
         result = CompletedProcess(args=command, returncode=-1, stdout=b'', stderr=b'')
 
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -109,6 +109,7 @@ cdef class Session(object):
         self._hash_py = None
         self._fingerprint_py = None
         self._keytype_py = None
+        self._retries = 0
         # Due to delayed freeing of channels, some older libssh versions might expect
         # the callbacks to be around even after we free the underlying channels so
         # we should free them only when we terminate the session.
@@ -236,8 +237,16 @@ cdef class Session(object):
         file should be validated. It defaults to True
         :type host_key_checking: boolean
 
+        :param open_session_retries: The number of retries to attempt when libssh
+                                      channel function ``ssh_channel_open_session()`` returns ``SSH_AGAIN``. It defaults
+                                      to 0, no retries attempted.
+        :type open_session_retries: integer
+
         :param timeout: The timeout in seconds for the TCP connect
         :type timeout: long integer
+
+        :param timeout_usec: The timeout in microseconds for the TCP connect
+        :type timeout_usec: long integer
 
         :param port: The ssh server port to connect to
         :type port: integer
@@ -261,6 +270,9 @@ cdef class Session(object):
             except Exception:
                 libssh.ssh_disconnect(self._libssh_session)
                 raise
+
+        if 'open_session_retries' in kwargs:
+            self._retries = kwargs['open_session_retries']
 
         # We need to userauth_none before we can query the available auth types
         rc = libssh.ssh_userauth_none(self._libssh_session, NULL)
@@ -554,3 +566,6 @@ cdef class Session(object):
 
 cdef libssh.ssh_session get_libssh_session(Session session):
     return session._libssh_session
+
+cdef int get_session_retries(Session session):
+    return session._retries
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -31,6 +31,7 @@ OPTS_MAP = {
     "user": libssh.SSH_OPTIONS_USER,
     "port": libssh.SSH_OPTIONS_PORT,
     "timeout": libssh.SSH_OPTIONS_TIMEOUT,
+    "timeout_usec": libssh.SSH_OPTIONS_TIMEOUT_USEC,
     "knownhosts": libssh.SSH_OPTIONS_KNOWNHOSTS,
     "proxycommand": libssh.SSH_OPTIONS_PROXYCOMMAND,
     "key_exchange_algorithms": libssh.SSH_OPTIONS_KEY_EXCHANGE,
@@ -175,7 +176,7 @@ cdef class Session(object):
         elif key == "port":
             value_uint = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_uint)
-        elif key == "timeout":
+        elif key in {"timeout", "timeout_usec"}:
             value_long = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_long)
         else:
--- a/tests/unit/channel_test.py
+++ b/tests/unit/channel_test.py
@@ -8,12 +8,17 @@ import time
 
 import pytest
 
+from pylibsshext.errors import LibsshChannelException
 from pylibsshext.session import Session
 
 
 COMMAND_TIMEOUT = 30
 POLL_EXIT_CODE_TIMEOUT = 5
 POLL_TIMEOUT = 5000
+# Lowest possible timeout_usec is 1000 because timeout_usec / 1000 must be > 0,
+# otherwise libssh sets timeout to 10000
+SMALL_TIMEOUT_USEC = 1000
+LARGE_TIMEOUT_SEC = 10
 
 
 @pytest.fixture
@@ -31,6 +36,42 @@ def ssh_channel(ssh_client_session):
         yield chan
     finally:
         chan.close()
+
+
+def test_open_session_small_timeout(ssh_session_connect):
+    """Test opening a new channel with a small timeout value.
+
+    This generates an exception from ``ssh_channel_open_session()``
+    returning ``SSH_AGAIN`` with the ``usec`` timeout and default
+    ``open_session_retries`` value of ``0``.
+    """
+    ssh_session = Session()
+    ssh_session_connect(ssh_session)
+    ssh_session.set_ssh_options('timeout_usec', SMALL_TIMEOUT_USEC)
+    error_msg = '^Failed to open_session'
+    with pytest.raises(LibsshChannelException, match=error_msg):
+        ssh_session.new_channel()
+    ssh_session.close()
+
+
+def test_open_session_large_timeout(ssh_session_connect):
+    """Test opening a new channel with a large timeout value."""
+    ssh_session = Session()
+    ssh_session_connect(ssh_session)
+    ssh_session.set_ssh_options('timeout', LARGE_TIMEOUT_SEC)
+    ssh_channel = ssh_session.new_channel()
+    ssh_channel.close()
+    ssh_session.close()
+
+
+def test_open_session_small_timeout_with_retries(ssh_session_connect_retries):
+    """Test with a small timeout value and retries set."""
+    ssh_session = Session()
+    ssh_session_connect_retries(ssh_session)
+    ssh_session.set_ssh_options('timeout_usec', SMALL_TIMEOUT_USEC)
+    ssh_channel = ssh_session.new_channel()
+    ssh_channel.close()
+    ssh_session.close()
 
 
 def exec_second_command(ssh_channel):
--- a/src/pylibsshext/channel.pxd
+++ b/src/pylibsshext/channel.pxd
@@ -23,6 +23,7 @@ cdef class Channel:
     cdef  _session
     cdef libssh.ssh_channel _libssh_channel
     cdef libssh.ssh_session _libssh_session
+    cdef _open_session_with_retries(self, libssh.ssh_channel test_channel)
 
 cdef class ChannelCallback:
     cdef callbacks.ssh_channel_callbacks_struct callback
--- a/src/pylibsshext/session.pxd
+++ b/src/pylibsshext/session.pxd
@@ -26,6 +26,8 @@ cdef class Session:
     cdef _hash_py
     cdef _fingerprint_py
     cdef _keytype_py
+    cdef _retries
     cdef _channel_callbacks
 
 cdef libssh.ssh_session get_libssh_session(Session session)
+cdef int get_session_retries(Session session)
--- a/tests/_service_utils.py
+++ b/tests/_service_utils.py
@@ -69,6 +69,7 @@ def wait_for_svc_ready_state(
 
 def ensure_ssh_session_connected(  # noqa: WPS317
         ssh_session, sshd_addr, ssh_clientkey_path,  # noqa: WPS318
+        ssh_session_retries=0,
 ):
     """Attempt connecting to the SSH server until successful.
 
@@ -89,4 +90,5 @@ def ensure_ssh_session_connected(  # noqa: WPS317
         private_key=ssh_clientkey_path.read_bytes(),
         host_key_checking=False,
         look_for_keys=False,
+        open_session_retries=ssh_session_retries,
     )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -127,6 +127,25 @@ def ssh_client_session(ssh_session_connect):
 
 
 @pytest.fixture
+def ssh_session_connect_retries(sshd_addr, ssh_clientkey_path):
+    """
+    Authenticate existing session object against SSHD with a private SSH key.
+
+    This sets ``open_session_retries=100`` and it returns a function
+    that takes session as parameter.
+
+    :returns: Function that will connect the session.
+    :rtype: Callback
+    """
+    return partial(
+        ensure_ssh_session_connected,
+        sshd_addr=sshd_addr,
+        ssh_clientkey_path=ssh_clientkey_path,
+        ssh_session_retries=100,
+    )
+
+
+@pytest.fixture
 def ssh_session_connect(sshd_addr, ssh_clientkey_path):
     """
     Authenticate existing session object against SSHD with a private SSH key.


docs/changelog-fragments/756.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+The ``Channel`` class calls to libssh ``ssh_channel_open_session`` and
+``ssh_userauth_password`` are now retried when libssh returns SSH_AGAIN.
+:user:`justin-stephenson`.
```
webknjaz:
```suggestion
``ssh_userauth_password`` are now retried when ``libssh`` returns ``SSH_AGAIN``
-- by :user:`justin-stephenson`.
```

docs/changelog-fragments/756.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+The ``Channel`` class calls to libssh ``ssh_channel_open_session`` and
```
webknjaz: I'm pretty sure classes don't call things. They represent states. Could you rephrase?

webknjaz (changes requested):
I wonder if this could result in an infinite loop..

Also, I don't think there's a proof that this works as intended without tests. Add them.

Jakuje:
> I wonder if this could result in an infinite loop..

Yes, this would go into infinite loop if the server dies and does not properly disconnect. And timeouts are to handle this issue.

Retrying unconditionally and infinitely is ok for tests, but for real-world application, the pylibssh should do at very least some check with `ssh_is_connected()` or something.

Or setting some limit how many times you could retry. But in this case, why not raise the timeout itself?

justin-stephenson:
Thank you for your comments and review.

> I wonder if this could result in an infinite loop..
> 
> Also, I don't think a proof that this works as intended without tests. Add them.

I tried to add a test for this but unfortunately I couldn't reproduce the scenario we see in the test environment. I'll try again.

> > I wonder if this could result in an infinite loop..
> 
> Yes, this would go into infinite loop if the server dies and does not properly disconnect. And timeouts are to handle this issue.
> 
> Retrying unconditionally and infinitely is ok for tests, but for real-world application, the pylibssh should do at very least some check with `ssh_is_connected()` or something.
> 
> Or setting some limit how many times you could retry. But in this case, why not raise the timeout itself?

I can change this PR to add a call to `ssh_is_connected()` to avoid an infinite loop, or I can raise a different exception when SSH_AGAIN is returned (like `LibsshChannelAgain`) then we will handle this exception in our calls to pylibssh methods. 

Whichever you prefer is acceptable for us, just let me know and i'll make those changes.

Jakuje:
I am actually wondering how you are getting the SSH_AGAIN in these two places with pylibssh. The sessions in libssh are blocking by default. The only way to change the session to non-blocking mode is to use `ssh_set_blocking()` or doing some variation of `ssh_channel_read_nonblocking()`, but I see your changes completely elsewhere, this should not come into the effect and I do not see these functions exposed in the pylibssh either.

But there might be the oddness that setting low timeout might actually return the SSH_AGAIN in places where it should not, according to the documentation, which would be a bug in libssh that needs to be fixed.

What brought you initially to set smaller timeouts? Is a viable workaround to raise the timeouts?

justin-stephenson:
> I am actually wondering how you are getting the SSH_AGAIN in these two places with pylibssh. The sessions in libssh are blocking by default. The only way to change the session to non-blocking mode is to use `ssh_set_blocking()` or doing some variation of `ssh_channel_read_nonblocking()`, but I see your changes completely elsewhere, this should not come into the effect and I do not see these functions exposed in the pylibssh either.

The error we currently see in our PRCI is specific to `ssh_channel_open_session` failure:

`	FAILED tests/test_authentication.py::test_authentication__user_login_with_overriding_home_directory[domain] (ldap) - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]`

I added the `session: ` commit just as a nice to have because `ssh_userauth_password()` can return `SSH_AGAIN` per the libssh API docs, but the `channel.pyx` commit is the main issue we are hitting currently.

> 
> But there might be the oddness that setting low timeout might actually return the SSH_AGAIN in places where it should not, according to the documentation, which would be a bug in libssh that needs to be fixed.
> 
> What brought you initially to set smaller timeouts? Is a viable workaround to raise the timeouts?

In our code we set `.set_ssh_options("timeout", 1)` because in our `pytest-mh` code we allow users to to execute commands over SSH on hosts with an arbitrary timeout value set, such as:

`     client.host.conn.run(..., timeout=X)`

If I understand correctly, setting this low `set_ssh_options("timeout")"` value is necessary for the above to work as expected because Python will not deliver signal if the code is blocked in C library The signal is delivered only after we get back to the Python code. @pbrezina can correct me here.

-- for reference https://github.com/next-actions/pytest-mh/blob/master/pytest_mh/conn/ssh.py

Jakuje:
Ok, setting the libssh timeout is the timeout you are giving to the libssh to return to you. but if you are setting the low timeout to get the signals delivered, then either pylibssh or the caller needs to retry. The pylibssh code is really not written to support the retries around here so my proposal would be to create some pylibssh timeout/retry counter to avoid infinite cycle when stuff will go wrong. What do you think?

It can be either separate pyblissh option, or it can be somehow intercepted when we set the libssh timeout to set it to some multiply of the user specified value to return the handling to the python code. Or the second option by default with possible override.

And obviously we need some tests with this option, otherwise its untested broken code. I bet we can get some slow CI runners where this would demonstrate from time to time.

justin-stephenson:
> Ok, setting the libssh timeout is the timeout you are giving to the libssh to return to you. but if you are setting the low timeout to get the signals delivered, then either pylibssh or the caller needs to retry. The pylibssh code is really not written to support the retries around here so my proposal would be to create some pylibssh timeout/retry counter to avoid infinite cycle when stuff will go wrong. What do you think?
> 
> It can be either separate pyblissh option, or it can be somehow intercepted when we set the libssh timeout to set it to some multiply of the user specified value to return the handling to the python code. Or the second option by default with possible override.

I went ahead and updated the PR with your suggestion, by adding a new option `open_session_retries` which can be provided to the `connect()` method in `src/pylibsshext/session.pyx`. To make it less invasive I I set the default value for this to 0, so default libssh behavior will not change. Please take a look.
> 
> And obviously we need some tests with this option, otherwise its untested broken code. I bet we can get some slow CI runners where this would demonstrate from time to time.

I added some test scaffolding that is not done yet. In the test environment I always see `ssh_channel_open_session()`  returns `SSH_OK` instead of `SSH_AGAIN` even with setting a low timeout therefore I don't see how to test the retries properly in the test environment.

Jakuje:
> I added some test scaffolding that is not done yet. In the test environment I always see ssh_channel_open_session() returns SSH_OK instead of SSH_AGAIN even with setting a low timeout therefore I don't see how to test the retries properly in the test environment.

The libssh has an option `SSH_OPTIONS_TIMEOUT_USEC` to set subsecond timeouts so if this could help reproducing the issue (or spending less time in c code), we can either expose this option, or do again some mangling to allow float input on the python side and then convert it to `SSH_OPTIONS_TIMEOUT` or `SSH_OPTIONS_TIMEOUT_USEC`. Not sure what would be nicer/easier from the API point of view, but I think the second approach sounds mostly transparent for users.

https://api.libssh.org/master/group__libssh__session.html#ga7a801b85800baa3f4e16f5b47db0a73d

webknjaz:
I clicked "rebase" so this PR pulls in the CI fixes.

webknjaz:
The change note was lost in your last force-push.

justin-stephenson:
> > I added some test scaffolding that is not done yet. In the test environment I always see ssh_channel_open_session() returns SSH_OK instead of SSH_AGAIN even with setting a low timeout therefore I don't see how to test the retries properly in the test environment.
> 
> The libssh has an option `SSH_OPTIONS_TIMEOUT_USEC` to set subsecond timeouts so if this could help reproducing the issue (or spending less time in c code), we can either expose this option, or do again some mangling to allow float input on the python side and then convert it to `SSH_OPTIONS_TIMEOUT` or `SSH_OPTIONS_TIMEOUT_USEC`. Not sure what would be nicer/easier from the API point of view, but I think the second approach sounds mostly transparent for users.
> 
> https://api.libssh.org/master/group__libssh__session.html#ga7a801b85800baa3f4e16f5b47db0a73d

Thank you. I was able to add 2 tests for this with subsecond timeout: 1 which reproduces the `LibsshChannelException: Failed to open_session:` exception and 1 which fixes this by setting the new `open_session_retries` value. I updated the PR it is now ready for review.

justin-stephenson:
> The change note was lost in your last force-push.

It is added now.

Jakuje (approved):
I think this is conceptually going in the right direction. It will need some polishing to make the CI and  happy. I would also try to consider if other places where this might happen (from my understanding this could happen almost everywhere, but I might be wrong. If I am not, we will need more generic description than we have now).
At least for what we have a coverage could be tried by setting small timeout in the test fixture `ssh_session_connect()` to see where it will fail.

justin-stephenson:
> I think this is conceptually going in the right direction. It will need some polishing to make the CI and happy. I would also try to consider if other places where this might happen (from my understanding this could happen almost everywhere, but I might be wrong. If I am not, we will need more generic description than we have now). At least for what we have a coverage could be tried by setting small timeout in the test fixture `ssh_session_connect()` to see where it will fail.

According to libssh API docs the Channel `request_*` methods in https://github.com/ansible/pylibssh/blob/devel/src/pylibsshext/channel.pyx may return SSH_AGAIN, but I am not sure if it makes sense to add similar retry logic to these methods if there is no real world evidence that it helps, only speculation. I will defer to your judgement however. If retry logic needs to be added elsewhere then I would likely need to re-work this PR to create a `@retry` python decorator which can be used from multiple places.

In the test environment, all tests still succeed with setting a `1` second timeout. setting SSH_OPTIONS_TIMEOUT_USEC to `10000` in `ssh_client_session` causes the following tests to fail but it is quite a small timeout value therefore I don't think it is a representation of what may happen to pylibssh users.

```
ERROR tests/unit/channel_test.py::test_exec_command_stderr - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_is_eof - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_request_exec - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_send_eof - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_exec_command - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_channel_exit_status - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_double_close - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/channel_test.py::test_send_signal - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/sftp_test.py::test_make_sftp - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/channel_test.py::test_recv_eof - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
ERROR tests/unit/sftp_test.py::test_put_existing[small-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_put_existing[large-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_put[small-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_put[large-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_get[small-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_get_existing[large-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_get_existing[small-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
ERROR tests/unit/sftp_test.py::test_get[large-payload] - pylibsshext.errors.LibsshSFTPException: Failed to create new session
FAILED tests/unit/scp_test.py::test_get[small-payload] - pylibsshext.errors.LibsshSCPException: Error receiving information about file: b'SCP: ssh_channel_read timeout'
FAILED tests/unit/scp_test.py::test_put[small-payload] - pylibsshext.errors.LibsshSCPException: Initializing SCP session of remote file [/tmp/pytest-of-justin/pytest-153/popen-gw7/test_put_small_payload_0/dst-file.txt] for write failed with error [b'SCP: ssh_channel_read timeout']
FAILED tests/unit/scp_test.py::test_get_existing_local[large-payload] - pylibsshext.errors.LibsshSCPException: Error receiving information about file: b'SCP: ssh_channel_read timeout'
FAILED tests/unit/scp_test.py::test_get_existing_local[small-payload] - pylibsshext.errors.LibsshSCPException: Error receiving information about file: b'SCP: ssh_channel_read timeout'
FAILED tests/unit/channel_test.py::test_read_bulk_response - pylibsshext.errors.LibsshChannelException: Failed to open_session: [-2]
FAILED tests/unit/scp_test.py::test_get[large-payload] - pylibsshext.errors.LibsshSCPException: Error receiving information about file: b'SCP: ssh_channel_read timeout'
FAILED tests/unit/scp_test.py::test_put[large-payload] - pylibsshext.errors.LibsshSCPException: Initializing SCP session of remote file [/tmp/pytest-of-justin/pytest-153/popen-gw7/test_put_large_payload_0/dst-file.txt] for write failed with error [b'SCP: ssh_channel_read timeout']
```

docs/changelog-fragments/756.feature.rst
webknjaz, line 1: Looks like this has two separate change notes smashed into the same file. They should go into separate files. Also, one variable/indentifier seems to have missing backticks.

src/pylibsshext/channel.pyx
webknjaz, line 1: I don't like that this adds a lot of copypaste. Can it be moved into a helper?

src/pylibsshext/channel.pyx
```diff
@@ -63,12 +63,20 @@ cdef class Channel:
...
+            rc = libssh.ssh_channel_open_session(self._libssh_channel)
+            if rc == libssh.SSH_OK:
+                break
+            if rc == libssh.SSH_AGAIN and attempt < retry:
+                continue
+            # either SSH_ERROR, or SSH_AGAIN with final attempt
+            if rc != libssh.SSH_OK:
+                libssh.ssh_channel_free(self._libssh_channel)
+                self._libssh_channel = NULL
+                raise LibsshChannelException("Failed to open_session: [%d]" % rc)
```
webknjaz: can't this use an f-string?

src/pylibsshext/channel.pyx
```diff
@@ -164,10 +172,18 @@ cdef class Channel:
...
+        for attempt in range(retry + 1):
+            rc = libssh.ssh_channel_open_session(channel)
+            if rc == libssh.SSH_OK:
+                break
+            if rc == libssh.SSH_AGAIN and attempt < retry:
+                continue
+            # either SSH_ERROR, or SSH_AGAIN with final attempt
+            if rc != libssh.SSH_OK:
+                libssh.ssh_channel_free(channel)
+                raise LibsshChannelException("Failed to open_session: [{0}]".format(rc))
```
webknjaz: f-string?

src/pylibsshext/session.pyx
```diff
@@ -175,7 +177,7 @@ cdef class Session(object):
         elif key == "port":
             value_uint = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_uint)
-        elif key == "timeout":
+        elif key == "timeout" or key == "timeout_usec":
```
webknjaz:
```suggestion
        elif key in {"timeout", "timeout_usec"}:
```

src/pylibsshext/session.pyx
```diff
@@ -261,6 +271,9 @@ cdef class Session(object):
                 libssh.ssh_disconnect(self._libssh_session)
                 raise
 
+        if kwargs.get('open_session_retries'):
+            self._retries = kwargs.get('open_session_retries')
```
webknjaz: What if it's `0` or something falsy but is still set?
```suggestion
        if 'open_session_retries' in kwargs:
            self._retries = kwargs['open_session_retries']
```

src/pylibsshext/session.pyx
```diff
@@ -235,9 +237,17 @@ cdef class Session(object):
         file should be validated. It defaults to True
         :type host_key_checking: boolean
 
+        :param open_session_retries: The number of retries to attempt when libssh
+        channel function ssh_channel_open_session() returns SSH_AGAIN. It defaults
+        to 0, no retries attempted.
```
webknjaz: This should probably be indented and use proper RST
```suggestion
                                      channel function ``ssh_channel_open_session()`` returns ``SSH_AGAIN``. It defaults
                                      to 0, no retries attempted.
```

src/pylibsshext/channel.pyx
justin-stephenson: Done, see `_open_session_with_retries()`

docs/changelog-fragments/756.feature.rst
justin-stephenson: Fixed.

src/pylibsshext/channel.pyx
justin-stephenson: Yes, fixed.

src/pylibsshext/channel.pyx
justin-stephenson: Yes, fixed.

Jakuje (approved):
thank you! Looks good to me

docs/changelog-fragments/756.feature.1.rst
```diff
@@ -0,0 +1 @@
+Added a ``pylibsshext.session.connect()`` parameter ``open_session_retries`` -- by :user:`justin-stephenson`.
```
webknjaz: Could you add a paragraph expanding on the user-facing effect?

tests/unit/channel_test.py
```diff
@@ -33,6 +35,34 @@ def ssh_channel(ssh_client_session):
...
+    """Test with a low timeout value and retries set.
+
+    This sets 'open_session_retries=10' and with the retries
+    ssh_channel_open_session() will succeed.
+    """
+    ssh_session = Session()
+    ssh_session_connect_retries(ssh_session)
+    ssh_session.set_ssh_options('timeout_usec', LOW_TIMEOUT_USEC)
+    ssh_channel = ssh_session.new_channel()
+    ssh_channel.close()
```
webknjaz, line 73: This probably needs a try-finally. And maybe close the session too.

tests/unit/channel_test.py
```diff
@@ -33,6 +35,34 @@ def ssh_channel(ssh_client_session):
...
 
 
+def test_open_session_timeout(ssh_session_connect):
+    """Test opening a new channel with a low timeout value.
+
+    This generates an exception from ssh_channel_open_session()
+    returning SSH_AGAIN with the usec timeout and default
+    open_session_retries value of 0.
+    """
+    ssh_session = Session()
```
webknjaz, line 48: It doesn't seem like the session gets closed in here.

tests/unit/channel_test.py
webknjaz: cc @Jakuje

tests/unit/channel_test.py
webknjaz: cc @Jakuje

docs/changelog-fragments/756.feature.1.rst
justin-stephenson: Added.

tests/unit/channel_test.py
justin-stephenson: Updated.

tests/unit/channel_test.py
justin-stephenson: Added the .close()

tests/unit/channel_test.py
justin-stephenson: @webknjaz Linter complains about this. Is some change needed here still? I'm not clear on the way to fix this.
```

  63:5     WPS501 Found `finally` in `try` block without `except`
  try:
  ^

WPS501: Found `finally` in `try` block without `except`
  1     tests/unit/channel_test.py
Total: 1
```

Jakuje: See https://wemake-python-styleguide.readthedocs.io/en/latest/pages/usage/violations/refactoring.html#wemake_python_styleguide.violations.refactoring.UselessFinallyViolation
I am not a python developer so not sure if the `ssh_channel` can work with the `with` (likely not?) so I see we either need to waive this warning or add dummy `except:` to make the linter happy.

justin-stephenson: Okay I added `# noqa: WPS501`  at the `try` line

justin-stephenson: Now it complains `WPS402 — Forbid too many # noqa comments.`

justin-stephenson:
Hi @webknjaz It would be great to have this merged soon, please take a look when you have a chance. Thanks for your help and reviews by the way.

justin-stephenson:
> Hi @webknjaz It would be great to have this merged soon, please take a look when you have a chance. Thanks for your help and reviews by the way.

Also, after this is merged do you have any estimation of when you will issue a new release of pylibssh ?

.github/workflows/build-manylinux-container-images.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           YEAR: _2_31  # There are no base images prior to 2.31 for this arch
 
     env:
-      LIBSSH_VERSION: 0.11.2
+      LIBSSH_VERSION: 0.11.3
```
webknjaz: This shouldn't be changed in this PR. It'll only confuse you into thinking that it has any effect on your CI run but it doesn't. You'll just achieve triggering more pointless jobs while working on the actual change.

docs/changelog-fragments/756.feature.1.rst
```diff
@@ -0,0 +1,8 @@
+Added a ``pylibsshext.session.connect()`` parameter
+``open_session_retries`` -- by :user:`justin-stephenson`.
+
+The ``open_session_retries`` session ``connect()``
+parameter allows a configurable number of retries if
+libssh ``ssh_channel_open_session()`` returns SSH_AGAIN.
```
webknjaz:
```suggestion
libssh ``ssh_channel_open_session()`` returns ``SSH_AGAIN``.
```

docs/changelog-fragments/756.feature.2.rst
webknjaz, line 1: More detail on the end-user effect here as well?

docs/changelog-fragments/765.contrib.rst
webknjaz, line 1: Submit separately.

src/pylibsshext/channel.pyx
```diff
@@ -158,16 +154,28 @@ cdef class Channel:
...
+        for attempt in range(retry + 1):
+            rc = libssh.ssh_channel_open_session(channel)
+            if rc == libssh.SSH_OK:
+                break
+            if rc == libssh.SSH_AGAIN and attempt < retry:
+                continue
+            # either SSH_ERROR, or SSH_AGAIN with final attempt
+            if rc != libssh.SSH_OK:
+                libssh.ssh_channel_free(channel)
+                self._libssh_channel = NULL
```
webknjaz: Should we swap these? It feels safer to drop the ref first:
```suggestion
                self._libssh_channel = NULL
                libssh.ssh_channel_free(channel)
```

src/pylibsshext/channel.pyx
webknjaz: Ack. Marking the thread as resolved.

tests/unit/channel_test.py
webknjaz: Session doesn't have a context manager interface. Neither does channel. It'd be good if they did but that's out of the scope here. We could wrap things with `@contextlib.contextmanager` but in tests it should be fine to just drop that try/except for now (keeping the `close()` calls).

webknjaz:
> > Hi @webknjaz It would be great to have this merged soon, please take a look when you have a chance. Thanks for your help and reviews by the way.
> 
> Also, after this is merged do you have any estimation of when you will issue a new release of pylibssh ?

I was planning to have the next release blocked on #699 (which is blocked on #73 blocking #744). Though, perhaps I could just release what we have so far -- the infra for that seems healthy atm.

docs/changelog-fragments/765.contrib.rst
webknjaz: Unless.. It's tightly coupled, is it? It's probably not since the CI is passing with `0.11.2` in this PR.

docs/changelog-fragments/756.feature.2.rst
justin-stephenson: This was only needed to allow setting a microsecond timeout, making it possible to reproduce the issue we were seeing in pylibssh tests added in this PR. As it is a [SSH option](https://api.libssh.org/stable/group__libssh__session.html#ga7a801b85800baa3f4e16f5b47db0a73d) which can be set I exposed it the same way as `SSH_OPTIONS_TIMEOUT`.

```
OPTS_MAP = {
    "timeout": libssh.SSH_OPTIONS_TIMEOUT,
    "timeout_usec": libssh.SSH_OPTIONS_TIMEOUT_USEC,
```

I updated the text to add `This allows setting the ``SSH_OPTIONS_TIMEOUT_USEC`` ssh option, though ``SSH_OPTIONS_TIMEOUT`` is a more practical option.`

Feel free to edit as you see fit.

docs/changelog-fragments/765.contrib.rst
justin-stephenson: It is needed for tests in `session: Add 'open_session_retries' option` to run. It is independent from the `756.feature.1.rst` changelog fragment however.

`+    ssh_session.set_ssh_options('timeout_usec', LOW_TIMEOUT_USEC)`

src/pylibsshext/channel.pyx
justin-stephenson: Swapped.

docs/changelog-fragments/756.feature.1.rst
justin-stephenson: Fixed.

.github/workflows/build-manylinux-container-images.yml
justin-stephenson: I dropped this commit which I pulled in by accident during rebase.

tests/unit/channel_test.py
justin-stephenson: Okay I dropped the try/except

docs/changelog-fragments/756.feature.2.rst
```diff
@@ -0,0 +1,6 @@
+Added a ``pylibsshext.session.connect()`` parameter
+``timeout_usec`` to set ``SSH_OPTIONS_TIMEOUT_USEC``.
+This allows setting the ``SSH_OPTIONS_TIMEOUT_USEC``
+ssh option, though ``SSH_OPTIONS_TIMEOUT`` is a more
+practical option.
+-- by :user:`justin-stephenson`.
```
webknjaz:
```suggestion

-- by :user:`justin-stephenson`
```

docs/changelog-fragments/756.feature.2.rst
```diff
@@ -0,0 +1,6 @@
+Added a ``pylibsshext.session.connect()`` parameter
+``timeout_usec`` to set ``SSH_OPTIONS_TIMEOUT_USEC``.
```
webknjaz, line 2:
```suggestion
``timeout_usec`` to set ``SSH_OPTIONS_TIMEOUT_USEC``.

```

tests/unit/channel_test.py
webknjaz: Alright. Marking as resolved.

tests/unit/channel_test.py
```diff
@@ -33,6 +35,36 @@ def ssh_channel(ssh_client_session):
         chan.close()
 
 
+def test_open_session_timeout(ssh_session_connect):
+    """Test opening a new channel with a low timeout value.
+
+    This generates an exception from ssh_channel_open_session()
+    returning SSH_AGAIN with the usec timeout and default
+    open_session_retries value of 0.
```
webknjaz:
```suggestion
    This generates an exception from ``ssh_channel_open_session()``
    returning ``SSH_AGAIN`` with the ``usec`` timeout and default
    ``open_session_retries`` value of ``0``.
```

tests/unit/channel_test.py
```diff
@@ -33,6 +35,36 @@ def ssh_channel(ssh_client_session):
...
+    with pytest.raises(LibsshChannelException, match=error_msg):
+        ssh_session.new_channel()
+    ssh_session.close()
+
+
+def test_open_session_with_retries(ssh_session_connect_retries):
+    """Test with a low timeout value and retries set.
+
+    This sets 'open_session_retries=10' and with the retries
+    ssh_channel_open_session() will succeed.
```
webknjaz:
```suggestion
    This sets ``open_session_retries=10`` and with the retries
    ``ssh_channel_open_session()`` will succeed.
```

tests/unit/channel_test.py
```diff
@@ -8,12 +8,14 @@
...
 import pytest
 
+from pylibsshext.errors import LibsshChannelException
 from pylibsshext.session import Session
 
 
 COMMAND_TIMEOUT = 30
 POLL_EXIT_CODE_TIMEOUT = 5
 POLL_TIMEOUT = 5000
+LOW_TIMEOUT_USEC = 10000
```
webknjaz: Why is this 10k? Can it be just 1 or something really low? This could use an explanatory code comment.

justin-stephenson: I added a comment with explanation.

justin-stephenson: In short, setting the subsecond timeout to some lower values would cause other tests to fail, and setting a higher would not reproduce the SSH_AGAIN. In our SSSD test environment, a 1 second timeout (not subsecond) would reproduce the SSH AGAIN failure for some tests, but in pylibssh test environment it needs to be a specific subsecond value which can reproduce the error.

Jakuje: That sounds quite fragile if raising it to some arbitrary value is not reproducing and lowering is breaking elsewhere.

I would rather have ridiculously small timeout with large amount of retries (which would basically switch libssh into non-blocking mode) and fix the other cases that break.

justin-stephenson: I was incorrectly remembering about other tests failing. 

I tried again and if I change `LOW_TIMEOUT_USEC` to `1` it causes only `test_open_session_timeout` to fail because the expected exception is not raised `LibsshChannelException Failed to open session` .

The test actually passes with `LOW_TIMEOUT_USEC` as `1` with the following diff.
```diff

--- a/tests/unit/channel_test.py
+++ b/tests/unit/channel_test.py
@@ -19,7 +19,7 @@ POLL_TIMEOUT = 5000
 # value for the test environment to allow test_open_session_with_retries()
 # to succeed with 10 attempted retries, and also fail with no retries in
 # test_open_session_timeout().
-LOW_TIMEOUT_USEC = 10000
+LOW_TIMEOUT_USEC = 1
 
 
 @pytest.fixture
@@ -50,8 +50,7 @@ def test_open_session_timeout(ssh_session_connect):
     ssh_session_connect(ssh_session)
     ssh_session.set_ssh_options('timeout_usec', LOW_TIMEOUT_USEC)
     error_msg = '^Failed to open_session'
-    with pytest.raises(LibsshChannelException, match=error_msg):
-        ssh_session.new_channel()
+    ssh_session.new_channel()
     ssh_session.close()
```

Therefore the reason to set `10000` was to actually trigger the LibsshChannelException (SSH_AGAIN) we are seeing.

Let me know if you want some changes here.

tests/unit/channel_test.py
```diff
@@ -33,6 +39,36 @@ def ssh_channel(ssh_client_session):
...
+    error_msg = '^Failed to open_session'
+    with pytest.raises(LibsshChannelException, match=error_msg):
+        ssh_session.new_channel()
+    ssh_session.close()
+
+
+def test_open_session_with_retries(ssh_session_connect_retries):
+    """Test with a low timeout value and retries set.
+
+    This sets ``open_session_retries=10`` and with the retries
```
webknjaz: This should probably go into the fixture that is what actually sets the value.

tests/unit/channel_test.py
webknjaz: Sounds like different cases are reproducible with different timeout values and relying on the same  constant makes the tests fragile and unnecessarily coupled.

And I still don't understand why lower timeout values don't trigger problems ­could you explain this bit?

webknjaz:
@Jakuje any idea what might be causing these timeouts? https://github.com/ansible/pylibssh/actions/runs/18018035693/job/51268709820?pr=756#step:17:101

tests/unit/channel_test.py
justin-stephenson: > Sounds like different cases are reproducible with different timeout values and relying on the same constant makes the tests fragile and unnecessarily coupled.

I agree.
> 
> And I still don't understand why lower timeout values don't trigger problems ­could you explain this bit?

I honestly don't know and I don't really have knowledge of libssh enough to dig into it.

tests/unit/channel_test.py
justin-stephenson: Done.

Jakuje:
> @Jakuje any idea what might be causing these timeouts? https://github.com/ansible/pylibssh/actions/runs/18018035693/job/51268709820?pr=756#step:17:101

The timeouts are in the `ssh_clientkey_path()`, which generates 8k RSA test keys. Does it fail reproducibly or just randomly? Can you try with #768, which is using ecdsa keys for tests, which are more lightweight?

webknjaz:
> > @Jakuje any idea what might be causing these timeouts? https://github.com/ansible/pylibssh/actions/runs/18018035693/job/51268709820?pr=756#step:17:101
> 
> The timeouts are in the `ssh_clientkey_path()`, which generates 8k RSA test keys. Does it fail reproducibly or just randomly? Can you try with #768, which is using ecdsa keys for tests, which are more lightweight?

Thanks, merged that one!

webknjaz:
I clicked `rebase` to see if that other PR helped with this.

tests/unit/channel_test.py
webknjaz: @Jakuje do you think it's reasonable to use separate constants for tests? Any other insight?

Jakuje: If we want to simulate long latency networks or slow system, we need to test with aggressively small timeouts to invoke the code paths that return AGAIN. In libssh we are trying to have API tested in both blocking and non-blocking mode. This is basically what is exposed to pylibssh user (default is blocking, but setting timeout switches some code to non-blocking at cases. That said, I would prefer to have tests with both smaller and larger timeouts to make sure they keep working under these conditions. But there might be some bugs to catch ...

justin-stephenson: @Jakuje Can you provide example small/larger timeout values (timeout or timeout_usec) you would like to see tested? I can try to add tests for them.

Jakuje: good small number sounds like number 1 and good large probably 1 second (1_000_000 micro seconds) or even 10 seconds (10_000_000 micro seconds). I worry that setting something larger will make the test timeouts too large when stuff will go south.

justin-stephenson: > 
> And I still don't understand why lower timeout values don't trigger problems ­could you explain this bit?

The issue I realized is that setting timeout_usec to `1`, `10`, or `100` will not work due to https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/client.c?ref_type=heads#L643-646 therefore `1000` is the lowest timeout_usec value which can be set.

justin-stephenson: I renamed and updated the test `test_open_session_small_timeout` to use `SMALL_TIMEOUT_USEC = 1000` which is the lowest possible timeout to set (see above comment).

I also added a test `test_open_session_large_timeout` which sets `timeout` value to `10` seconds with `LARGE_TIMEOUT_SEC = 10`

Please check these updates when you are able to.

Jakuje: Good point. In that case, please add a comment in the code to justify the used numbers. Thanks for digging this up!

justin-stephenson: > Good point. In that case, please add a comment in the code to justify the used numbers. Thanks for digging this up!

I did already, thanks. Also with `timeout_usec` set to `1000` I had to bump up retries to `100` for the small timeout test to succeed

webknjaz (approved):
Thanks @justin-stephenson @Jakuje for polishing this to a great shape!

tests/unit/channel_test.py
```diff
@@ -33,6 +38,42 @@ def ssh_channel(ssh_client_session):
...
+    """Test opening a new channel with a large timeout value."""
+    ssh_session = Session()
+    ssh_session_connect(ssh_session)
+    ssh_session.set_ssh_options('timeout', LARGE_TIMEOUT_SEC)
+    ssh_channel = ssh_session.new_channel()
+    ssh_channel.close()
+    ssh_session.close()
+
+
+def test_open_session_small_timeout_with_retries(ssh_session_connect_retries):
```
webknjaz, line 67: @justin-stephenson this test failed at least on one occasion post-merge: https://github.com/ansible/pylibssh/actions/runs/18431666269#summary-52519922027

tests/unit/channel_test.py
```diff
@@ -33,6 +38,42 @@ def ssh_channel(ssh_client_session):
...
+
+    This generates an exception from ``ssh_channel_open_session()``
+    returning ``SSH_AGAIN`` with the ``usec`` timeout and default
+    ``open_session_retries`` value of ``0``.
+    """
+    ssh_session = Session()
+    ssh_session_connect(ssh_session)
+    ssh_session.set_ssh_options('timeout_usec', SMALL_TIMEOUT_USEC)
+    error_msg = '^Failed to open_session'
+    with pytest.raises(LibsshChannelException, match=error_msg):
```
webknjaz, line 52: @Jakuje FTR this has just failed for me locally on the latest `devel` (well, actually on your branch from #597 but it shouldn't matter):
```python-traceback
========================================================================= FAILURES =========================================================================
_____________________________________________________________ test_open_session_small_timeout ______________________________________________________________
[gw5] linux -- Python 3.12.12 ~/src/github/ansible/pylibssh/.tox/py/bin/python

ssh_session_connect = functools.partial(<function ensure_ssh_session_connected at 0x7f4da816a660>, sshd_addr=('127.0.0.1', 48057), ssh_clientkey_path=PosixPath('/tmp/pytest-of-<username>/pytest-0/popen-gw5/test_open_session_small_timeou0/sshd/ssh_client_ecdsa_key'))

    def test_open_session_small_timeout(ssh_session_connect):
        """Test opening a new channel with a small timeout value.
    
        This generates an exception from ``ssh_channel_open_session()``
        returning ``SSH_AGAIN`` with the ``usec`` timeout and default
        ``open_session_retries`` value of ``0``.
        """
        ssh_session = Session()
        ssh_session_connect(ssh_session)
        ssh_session.set_ssh_options('timeout_usec', SMALL_TIMEOUT_USEC)
        error_msg = '^Failed to open_session'
>       with pytest.raises(LibsshChannelException, match=error_msg):
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       Failed: DID NOT RAISE <class 'pylibsshext.errors.LibsshChannelException'>

error_msg  = '^Failed to open_session'
ssh_session = <pylibsshext.session.Session object at 0x7f4da120e4a0>
ssh_session_connect = functools.partial(<function ensure_ssh_session_connected at 0x7f4da816a660>, sshd_addr=('127.0.0.1', 48057), ssh_clientkey_path=PosixPath('/tmp/pytest-of-<username>/pytest-0/popen-gw5/test_open_session_small_timeou0/sshd/ssh_client_ecdsa_key'))

tests/unit/channel_test.py:50: Failed
```

webknjaz: Maybe we could stick https://pypi.org/p/pytest-rerunfailures and make it go away..

Jakuje: Anything that depends on timing is fragile. I do not think we can make it more reliable easily (unless we would include some artificial sleeps at places) so if rerun is going to solve it, I guess its probably best we can do.

webknjaz: I was kinda hoping there'd be some way to sequence pieces of automation to trigger this more reliably. But yeah, having reruns is probably a good first step towards reducing the CI flakiness.

PR #758, opened 2025-08-19
Fix ubi9 builds

Jakuje:
##### SUMMARY
ci: python3-pytest was moved to AppStream in CentOS 9

Built on top of #757, which fixes another breakage.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below
curl: (22) The requested URL returned error: 404
error: skipping https://rpmfind.net/linux/centos-stream/9-stream/CRB/x86_64/os/Packages/python3-pytest-6.2.2-4.el9.noarch.rpm - transfer failed
```

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1107,7 +1107,7 @@ jobs:
         --nodeps
         https://rpmfind.net/linux/centos-stream/"$(
           rpm --eval '%{rhel}'
-        )"-stream/CRB/x86_64/os/Packages/python3-pytest-6.2.2-4${{
+        )"-stream/AppStream/x86_64/os/Packages/python3-pytest-6.2.2-6${{
           steps.distribution-meta.outputs.dist-tag
         }}.noarch.rpm
         https://rpmfind.net/linux/centos-stream/"$(
--- a/docs/changelog-fragments/758.contrib.rst
+++ b/docs/changelog-fragments/758.contrib.rst
@@ -1,0 +1,2 @@
+Fixed link to python3-pytest for CentOS 9 Stream as it was recently moved from
+CRB to AppStream -- by :user:`Jakuje`.


webknjaz:
@Jakuje let's scope this to just UBI fixes, I'll merge the `setuptools-scm` bump separately.

Jakuje:
> @Jakuje let's scope this to just UBI fixes, I'll merge the `setuptools-scm` bump separately.

without the setuptools-scm update it explodes much easier than it gets to the ubi failures.

But even after fixing the pytest, there is something bringing in the `toml`, which I do not understand where it comes from and how to fix so not sure if it is something I can help much more from now on ...

webknjaz:
I force-merged #757 and clicking "rebase" in this one.

webknjaz:
> But even after fixing the pytest, there is something bringing in the `toml`, which I do not understand where it comes from and how to fix so not sure if it is something I can help much more from now on ...

Okay, I think I'll force-merge this one too, once the CI completes. And figure that out separately.

The exception is caused by https://github.com/pypa/build/blob/0.3.1.post1/src/build/__init__.py#L19. It's an ancient version of pypa/build that's installed manually: https://github.com/ansible/pylibssh/blob/6dcfbb6/packaging/rpm/ansible-pylibssh.spec#L42.

We should probably try looking into the possibility to run newer versions or check if the RPM repos already have it (this was a hack for when it wasn't available in RHEL).

My guess is that maybe the `toml` project was somehow available in the images in the past through some other dep we don't manage and not it's not. Newever versions of `build` don't use it: they rely on `tomllib` from stdlib on Python 3.11+ with a fallback to the API-compatible `tomli` project.

webknjaz:
UPD: simplifying the RPM spec works locally, I'll send a PR once I polish its overall shape.

webknjaz:
https://github.com/ansible/pylibssh/pull/759

PR #759, opened 2025-08-20
📦🧪 Simplify the RPM spec

webknjaz:
This patch removes the SRPM-vendored PyPI-sourced dependencies and migrates packaging to rely on `pyproject-rpm-macros` under RHEL.

Previously, this set of helpers was only in use on Fedora.

This move allows us to get rid of a lot of conditionals that were making people's lives difficult.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -996,9 +996,6 @@ jobs:
         - tag: fedora:40
         - tag: fedora:41
         - tag: fedora:42
-        # No matching package to install: 'python3dist(wheel)'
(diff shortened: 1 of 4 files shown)


Jakuje:
This is quite large overhaul, but looks like working good (except for some lints). Read through the change and did not notice anything off.

webknjaz:
@Jakuje yep, there were a lot of ancient hacks and it took some time testing what can be removed from the spec. It was late yesterday, so I got to finishing minor linting tweaks just today. I'll let it run  the CI once more before merging.

Issue #729, opened 2025-06-02
Integrate Codspeed to track benchmarks


webknjaz:
https://codspeed.io

webknjaz:
Somebody has to approve the https://github.com/apps/codspeed-hq app install into the org.

webknjaz:
Alright, gundalow approved it. So we'll be able to make use of it now. I've provisioned a secret into the repo. The next step would be to write some benchmarks with `pytest-codspeed` via https://codspeed.io/docs/benchmarks/python and integrate GHA: https://codspeed.io/docs/integrations/ci/github-actions.

webknjaz:
Blocked on https://github.com/CodSpeedHQ/runner/issues/135

PR #778, opened 2025-10-15
📦Only enable line tracing when w/Cython tracing

### LINKED ISSUE #767: [TODO] Ensure release builds don't include debug symbols
NOTE: sync https://github.com/aio-libs/yarl/pull/1521.

webknjaz:
This patch modifies the build backend to dynamically enable Cython line tracing only when explicitly requested via the `with-cython-tracing=true` config setting. Previously, having `linetrace = "True"` in `pyproject.toml` was making our PyPI-published wheels slower.

Now, line tracing is opt-in:
- Regular builds: `pip install .` (no line tracing)
- Tracing builds: `pip install . --config-setting=with-cython-tracing=true`

  (enables line tracing)

When tracing is requested, the build backend automatically adds the `linetrace=True` and `profile=True` Cython directives and sets the appropriate C compiler flags.

Resolves #767

##### ISSUE TYPE

- Packaging Pull Request

##### BLOCKERS

* [ ] **Revert f8121c5**
* [ ] #729

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -149,8 +149,7 @@ jobs:
           )
           && true || false
         }}
-      profiling-enabled: >-
-        ${{ steps.profiling-check.outputs.profiling-enabled || false }}
+      profiling-enabled: false
       cache-key-for-dep-files: >-
         ${{ steps.calc-cache-key-files.outputs.cache-key-for-dep-files }}
       git-tag: ${{ steps.git-tag.outputs.tag }}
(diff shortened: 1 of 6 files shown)


webknjaz:
With f8121c5 (**MUST BE REVERTED BEFORE MERGING**), it's visible that the artifact sizes differ. However, to measure the performance properly, we need to integrate #729 first. So I'm considering this PR blocked until then.

webknjaz:
Related: building limited-api/ABI3 is incompatible with line tracing — #819.

Issue #207, opened 2021-05-08
Publish a platform-specific wheel for Apple M1 (macOS ARM64) on PyPI


oliverl-21:
##### SUMMARY

add Package for macos arm64

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

N/A

webknjaz:
It's an interesting idea but until GitHub adds respective VMs to GHA, there's nothing we can do about this. So for now the users will have to install ansible-pylibssh from the sdist.

Refs:
* https://github.com/actions/virtual-environments/issues/2187
* https://github.com/actions/runner/issues/805

radokristof:
Well for me even building it locally does not work.
Before Ventura I was able to build it and install it locally. Now this fails with an error related to clang:

```
* Installing packages in isolated environment... (wheel)
* Building wheel...
[1/1] Cythonizing /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/pylibsshext/_libssh_version.pyx
running build_ext
building 'pylibsshext._libssh_version' extension
creating /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/tmpngpxapqx/private
creating /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/tmpngpxapqx/private/var
creating /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/tmpngpxapqx/private/var/folders
creating /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/tmpngpxapqx/private/var/folders/6l
creating /private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn/T/build-via-sdist-xbvujv80/ansible-pylibssh-1.0.1a1/src/tmpngpxapqx/private/var/folders/6l/48k83hdn30bc24655lbtzgx80000gn
[...]

PR #783, opened 2025-10-28
ci(macos): add workflow to build universal2 OpenSSL + libssh from source

Ruchip16:
##### SUMMARY

This PR introduces a new GitHub Actions workflow, `macOS-universal.yaml` — that builds and packages OpenSSL and libssh from source on macOS, producing universal2 artifacts (for both arm64 and x86_64 architectures).
The goal is to replace the current `brew install libssh` dependency in CI and enable fully reproducible macOS builds of libssh and its dependencies.

ci(macos):  It builds OpenSSL + libssh per arch, uploads artifacts, then merges with lipo into universal2.

Refs: #699 #207 #73
Issues: [ACA-3946](https://issues.redhat.com/browse/ACA-3946)

##### ISSUE TYPE

- Maintenance Pull Request
- Packaging Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1219,6 +1219,7 @@ jobs:
     name: 🧪 Test${{ '' }}  # nest jobs under the same sidebar category
     needs:
     - build-bin-macos
+    - macos-merge-universal
     - build-src
     - pre-setup  # transitive, for accessing settings
     strategy:
@@ -1251,6 +1252,9 @@ jobs:
         ${{ needs.pre-setup.outputs.sdist-artifact-name }}
       dists-artifact-name: >-
         ${{ needs.pre-setup.outputs.dists-artifact-name }}
+      macos-universal-artifact-name: macos-universal2-artifacts
+      openssl-lib-name: openssl
+      libssh-lib-name: libssh
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
 
--- a/.github/workflows/macos-universal.yml
+++ b/.github/workflows/macos-universal.yml
@@ -1,0 +1,207 @@
+name: macOS universal OpenSSL+libssh
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ devel ]
+  pull_request:
+
+env:
+  OPENSSL_VER: "3.1.4"
+  LIBSSH_VER: "0.11.2"
+
+jobs:
+  macos-arm64-build:
+    name: Build (arm64)
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Prep dirs
+        run: |
+          mkdir -p build arm64
+          echo "ROOT=$PWD" >> $GITHUB_ENV
+
+      - name: Build OpenSSL (arm64)
+        run: |
+          set -euxo pipefail
+          cd build
+          curl -LO "https://www.openssl.org/source/openssl-${OPENSSL_VER}.tar.gz"
+          tar xzf "openssl-${OPENSSL_VER}.tar.gz"
+          cd "openssl-${OPENSSL_VER}"
+          ./Configure darwin64-arm64-cc --prefix="$GITHUB_WORKSPACE/arm64/openssl" --libdir=lib \
+            no-tests no-ssl3 no-weak-ssl-ciphers enable-ec_nistp_64_gcc_128
+          make -j"$(sysctl -n hw.ncpu)"
+          make install_sw
+          file "$GITHUB_WORKSPACE/arm64/openssl/lib/libssl.dylib"
+
+      - name: Build libssh (arm64)
+        run: |
+          set -euxo pipefail
+          brew install cmake zlib || true
+          cd build
+          curl -LO "https://www.libssh.org/files/0.11/libssh-${LIBSSH_VER}.tar.xz"
+          tar xJf "libssh-${LIBSSH_VER}.tar.xz"
+          mkdir -p "libssh-${LIBSSH_VER}/build-arm64"
+          cd "libssh-${LIBSSH_VER}/build-arm64"
+          cmake .. \
+            -DCMAKE_BUILD_TYPE=MinSizeRel \
+            -DCMAKE_INSTALL_PREFIX="$GITHUB_WORKSPACE/arm64/libssh" \
+            -DCMAKE_OSX_ARCHITECTURES="arm64" \
+            -DBUILD_SHARED_LIBS=ON \
+            -DUNIT_TESTING=OFF -DCLIENT_TESTING=OFF -DSERVER_TESTING=OFF \
+            -DWITH_EXAMPLES=OFF -DWITH_GSSAPI=ON -DWITH_SERVER=OFF -DWITH_PCAP=OFF -DWITH_ZLIB=ON \
+            -DOPENSSL_ROOT_DIR="$GITHUB_WORKSPACE/arm64/openssl" \
+            -DOPENSSL_INCLUDE_DIR="$GITHUB_WORKSPACE/arm64/openssl/include" \
+            -DOPENSSL_CRYPTO_LIBRARY="$GITHUB_WORKSPACE/arm64/openssl/lib/libcrypto.dylib" \
+            -DOPENSSL_SSL_LIBRARY="$GITHUB_WORKSPACE/arm64/openssl/lib/libssl.dylib"
+          make -j"$(sysctl -n hw.ncpu)"
+          make install/strip
+          file "$GITHUB_WORKSPACE/arm64/libssh/lib/libssh.dylib"
+
+      - name: Package artifacts (arm64)
+        run: |
+          set -euxo pipefail
+          cd "$GITHUB_WORKSPACE/arm64"
+          tar czf "$GITHUB_WORKSPACE/openssl-${{ env.OPENSSL_VER }}-arm64.tgz" openssl
+          tar czf "$GITHUB_WORKSPACE/libssh-${{ env.LIBSSH_VER }}-arm64.tgz" libssh
+
+      - name: Upload artifacts (arm64)
+        uses: actions/upload-artifact@v4
+        with:
+          name: macos-arm64-artifacts
+          path: |
+            openssl-${{ env.OPENSSL_VER }}-arm64.tgz
+            libssh-${{ env.LIBSSH_VER }}-arm64.tgz
+          if-no-files-found: error
+
+  macos-x86_64-build:
+    name: Build (x86_64)
+    runs-on: macos-13
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Prep dirs
+        run: |
+          mkdir -p build x86_64
+
+      - name: Build OpenSSL (x86_64)
+        run: |
+          set -euxo pipefail
+          cd build
+          curl -LO "https://www.openssl.org/source/openssl-${OPENSSL_VER}.tar.gz"
+          tar xzf "openssl-${OPENSSL_VER}.tar.gz"
+          cd "openssl-${OPENSSL_VER}"
+          ./Configure darwin64-x86_64-cc --prefix="$GITHUB_WORKSPACE/x86_64/openssl" --libdir=lib \
+            no-tests no-ssl3 no-weak-ssl-ciphers enable-ec_nistp_64_gcc_128
+          make -j"$(sysctl -n hw.ncpu)"
+          make install_sw
+          file "$GITHUB_WORKSPACE/x86_64/openssl/lib/libssl.dylib"
+
+      - name: Build libssh (x86_64)
+        run: |
+          set -euxo pipefail
+          brew install cmake zlib || true
+          cd build
+          curl -LO "https://www.libssh.org/files/0.11/libssh-${LIBSSH_VER}.tar.xz"
+          tar xJf "libssh-${LIBSSH_VER}.tar.xz"
+          mkdir -p "libssh-${LIBSSH_VER}/build-x86_64"
+          cd "libssh-${LIBSSH_VER}/build-x86_64"
+          cmake .. \
+            -DCMAKE_BUILD_TYPE=MinSizeRel \
+            -DCMAKE_INSTALL_PREFIX="$GITHUB_WORKSPACE/x86_64/libssh" \
+            -DCMAKE_OSX_ARCHITECTURES="x86_64" \
+            -DBUILD_SHARED_LIBS=ON \
+            -DUNIT_TESTING=OFF -DCLIENT_TESTING=OFF -DSERVER_TESTING=OFF \
+            -DWITH_EXAMPLES=OFF -DWITH_GSSAPI=ON -DWITH_SERVER=OFF -DWITH_PCAP=OFF -DWITH_ZLIB=ON \
+            -DOPENSSL_ROOT_DIR="$GITHUB_WORKSPACE/x86_64/openssl" \
+            -DOPENSSL_INCLUDE_DIR="$GITHUB_WORKSPACE/x86_64/openssl/include" \
+            -DOPENSSL_CRYPTO_LIBRARY="$GITHUB_WORKSPACE/x86_64/openssl/lib/libcrypto.dylib" \
+            -DOPENSSL_SSL_LIBRARY="$GITHUB_WORKSPACE/x86_64/openssl/lib/libssl.dylib"
+          make -j"$(sysctl -n hw.ncpu)"
+          make install/strip
+          file "$GITHUB_WORKSPACE/x86_64/libssh/lib/libssh.dylib"
+
+      - name: Package artifacts (x86_64)
+        run: |
+          set -euxo pipefail
+          cd "$GITHUB_WORKSPACE/x86_64"
+          tar czf "$GITHUB_WORKSPACE/openssl-${{ env.OPENSSL_VER }}-x86_64.tgz" openssl
+          tar czf "$GITHUB_WORKSPACE/libssh-${{ env.LIBSSH_VER }}-x86_64.tgz" libssh
+
+      - name: Upload artifacts (x86_64)
+        uses: actions/upload-artifact@v4
+        with:
+          name: macos-x86_64-artifacts
+          path: |
+            openssl-${{ env.OPENSSL_VER }}-x86_64.tgz
+            libssh-${{ env.LIBSSH_VER }}-x86_64.tgz
+          if-no-files-found: error
+
+  macos-merge-universal:
+    name: Merge → universal2
+    runs-on: macos-14
+    needs: [macos-arm64-build, macos-x86_64-build]
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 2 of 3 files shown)


KB-perByte:
@Ruchip16 did we agree on a new workflow for macos?
I am not sure!

Ruchip16:
> @Ruchip16 did we agree on a new workflow for macos? I am not sure!

The intent was about the build logic, not the workflow boundary. 
We can incorporte the 3 builds into existing workflow as well, since this is a WIP, for now just to keep it clean, I created a new workflow but we can embed this into existing ones

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+  pull_request:
+
+env:
+  OPENSSL_VER: "3.1.4"
+  LIBSSH_VER: "0.11.2"
+
+jobs:
+  macos-arm64-build:
+    name: Build (arm64)
+    runs-on: macos-14
```
webknjaz: Can we use

```suggestion
    runs-on: macos-15
```

already? We might need to set an older API target through an env var, though.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          name: macos-arm64-artifacts
+          path: |
+            openssl-${{ env.OPENSSL_VER }}-arm64.tgz
+            libssh-${{ env.LIBSSH_VER }}-arm64.tgz
+          if-no-files-found: error
+
+  macos-x86_64-build:
+    name: Build (x86_64)
+    runs-on: macos-13
```
webknjaz: GH ended up making intel available recently
```suggestion
    runs-on: macos-15-intel
```

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+      - name: Upload artifacts (arm64)
+        uses: actions/upload-artifact@v4
+        with:
+          name: macos-arm64-artifacts
+          path: |
+            openssl-${{ env.OPENSSL_VER }}-arm64.tgz
+            libssh-${{ env.LIBSSH_VER }}-arm64.tgz
+          if-no-files-found: error
+
+  macos-x86_64-build:
```
webknjaz, line 103: PyCA seems to build both under the same VM variant, just controlling the target via build settings. I wonder if we can do that and whether it'll mean any performance penalty..

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+          mkdir -p universal/openssl/lib universal/openssl/include universal/libssh/lib universal/libssh/include
+
+          cp -R merge/openssl/include/* universal/openssl/include/
+          cp -R merge/libssh/include/*   universal/libssh/include/
+
+          lipo -create merge/openssl/lib/libssl.dylib   merge/x86_64/openssl/lib/libssl.dylib   -output universal/openssl/lib/libssl.dylib
+          lipo -create merge/openssl/lib/libcrypto.dylib merge/x86_64/openssl/lib/libcrypto.dylib -output universal/openssl/lib/libcrypto.dylib
+          lipo -create merge/libssh/lib/libssh.dylib     merge/x86_64/libssh/lib/libssh.dylib     -output universal/libssh/lib/libssh.dylib
+
+          # optional static libs if built:
```
webknjaz: why optional?

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+          make -j"$(sysctl -n hw.ncpu)"
+          make install/strip
+          file "$GITHUB_WORKSPACE/x86_64/libssh/lib/libssh.dylib"
+
+      - name: Package artifacts (x86_64)
+        run: |
+          set -euxo pipefail
+          cd "$GITHUB_WORKSPACE/x86_64"
+          tar czf "$GITHUB_WORKSPACE/openssl-${{ env.OPENSSL_VER }}-x86_64.tgz" openssl
+          tar czf "$GITHUB_WORKSPACE/libssh-${{ env.LIBSSH_VER }}-x86_64.tgz" libssh
```
webknjaz: these probably don't need to be double-achrived

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+            -DCMAKE_OSX_ARCHITECTURES="arm64" \
+            -DBUILD_SHARED_LIBS=ON \
+            -DUNIT_TESTING=OFF -DCLIENT_TESTING=OFF -DSERVER_TESTING=OFF \
+            -DWITH_EXAMPLES=OFF -DWITH_GSSAPI=ON -DWITH_SERVER=OFF -DWITH_PCAP=OFF -DWITH_ZLIB=ON \
+            -DOPENSSL_ROOT_DIR="$GITHUB_WORKSPACE/arm64/openssl" \
+            -DOPENSSL_INCLUDE_DIR="$GITHUB_WORKSPACE/arm64/openssl/include" \
+            -DOPENSSL_CRYPTO_LIBRARY="$GITHUB_WORKSPACE/arm64/openssl/lib/libcrypto.dylib" \
+            -DOPENSSL_SSL_LIBRARY="$GITHUB_WORKSPACE/arm64/openssl/lib/libssl.dylib"
+          make -j"$(sysctl -n hw.ncpu)"
+          make install/strip
```
webknjaz: this probably shouldn't be run

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+        run: |
+          set -euxo pipefail
+          cd build
+          curl -LO "https://www.openssl.org/source/openssl-${OPENSSL_VER}.tar.gz"
+          tar xzf "openssl-${OPENSSL_VER}.tar.gz"
+          cd "openssl-${OPENSSL_VER}"
+          ./Configure darwin64-arm64-cc --prefix="$GITHUB_WORKSPACE/arm64/openssl" --libdir=lib \
+            no-tests no-ssl3 no-weak-ssl-ciphers enable-ec_nistp_64_gcc_128
+          make -j"$(sysctl -n hw.ncpu)"
+          make install_sw
```
webknjaz: why is this needed?

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+      - name: Build OpenSSL (arm64)
+        run: |
+          set -euxo pipefail
+          cd build
+          curl -LO "https://www.openssl.org/source/openssl-${OPENSSL_VER}.tar.gz"
+          tar xzf "openssl-${OPENSSL_VER}.tar.gz"
+          cd "openssl-${OPENSSL_VER}"
+          ./Configure darwin64-arm64-cc --prefix="$GITHUB_WORKSPACE/arm64/openssl" --libdir=lib \
+            no-tests no-ssl3 no-weak-ssl-ciphers enable-ec_nistp_64_gcc_128
+          make -j"$(sysctl -n hw.ncpu)"
```
webknjaz: It's best to split separate invocations across separate steps in the job definition. Otherwise, it'd difficult to scan the logs and sometimes outputs of multiple things can be confused with each other.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+  macos-arm64-build:
+    name: Build (arm64)
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Prep dirs
+        run: |
+          mkdir -p build arm64
+          echo "ROOT=$PWD" >> $GITHUB_ENV
```
webknjaz: this is unused

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+env:
+  OPENSSL_VER: "3.1.4"
+  LIBSSH_VER: "0.11.2"
+
+jobs:
+  macos-arm64-build:
+    name: Build (arm64)
+    runs-on: macos-14
+    steps:
+      - uses: actions/checkout@v4
```
webknjaz: Always name steps.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+
+      - name: Prep dirs
```
webknjaz: Use full words when explaining what steps do. Best avoid generic references to abstract ideas. Be specific.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+        run: |
+          mkdir -p build arm64
```
webknjaz: A bit of verbosity wouldn't hurt.
```suggestion
          mkdir -pv build arm64
```

Although, is this needed?

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+          echo "ROOT=$PWD" >> $GITHUB_ENV
+
+      - name: Build OpenSSL (arm64)
+        run: |
+          set -euxo pipefail
+          cd build
```
webknjaz: this can be set via `working-directory:`

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+      - name: Unpack
+        run: |
+          mkdir -p merge
+          tar xzf arm64/openssl-*-arm64.tgz -C merge
+          tar xzf arm64/libssh-*-arm64.tgz   -C merge
+          tar xzf x86_64/openssl-*-x86_64.tgz -C merge
+          tar xzf x86_64/libssh-*-x86_64.tgz  -C merge
+          ls -R merge
+
+      - name: Lipo → universal2
```
webknjaz: Something explanatory would be useful here. The context of the references is rather unclear.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          name: macos-arm64-artifacts
+          path: arm64
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: macos-x86_64-artifacts
+          path: x86_64
+
+      - name: Unpack
```
webknjaz: what exactly?

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: macos-arm64-artifacts
+          path: arm64
+
+      - uses: actions/download-artifact@v4
```
webknjaz: name?

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+    runs-on: macos-14
+    needs: [macos-arm64-build, macos-x86_64-build]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Download artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: macos-arm64-artifacts
+          path: arm64
```
webknjaz: this could probably just use pattern+merge instead of name

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          name: macos-x86_64-artifacts
+          path: |
+            openssl-${{ env.OPENSSL_VER }}-x86_64.tgz
+            libssh-${{ env.LIBSSH_VER }}-x86_64.tgz
+          if-no-files-found: error
+
+  macos-merge-universal:
+    name: Merge → universal2
+    runs-on: macos-14
```
webknjaz:
```suggestion
    runs-on: macos-15
```

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
...
+          fi
+
+          file universal/openssl/lib/libssl.dylib
+          lipo -info universal/openssl/lib/libssl.dylib
+          otool -L universal/libssh/lib/libssh.dylib
+
+      - name: Package universal artifacts
+        run: |
+          tar czf openssl-${{ env.OPENSSL_VER }}-universal2.tgz -C universal openssl
+          tar czf libssh-${{ env.LIBSSH_VER }}-universal2.tgz   -C universal libssh
```
webknjaz: having non-verbose invocations in the same step will result in a log smashed together

.github/workflows/macos-universal.yml
webknjaz, line 1: We'll probably need to integrate this back into the same workflow that builds the container images and reuse some vers. Though, we'll talk about it later.

Seeing the amount of copy-paste, I'm pretty sure the build job definitions can be squashed into one too.

We'll also need to see if we can/need to move some portions into scripts.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
+name: macOS universal OpenSSL+libssh
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [ devel ]
+  pull_request:
```
webknjaz, line 7: This will have to be integrated with the main artifact generation workflow. Or moved in. Something to think about.

.github/workflows/macos-universal.yml
```diff
@@ -0,0 +1,207 @@
+name: macOS universal OpenSSL+libssh
```
webknjaz:
```suggestion
name: Multi-arch macOS OpenSSL and libssh artifacts
```

webknjaz:
@Ruchip16 in case you haven't tried it, running `tox r -e lint -qq` locally will allow you to run the linters w/o waiting for the CI.

webknjaz:
See also: https://ansible-pylibssh.rtfd.io/en/latest/contributing/guidelines/

.github/workflows/macos-universal.yml
Ruchip16: yes

.github/workflows/ci-cd.yml
```diff
@@ -1158,6 +1158,7 @@ jobs:
     name: 🧪 Test${{ '' }}  # nest jobs under the same sidebar category
     needs:
     - build-bin-macos
+    - macos-merge-universal
```
webknjaz, line 1161: This breaks the CI. We can't and shouldn't depend on things that aren't in the same workflow run. It's a separate process that runs independently on schedule.
```suggestion
```

.github/workflows/reusable-tests.yml
```diff
@@ -39,6 +39,20 @@ on:  # yamllint disable-line rule:truthy
...
+      openssl-lib-name:
+        description: Directory name for OpenSSL inside the artifact
+        required: false
+        type: string
+        default: openssl
+      libssh-lib-name:
+        description: Directory name for libssh inside the artifact
+        required: false
+        type: string
+        default: libssh
```
webknjaz, line 55: We could probably just hardcode these for now. The artifacts aren't coming from the same workflow run anyway.

.github/workflows/reusable-tests.yml
webknjaz, line 1: I don't know if we should be changing the testing workflow, though. These artifacts are supposed to be used under `cibuildwheel`. That's the main target, at least.

.github/workflows/reusable-tests.yml
```diff
@@ -121,10 +135,30 @@ jobs:
...
+        tar xzf .deps/libssh-*-universal2.tgz  -C .deps/u
+        OPENSSL_DIR="$(pwd)/.deps/u/${{ inputs.openssl-lib-name }}"
+        LIBSSH_DIR="$(pwd)/.deps/u/${{ inputs.libssh-lib-name }}"
+        {
+          echo "OPENSSL_DIR=$OPENSSL_DIR"
+          echo "LIBSSH_DIR=$LIBSSH_DIR"
+          echo "PKG_CONFIG_PATH=$OPENSSL_DIR/lib/pkgconfig:$LIBSSH_DIR/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+          echo "LDFLAGS=-L$OPENSSL_DIR/lib -L$LIBSSH_DIR/lib ${LDFLAGS:-}"
+          echo "CPPFLAGS=-I$OPENSSL_DIR/include -I$LIBSSH_DIR/include ${CPPFLAGS:-}"
+        } >> "$GITHUB_ENV"
```
webknjaz, line 161: It should be possible to set these vars directly in the `env:` option of the `Pre-populate tox env` step down below. It's best to move it here.

.github/workflows/reusable-tests.yml
```diff
@@ -121,10 +135,30 @@ jobs:
...
-        runner.os == 'macOS'
-      run: brew install libssh
+        runner.os == 'macOS' && inputs.macos-universal-artifact-name != ''
+      uses: actions/download-artifact@v4
+      with:
+        name: ${{ inputs.macos-universal-artifact-name }}
+        path: .deps/
+    - name: (macOS) Unpack deps and export build env
+      if: >-
+        runner.os == 'macOS' && inputs.macos-universal-artifact-name != ''
```
webknjaz, line 147: This conditional should probably be
```suggestion
        runner.os == 'macOS' && inputs.dist-type == 'source'
```
instead.

.github/workflows/reusable-tests.yml
```diff
@@ -121,10 +135,30 @@ jobs:
         inputs.dist-type == 'source' &&
         runner.os == 'Linux'
       run: sudo apt update && sudo apt install build-essential libssl-dev
-    - name: Install libssh and openssl headers on macOS
+    - name: (macOS) Download universal2 OpenSSL+libssh
       if: >-
-        runner.os == 'macOS'
-      run: brew install libssh
+        runner.os == 'macOS' && inputs.macos-universal-artifact-name != ''
```
webknjaz, line 140:
```suggestion
        runner.os == 'macOS' && inputs.dist-type == 'source'
```

.github/workflows/reusable-tests.yml
```diff
@@ -121,10 +135,30 @@ jobs:
...
+      uses: actions/download-artifact@v4
```
webknjaz, line 141: This action can only download artifacts from within the same workflow run. To get things from other workflows (and workflow runs), we need something beyond the standard actions.

For example, PyCA uses this: https://github.com/pyca/cryptography/blob/84d85b4b2a21418a3110a06078f281d30e18a2bd/.github/workflows/ci.yml#L317-L325 — in their case, it gets artifacts from a companion repo, not the same. This is something to reproduce here and in the `cibuildwheel` workflow.

webknjaz (commented):
I'd like to note that it's best to only implement producing the artifacts within this PR. And not downloading/integrating them.

This is because the new workflow will have to actually run and create those artifacts for consumption, which is only possible post-merge.

So we'll merge a PR emitting said artifacts, run the workflow and only then will we start integrating them. This may end up requiring us scripting something around GH CLI and adding it into the `cibuildwheel` config for macos only. But that's a separate part of the effort. We won't be able to start implementing it until the previous part is merged.

So my recommendation is to revert things other than producing the artifacts and polish that+merge. Just do note the feedback on the reverted bits and take it into account in future PRs.

webknjaz:
@Ruchip16 this PR will likely need a change note of type `contrib`.

PR #790, opened 2026-01-28
Config file processing - takes in config_file from custom path

KB-perByte:
##### SUMMARY

This patch adds support for loading and applying settings from configuration files to `libssh` connection sessions.

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

https://github.com/ansible-collections/ansible.netcommon/issues/430

```console
#env:
$ python -c "from pylibsshext import __full_version__; print(__full_version__)"
<pylibsshext v0.1.dev2067+g570153995.d20260213 with libssh v0.12.0>
```

```sshconfig
# ../../custom_ssh_file
Host cisco-router-direct-demo
    HostName 54.1c.cx.yx
    User cisco
    Port 0000
```

```ini
# ./inventory
[ios]
test

[ios:vars]
ansible_host=cisco-router-direct-demo
ansible_network_os=cisco.ios.ios
ansible_ssh_password=TOPSECRETPASSWORD
ansible_connection=ansible.netcommon.network_cli
ansible_network_cli_ssh_type=libssh
ansible_libssh_config_file=../../custom_ssh_file
```

```yaml
# ./Playbook.yaml
---
- hosts: ios
  gather_facts: false
  tasks:
    - name: Just connect and gather
      cisco.ios.ios_hostname:
        config:
        state: gathered
```

log to prove the change
```console
2026-02-13 12:28:01,946 p=95788 u=batmon n=ansible INFO| <cisco-router-direct-demo> local domain socket path is /home/batmon/.ansible/pc/83b7a917df
2026-02-13 12:28:01,946 p=95788 u=batmon n=ansible INFO| <cisco-router-direct-demo> ANSIBLE_NETWORK_IMPORT_MODULES: enabled
2026-02-13 12:28:02,188 p=95788 u=batmon n=ansible INFO| <cisco-router-direct-demo> ANSIBLE_NETWORK_IMPORT_MODULES: found cisco.ios.ios_hostname at /home/batmon/Work/AnsibleNetwork/collections/ansible_collections/cisco/ios/plugins/modules/ios_hostname.py
2026-02-13 12:28:02,188 p=95788 u=batmon n=ansible INFO| <cisco-router-direct-demo> ANSIBLE_NETWORK_IMPORT_MODULES: running cisco.ios.ios_hostname
2026-02-13 12:28:02,188 p=95788 u=batmon n=ansible INFO| <cisco-router-direct-demo> ANSIBLE_NETWORK_IMPORT_MODULES: _load_params skipped for action plugin in direct execution
2026-02-13 12:28:02,189 p=95847 u=batmon n=ansible INFO| jsonrpc request: b'{"jsonrpc": "2.0", "method": "get_capabilities", "id": "daa57e56-b8f8-4fd5-9c0c-a5e7e69c0cc9", "params": [[], {}]}'
2026-02-13 12:28:02,189 p=95847 u=batmon n=ansible INFO| <cisco-router-direct-demo> USING PYLIBSSH VERSION 0.1.dev2067+g570153995.d20260213
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible INFO| <cisco-router-direct-demo> ESTABLISH LIBSSH CONNECTION FOR USER: None TO cisco-router-direct-demo (port from SSH config)
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_config_parse_file: Reading configuration data from /redacted/..//ssh_config
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_connect: libssh 0.11.3 (c) 2003-2025 Aris Adamantiadis, Andreas Schneider and libssh contributors. Distributed under the LGPL, please refer to COPYING file for information about your rights, using threading threads_pthread
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| getai: host 54.1c.cx.yx matches an IP address
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_socket_connect: Nonblocking connection socket: 7
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_connect: Socket connecting, now waiting for the callbacks to work
2026-02-13 12:28:02,190 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_connect: Actual timeout : 2000000
2026-02-13 12:28:02,485 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_socket_pollcallback: Received POLLOUT in connecting state
2026-02-13 12:28:02,485 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_client_connection_callback: session_state=2
2026-02-13 12:28:02,485 p=95847 u=batmon n=ansible-pylibssh DEBUG| ssh_socket_unbuffered_write: Enabling POLLOUT for socket
2026-02-13 12:28:02,868 p=95847 u=batmon n=ansible-pylibssh DEBUG| callback_receive_banner: Received banner: SSH-2.0-Cisco-1.25
```

### DIFF
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -1,0 +1,28 @@
+"""Connect using a custom SSH config file.
+
+Use this when you have a dedicated config file (e.g. for Host aliases,
+ProxyCommand, or per-host options) and want the session to use it
+instead of the default ~/.ssh/config.
+"""
+from pylibsshext.errors import LibsshSessionException
+from pylibsshext.session import Session
+
+ssh = Session()
+
+# Host alias or hostname that appears in your custom config
+HOST = "myalias"
+USER = "myuser"
+CONFIG_FILE = "/path/to/my/ssh_config"
+
+try:
+    ssh.connect(
+        host=HOST,
+        user=USER,
+        config_file=CONFIG_FILE,
+    )
+except LibsshSessionException as exc:
+    print(f"Failed to connect: {exc!s}")
+
+if ssh.is_connected:
+    print("Connected.")
+    ssh.close()
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -10,7 +10,10 @@ specific config file. When using ``parse_config()``, set ``host`` first
 so that ``Host`` blocks in the config match correctly.
 """
 
-from pylibsshext.errors import LibsshConfigParseException, LibsshSessionException
+from pylibsshext.errors import (
+    LibsshConfigParseException,
+    LibsshSessionException,
+)
 from pylibsshext.session import Session
 
 
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -36,8 +36,9 @@ except LibsshSessionException as ssh_session_connection_error:
     print(f'Failed to connect: {ssh_session_connection_error!s}')
 
 # Alternative: call parse_config() explicitly (e.g. with a pathlib.Path)
-# ssh = Session(host=HOST)
-# ssh.parse_config(Path(CONFIG_FILE))  # or str/bytes
+# ssh = Session()
+# ssh.set_ssh_options('host', HOST)
+# ssh.parse_config(Path(CONFIG_FILE))  # or str
 # ssh.connect(user=USER)
 
 if ssh.is_connected:
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -30,10 +30,10 @@ try:
         user=USER,
         config_file=CONFIG_FILE,
     )
-except LibsshConfigParseException as exc:
-    print(f'Failed to parse SSH config: {exc!s}')
-except LibsshSessionException as exc:
-    print(f'Failed to connect: {exc!s}')
+except LibsshConfigParseException as ssh_config_load_error:
+    print(f'Failed to load the SSH config: {ssh_config_load_error!s}')
+except LibsshSessionException as ssh_session_connection_error:
+    print(f'Failed to connect: {ssh_session_connection_error!s}')
 
 if ssh.is_connected:
     print('Connected.')
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -10,8 +10,6 @@ specific config file. When using ``parse_config()``, set ``host`` first
 so that ``Host`` blocks in the config match correctly.
 """
 
-from pathlib import Path
-
 from pylibsshext.errors import (
     LibsshConfigParseException,
     LibsshSessionException,
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -2,7 +2,7 @@
 
 Use this when you have a dedicated config file (e.g. for Host aliases,
 ProxyCommand, or per-host options) and want the session to use it
-instead of the default ~/.ssh/config.
+instead of the default ~/.ssh/config and /etc/ssh/ssh_config.
 """
 
 from pylibsshext.errors import LibsshSessionException
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -10,6 +10,8 @@ specific config file. When using ``parse_config()``, set ``host`` first
 so that ``Host`` blocks in the config match correctly.
 """
 
+from pathlib import Path
+
 from pylibsshext.errors import (
     LibsshConfigParseException,
     LibsshSessionException,
@@ -34,6 +36,11 @@ except LibsshConfigParseException as ssh_config_load_error:
     print(f'Failed to load the SSH config: {ssh_config_load_error!s}')
 except LibsshSessionException as ssh_session_connection_error:
     print(f'Failed to connect: {ssh_session_connection_error!s}')
+
+# Alternative: call parse_config() explicitly (e.g. with a pathlib.Path)
+# ssh = Session(host=HOST)
+# ssh.parse_config(Path(CONFIG_FILE))  # or str/bytes
+# ssh.connect(user=USER)
 
 if ssh.is_connected:
     print('Connected.')
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -1,11 +1,16 @@
 """Connect using a custom SSH config file.
 
-Use this when you have a dedicated config file (e.g. for Host aliases,
-ProxyCommand, or per-host options) and want the session to use it
+Use this when you have a dedicated config file (e.g. for ``Host`` aliases,
+``ProxyCommand``, or per-host options) and want the session to use it
 instead of the default ~/.ssh/config and /etc/ssh/ssh_config.
+
+You can either pass ``config_file`` to ``Session.connect()`` (recommended),
+or call ``Session.parse_config(path)`` before ``connect()`` to load a
+specific config file. When using ``parse_config()``, set ``host`` first
+so that ``Host`` blocks in the config match correctly.
 """
 
-from pylibsshext.errors import LibsshSessionException
+from pylibsshext.errors import LibsshConfigParseException, LibsshSessionException
 from pylibsshext.session import Session
 
 
@@ -22,6 +27,8 @@ try:
         user=USER,
         config_file=CONFIG_FILE,
     )
+except LibsshConfigParseException as exc:
+    print(f'Failed to parse SSH config: {exc!s}')
 except LibsshSessionException as exc:
     print(f'Failed to connect: {exc!s}')
 
--- a/docs/_samples/ssh_config_connect.py
+++ b/docs/_samples/ssh_config_connect.py
@@ -4,15 +4,17 @@ Use this when you have a dedicated config file (e.g. for Host aliases,
 ProxyCommand, or per-host options) and want the session to use it
 instead of the default ~/.ssh/config.
 """
+
 from pylibsshext.errors import LibsshSessionException
 from pylibsshext.session import Session
+
 
 ssh = Session()
 
 # Host alias or hostname that appears in your custom config
-HOST = "myalias"
-USER = "myuser"
-CONFIG_FILE = "/path/to/my/ssh_config"
+HOST = 'myalias'
+USER = 'myuser'
+CONFIG_FILE = '/path/to/my/ssh_config'
 
 try:
     ssh.connect(
@@ -21,8 +23,8 @@ try:
         config_file=CONFIG_FILE,
     )
 except LibsshSessionException as exc:
-    print(f"Failed to connect: {exc!s}")
+    print(f'Failed to connect: {exc!s}')
 
 if ssh.is_connected:
-    print("Connected.")
+    print('Connected.')
     ssh.close()
--- a/docs/changelog-fragments/790.bugfix.rst
+++ b/docs/changelog-fragments/790.bugfix.rst
@@ -1,0 +1,8 @@
+Fixed the ``config_file`` parameter being ignored in
+``Session.connect()``. When a custom SSH config file path was passed
+via the ``config_file`` keyword argument, it was not being processed
+because it was missing from the options mapping. Now the config file
+is properly parsed before establishing the connection, enabling SSH
+config features like host aliases and custom settings to work correctly.
+
+-- by :user:`KB-perByte`
--- a/docs/changelog-fragments/790.bugfix.rst
+++ b/docs/changelog-fragments/790.bugfix.rst
@@ -1,8 +1,0 @@
-Fixed the ``config_file`` parameter being ignored in
-``Session.connect()``. When a custom SSH config file path was passed
-via the ``config_file`` keyword argument, it was not being processed
-because it was missing from the options mapping. Now the config file
-is properly parsed before establishing the connection, enabling SSH
-config features like host aliases and custom settings to work correctly.
-
--- by :user:`KB-perByte`
--- a/docs/changelog-fragments/790.feature.rst
+++ b/docs/changelog-fragments/790.feature.rst
@@ -1,6 +1,6 @@
 Added support for using a custom SSH config file when connecting. You can now pass
-``config_file`` to :meth:`Session.connect() <pylibsshext.session.Session.connect>` to
-parse a specific config file (e.g. for Host aliases, ``ProxyCommand``, or per-host
-settings) before the connection is established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
+``config_file`` to ``Session.connect()`` to parse a specific config file (e.g. for
+Host aliases, ``ProxyCommand``, or per-host settings) before the connection is
+established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
 
 -- by :user:`KB-perByte`
--- a/docs/changelog-fragments/790.feature.rst
+++ b/docs/changelog-fragments/790.feature.rst
@@ -3,4 +3,4 @@ Added support for using a custom SSH config file when connecting. You can now pa
 ``Host`` aliases, ``ProxyCommand``, or per-host settings) before the connection is
 established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
 
--- by :user:`KB-perByte`
+-- by :user:`donnerhacke` and :user:`KB-perByte`
--- a/docs/changelog-fragments/790.feature.rst
+++ b/docs/changelog-fragments/790.feature.rst
@@ -1,0 +1,6 @@
+Added support for using a custom SSH config file when connecting. You can now pass
+``config_file`` to :meth:`Session.connect() <pylibsshext.session.Session.connect>` to
+parse a specific config file (e.g. for Host aliases, ``ProxyCommand``, or per-host
+settings) before the connection is established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
+
+-- by :user:`KB-perByte`
--- a/docs/changelog-fragments/790.feature.rst
+++ b/docs/changelog-fragments/790.feature.rst
@@ -1,6 +1,6 @@
 Added support for using a custom SSH config file when connecting. You can now pass
 ``config_file`` to ``Session.connect()`` to parse a specific config file (e.g. for
-Host aliases, ``ProxyCommand``, or per-host settings) before the connection is
+``Host`` aliases, ``ProxyCommand``, or per-host settings) before the connection is
 established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
 
 -- by :user:`KB-perByte`
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -49,6 +49,23 @@ Connecting with remote SSH server
    :emphasize-lines: 7-13
 
 
+.. _custom-ssh-config:
+
+Connecting with a custom SSH config
+-----------------------------------
+
+To use a custom SSH config file (e.g. for Host aliases, ``ProxyCommand``, or
+per-host options), pass ``config_file`` to :meth:`~pylibsshext.session.Session.connect`.
+The host option should be set so that Host matching in the config works correctly.
+
+.. literalinclude:: _samples/ssh_config_connect.py
+   :language: python
+   :start-at: ssh = Session()
+   :end-before: if ssh.is_connected:
+   :dedent: 4
+   :emphasize-lines: 5-10
+
+
 Connecting over GSSAPI
 ----------------------
 
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -54,15 +54,17 @@ Connecting with remote SSH server
 Connecting with a custom SSH config
 -----------------------------------
 
-To use a custom SSH config file (e.g. for Host aliases, ``ProxyCommand``, or
-per-host options), pass ``config_file`` to :meth:`~pylibsshext.session.Session.connect`.
-The host option should be set so that Host matching in the config works correctly.
+By default, libssh loads configuration from ``~/.ssh/config`` and
+``/etc/ssh/ssh_config``. To use a custom SSH config file instead (e.g. for Host
+aliases, ``ProxyCommand``, or per-host options), pass ``config_file`` to
+``Session.connect()``. When you pass ``config_file``, libssh will not consult
+the default files above. The host option should be set so that Host matching in
+the config works correctly.
 
 .. literalinclude:: _samples/ssh_config_connect.py
    :language: python
    :start-at: ssh = Session()
    :end-before: if ssh.is_connected:
-   :dedent: 4
    :emphasize-lines: 5-10
 
 
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -65,7 +65,7 @@ consult the default files above. The ``host`` option should be set so that
    :language: python
    :start-at: ssh = Session()
    :end-before: if ssh.is_connected:
-   :emphasize-lines: 5-11
+   :emphasize-lines: 6,9-12
 
 
 Connecting over GSSAPI
--- a/docs/user_guide.rst
+++ b/docs/user_guide.rst
@@ -54,18 +54,18 @@ Connecting with remote SSH server
 Connecting with a custom SSH config
 -----------------------------------
 
-By default, libssh loads configuration from ``~/.ssh/config`` and
-``/etc/ssh/ssh_config``. To use a custom SSH config file instead (e.g. for Host
-aliases, ``ProxyCommand``, or per-host options), pass ``config_file`` to
-``Session.connect()``. When you pass ``config_file``, libssh will not consult
-the default files above. The host option should be set so that Host matching in
-the config works correctly.
+By default, libssh loads configuration from :file:`~/.ssh/config` and
+:file:`/etc/ssh/ssh_config`. To use a custom SSH config file instead (e.g. for
+``Host`` aliases, ``ProxyCommand``, or per-host options), pass ``config_file``
+to ``Session.connect()``. When you pass ``config_file``, libssh will not
+consult the default files above. The ``host`` option should be set so that
+``Host`` matching in the config works correctly.
 
 .. literalinclude:: _samples/ssh_config_connect.py
    :language: python
    :start-at: ssh = Session()
    :end-before: if ssh.is_connected:
-   :emphasize-lines: 5-10
+   :emphasize-lines: 5-11
 
 
 Connecting over GSSAPI
--- a/packaging/rpm/ansible-pylibssh.spec
+++ b/packaging/rpm/ansible-pylibssh.spec
@@ -1,11 +1,11 @@
 %global canonical_project_name %(grep --after-context=1 '^\\[metadata\\]$' setup.cfg | tail --lines=1 | sed --regexp-extended 's:^name = ([^ ]*):\\1:g')
 %global pypi_name %canonical_project_name
-%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:[.]:_:g')
+%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:\\.:_:g')
 %global srcname %normalized_core_packaging_metadata_project_name
 
 # NOTE: The target version may be set dynamically via
 # NOTE: rpmbuild --define "upstream_version 0.2.1.dev125+g0b5bde0"
-%global upstream_version_fallback %(ls -t dist/%{normalized_core_packaging_metadata_project_name}-*.tar.gz 2>/dev/null | head -n 1 | sed -E 's#^dist/%{normalized_core_packaging_metadata_project_name}-(.*)[.]tar[.]gz$#\1#')
+%global upstream_version_fallback %(ls -t dist/%{normalized_core_packaging_metadata_project_name}-*.tar.gz 2>/dev/null | head -n 1 | sed 's#^dist\\/%{normalized_core_packaging_metadata_project_name}-\\(.*\\)\\.tar\\.gz$#\\1#')
 # If "upstream_version" macro is unset, use the fallback defined above:
 %if "%{!?upstream_version:UNSET}" == "UNSET"
 %global upstream_version %{upstream_version_fallback}
--- a/packaging/rpm/ansible-pylibssh.spec
+++ b/packaging/rpm/ansible-pylibssh.spec
@@ -1,11 +1,11 @@
 %global canonical_project_name %(grep --after-context=1 '^\\[metadata\\]$' setup.cfg | tail --lines=1 | sed --regexp-extended 's:^name = ([^ ]*):\\1:g')
 %global pypi_name %canonical_project_name
-%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:\\.:_:g')
+%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:[.]:_:g')
 %global srcname %normalized_core_packaging_metadata_project_name
 
 # NOTE: The target version may be set dynamically via
 # NOTE: rpmbuild --define "upstream_version 0.2.1.dev125+g0b5bde0"
-%global upstream_version_fallback %(ls -t dist/%{normalized_core_packaging_metadata_project_name}-*.tar.gz 2>/dev/null | head -n 1 | sed 's#^dist\\/%{normalized_core_packaging_metadata_project_name}-\\(.*\\)\\.tar\\.gz$#\\1#')
+%global upstream_version_fallback %(ls -t dist/%{normalized_core_packaging_metadata_project_name}-*.tar.gz 2>/dev/null | head -n 1 | sed -E 's#^dist/%{normalized_core_packaging_metadata_project_name}-(.*)[.]tar[.]gz$#\1#')
 # If "upstream_version" macro is unset, use the fallback defined above:
 %if "%{!?upstream_version:UNSET}" == "UNSET"
 %global upstream_version %{upstream_version_fallback}
--- a/src/pylibsshext/errors.pyx
+++ b/src/pylibsshext/errors.pyx
@@ -34,6 +34,10 @@ cdef class LibsshSessionException(LibsshException):
     pass
 
 
+cdef class LibsshConfigParseException(LibsshSessionException):
+    """Raised when SSH config file parsing fails."""
+
+
 cdef class LibsshChannelException(LibsshException):
     pass
 
--- a/src/pylibsshext/errors.pyx
+++ b/src/pylibsshext/errors.pyx
@@ -35,7 +35,7 @@ cdef class LibsshSessionException(LibsshException):
 
 
 cdef class LibsshConfigParseException(LibsshSessionException):
-    """Raised when SSH config file parsing fails."""
+    """A failure to parse or load an SSH configuration file."""
 
 
 cdef class LibsshChannelException(LibsshException):
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -15,14 +15,13 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 
-import contextlib as _ctx
 import inspect
 import logging
 
 from cpython.bytes cimport PyBytes_AS_STRING
 
 from pylibsshext.channel import Channel
-from pylibsshext.errors cimport LibsshSessionException
+from pylibsshext.errors cimport LibsshConfigParseException, LibsshSessionException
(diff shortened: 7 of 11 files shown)


ganeshrn (commented):
If the PR is ready for review can you please move it out of draft

KB-perByte:
The CI is passing at this point!

Jakuje (approved):
looks good!
(reviewing just the config part as it includes also the logging part that was actually written by me in separate PR)

ganeshrn (approved):

ganeshrn:
@KB-perByte Is this variable `OPENSSH_PROXYJUMP`  required to be passed as a [task environment variable](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_environment.html) if yes we might have to update it as part of porting documentation from paramiko.

Jakuje:
> @KB-perByte Is this variable `OPENSSH_PROXYJUMP` required to be passed as a [task environment variable](https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_environment.html) if yes we might have to update it as part of porting documentation from paramiko.

The `OPENSSH_PROXYJUMP` should not be passed by default. We have now libssh 0.12 which should fix some corner cases of proxyjump handling in libssh so it would be better not to use the workaround with the variable and rather report if there will be some outstanding issues.

The `OPENSSH_PROXYJUMP` just makes libssh call openssh for proxyjumps, which is not ideal.

KB-perByte:
@ganeshrn, this PR is more concerned about the config file handling from a custom path, I am testing things with the newly released version of libssh, which should do justice to JumpHost configurations from configFiles. 

And as I have a +1 from @Jakuje for his changes for logging here in this PR, I'll update the changelog to reflect the same and get this merged.

Thanks for the reviews.

Jakuje:
Still ack. But the history now looks quite messy and is mixing 3 different things -- parsing of configuration file, logging and proxyjump. I would prefer to get the #597 merged separately (just rebased) and then rebase this + squash the fixup commits.

KB-perByte:
+1 I'll hold on to this PR, and rebase it with https://github.com/ansible/pylibssh/pull/597
@Jakuje

webknjaz:
Is this related to #479?

docs/changelog-fragments/790.bugfix.rst
```diff
@@ -0,0 +1,12 @@
+Fixed the ``config_file`` parameter being ignored in
+``Session.connect()``. When a custom SSH config file path was passed
+via the ``config_file`` keyword argument, it was not being processed
+because it was missing from the options mapping. Now the config file
+is properly parsed before establishing the connection, enabling SSH
+config features like host aliases and custom settings to work correctly.
+Logging module is also implemented which provides interface between libssh
+and python log facility, this supplements debug operations via ansible while
+using persistent logging. The changes are cherry picked from the actual
+implementation done by Jakuje in PR 597.
```
webknjaz: The correct way of linking prs/issues/gh is through the `sphinx-issues` roles:
```suggestion
implementation done by Jakuje in :pr:`597`.
```
(although, I'm hoping that PR to be absorbed separately)

KB-perByte:
converting this PR back to draft and removing logging specific changes out of this PR
and focus on this fix only https://github.com/ansible/pylibssh/pull/479

https://github.com/ansible/pylibssh/pull/597 to be considered for merge!

webknjaz:
> and focus on this fix only #479

@KB-perByte perhaps, just polish that PR? Or pull in their commits? Or if you go with this one, do make sure to implement https://hynek.me/til/easier-crediting-contributors-github/.

webknjaz:
@KB-perByte try to use rebasing instead of adding merge commits into topic branches since that leads to foxtrot merges in the tree.

src/pylibsshext/session.pyx
```diff
@@ -15,12 +15,15 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 import inspect
-import logging
 
 from cpython.bytes cimport PyBytes_AS_STRING
 
+import logging
```
webknjaz: Why did you move this?

docs/changelog-fragments/790.bugfix.rst
webknjaz, line 1: Please add docs/examples in addition to the change note. Also, this shouldn't be a `bugfix` — it seems like a feature to me so a different part of the change log should show this news fragment.

src/pylibsshext/session.pyx
```diff
@@ -15,12 +15,15 @@
...
+
 from pylibsshext.channel import Channel
+
 from pylibsshext.errors cimport LibsshSessionException
+
```
webknjaz: Any justification for these formatting changes? Why did you consider them related?

src/pylibsshext/session.pyx
```diff
@@ -263,6 +266,12 @@ cdef class Session(object):
             if (key in OPTS_MAP or key in OPTS_DIR_MAP) and (kwargs[key] is not None):
                 self.set_ssh_options(key, kwargs[key])
 
+        # Parse SSH config file after setting host (for Host matching) but before connecting
```
webknjaz: So this comment hints on the intention (sequencing). But could you expand on the motivation too?

src/pylibsshext/session.pyx
```diff
@@ -263,6 +266,12 @@ cdef class Session(object):
...
+        # Only parse if config_file is explicitly provided
+        config_file = kwargs.get('config_file')
+        if config_file is not None:
+            self.parse_config(config_file)
```
webknjaz: How about
```
import contextlib as _ctx
```
```suggestion
        with _ctx.suppress(KeyError):
            self.parse_config(kwargs['config_file'])
```
? Seems cleaner this way.

src/pylibsshext/session.pyx
```diff
@@ -542,6 +551,41 @@ cdef class Session(object):
     def sftp(self):
         return SFTP(self)
 
+    def parse_config(self, filename=None):
```
webknjaz: Let's start adding type annotations to new code:
```suggestion
    def parse_config(self, filename: bytes | str | None = None) -> None:
```

It's not difficult but will improve the quality of life long-term, going forward.

docs/changelog-fragments/790.bugfix.rst
```diff
@@ -0,0 +1,8 @@
+Fixed the ``config_file`` parameter being ignored in
```
webknjaz: I'd rephrase this in terms of how a new feature can be used. Plus once there's content in the docs, we could maybe link that too.

webknjaz (changes requested):
Here's a few things I'll expect before I can conclude that the patch does implement what it claims and that works (a.k.a. DoD):

- [x] The docs should show examples of how this is intended to be used.
- [ ] The tests should show how this functions too so there's no regressions.
- [x] Make sure to drop anything that's not directly related/refactoring/reformatting/etc — LLMs are good at generating way too much stuff and it even works under some circumstances but we're still responsible for keeping the patches maintainable, having Git tree tell a story and contain sufficient context. Be accountable.

docs/user_guide.rst
```diff
@@ -49,6 +49,23 @@ Connecting with remote SSH server
...
 
 
+.. _custom-ssh-config:
+
+Connecting with a custom SSH config
+-----------------------------------
+
+To use a custom SSH config file (e.g. for Host aliases, ``ProxyCommand``, or
+per-host options), pass ``config_file`` to :meth:`~pylibsshext.session.Session.connect`.
+The host option should be set so that Host matching in the config works correctly.
```
Jakuje: I would additionally mention here that

* the libssh loads default configuration from `~/.ssh/config` and `/etc/ssh/ssh_config` by default (if not overridden by this option).
* If you use this option, libssh will not consult the above two files

while it might be obvious for readers familiar with openssh, it might not be clear for people new to ssh in general

docs/_samples/ssh_config_connect.py
```diff
@@ -0,0 +1,30 @@
+"""Connect using a custom SSH config file.
+
+Use this when you have a dedicated config file (e.g. for Host aliases,
+ProxyCommand, or per-host options) and want the session to use it
+instead of the default ~/.ssh/config.
```
Jakuje: by default, libssh loads also `/etc/ssh/ssh_config`. would be good to have it also here for completeness.

KB-perByte: @Jakuje added, this was helpful.

Jakuje (approved):

ganeshrn (approved):

docs/user_guide.rst
```diff
@@ -49,6 +49,25 @@ Connecting with remote SSH server
    :emphasize-lines: 7-13
 
 
+.. _custom-ssh-config:
```
webknjaz, line 52: :books: Render preview: https://ansible-pylibssh--790.org.readthedocs.build/en/790/user_guide/#connecting-with-a-custom-ssh-config

docs/user_guide.rst
```diff
@@ -49,6 +49,25 @@ Connecting with remote SSH server
...
+aliases, ``ProxyCommand``, or per-host options), pass ``config_file`` to
+``Session.connect()``. When you pass ``config_file``, libssh will not consult
+the default files above. The host option should be set so that Host matching in
+the config works correctly.
+
+.. literalinclude:: _samples/ssh_config_connect.py
+   :language: python
+   :start-at: ssh = Session()
+   :end-before: if ssh.is_connected:
+   :emphasize-lines: 5-10
```
webknjaz: This looks wrong, looking at the render — could you explain why you decided to highlight exactly these lines? Seems off.

docs/user_guide.rst
```diff
@@ -49,6 +49,25 @@ Connecting with remote SSH server
    :emphasize-lines: 7-13
 
 
+.. _custom-ssh-config:
+
+Connecting with a custom SSH config
+-----------------------------------
+
+By default, libssh loads configuration from ``~/.ssh/config`` and
+``/etc/ssh/ssh_config``. To use a custom SSH config file instead (e.g. for Host
```
webknjaz: For paths, use the `:file:` role. Also, looks like you meant `Host` as a keyword, no?
```suggestion
By default, libssh loads configuration from :file:`~/.ssh/config` and
:file:`/etc/ssh/ssh_config`. To use a custom SSH config file instead (e.g. for ``Host``
```

docs/user_guide.rst
```diff
@@ -49,6 +49,25 @@ Connecting with remote SSH server
...
+aliases, ``ProxyCommand``, or per-host options), pass ``config_file`` to
+``Session.connect()``. When you pass ``config_file``, libssh will not consult
+the default files above. The host option should be set so that Host matching in
```
webknjaz:
```suggestion
the default files above. The ``host`` option should be set so that ``Host`` matching in
```

docs/user_guide.rst
webknjaz: *(otherwise, it probably shouldn't be capitalized — I wouldn't know w/o knowing what you intended)*

docs/user_guide.rst
```diff
@@ -49,6 +49,25 @@ Connecting with remote SSH server
    :emphasize-lines: 7-13
 
 
+.. _custom-ssh-config:
+
+Connecting with a custom SSH config
+-----------------------------------
+
+By default, libssh loads configuration from ``~/.ssh/config`` and
```
webknjaz: @Jakuje _for the future_ — perhaps we could add a Sphinx glossary entry and use it via the `:term:` role across the docs? Let's consider this later.

docs/changelog-fragments/790.feature.rst
```diff
@@ -0,0 +1,6 @@
+Added support for using a custom SSH config file when connecting. You can now pass
+``config_file`` to ``Session.connect()`` to parse a specific config file (e.g. for
+Host aliases, ``ProxyCommand``, or per-host settings) before the connection is
```
webknjaz: Could you explain why `Host` is capitalized but not highlighted? Is it a keyword in this context or not?

src/pylibsshext/session.pyx
```diff
@@ -256,13 +258,26 @@ cdef class Session(object):
...
         for key in kwargs:
             if (key in OPTS_MAP or key in OPTS_DIR_MAP) and (kwargs[key] is not None):
                 self.set_ssh_options(key, kwargs[key])
 
+        # Parse SSH config after host is set (so Host blocks match) but before connect,
+        # so that options from the config (e.g. HostName, User, Port, ProxyCommand) are
+        # applied to this connection. Only when config_file is passed explicitly.
+        with _ctx.suppress(KeyError):
+            path = kwargs['config_file']
+            if path is not None:
```
webknjaz, line 282: Why is it acceptable to have it as `None`? Also, why is `None` checked both on the calling side and in the helper method? It seems like this outer if-block is rather redundant. Make sure all 3 cases are covered with tests.

docs/_samples/ssh_config_connect.py
webknjaz, line 1: This seems to be lacking an example of the new public API — the `Session.parse_config()` method. When new APIs are added, it'd good to demo how they are expected to be called/handled.

src/pylibsshext/session.pyx
```diff
@@ -542,6 +557,41 @@ cdef class Session(object):
     def sftp(self):
         return SFTP(self)
 
+    def parse_config(self, filename: bytes | str | None = None) -> None:
+        """Parse SSH configuration file.
+
+        This parses the SSH configuration file and applies the settings
+        to the current session. If no filename is provided, it parses
+        the default configuration files (~/.ssh/config and system config).
```
webknjaz: The docstrings should generally use RST for highlighting for whenever we're able to extract this into the generated docs.
```suggestion
        the default configuration files (:file:`~/.ssh/config` and system config).
```

src/pylibsshext/session.pyx
```diff
@@ -542,6 +557,41 @@ cdef class Session(object):
...
+
+        Note: The host option should be set before calling this method
+        for Host matching to work correctly.
+
+        :param filename: Path to the SSH config file, or None for defaults.
+        :type filename: str or None
```
webknjaz: This does not match the type annotation that implies the `bytes` possibility too.
Though, in general, we don't need duplicate `:type:` entries in docstrings when there's respective annotations in the signature.
```suggestion
```

src/pylibsshext/session.pyx
```diff
@@ -542,6 +557,41 @@ cdef class Session(object):
...
+        if filename is not None:
+            if isinstance(filename, str):
+                b_filename = filename.encode("utf-8")
+            else:
+                b_filename = filename
+            c_filename = b_filename
+
+        rc = libssh.ssh_options_parse_config(self._libssh_session, c_filename)
+        if rc != libssh.SSH_OK:
+            raise LibsshSessionException(
```
webknjaz: We need a better exception subclass for this specific case. Also, make sure to have a test having `with pytest.raises(..., match=...):` proving that it actually happens.

src/pylibsshext/session.pyx
```diff
@@ -542,6 +557,41 @@ cdef class Session(object):
...
 
+    def parse_config(self, filename: bytes | str | None = None) -> None:
+        """Parse SSH configuration file.
+
+        This parses the SSH configuration file and applies the settings
+        to the current session. If no filename is provided, it parses
+        the default configuration files (~/.ssh/config and system config).
+
+        Note: The host option should be set before calling this method
+        for Host matching to work correctly.
```
webknjaz: Why is Host a proper name here?

src/pylibsshext/session.pyx
```diff
@@ -542,6 +557,41 @@ cdef class Session(object):
     def sftp(self):
         return SFTP(self)
 
+    def parse_config(self, filename: bytes | str | None = None) -> None:
```
webknjaz: Having looked through the patch as a whole, I must ask — is this *intended* to be a public API? If not — it should be made `_private()` instead (which would mean no need to document end-users calling it directly too).

webknjaz (changes requested):
So it seems like what's left are these two major matters to address:
* [x] the status of the public API vs. private (and the related docs)
* [x] the tests covering every new line added
* [ ] and there's a few inline notes about things that seem inconsistent

src/pylibsshext/session.pyx
```diff
@@ -14,13 +14,18 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
+
```
webknjaz: noooooooooooo plz nooo

docs/_samples/ssh_config_connect.py
```diff
@@ -0,0 +1,40 @@
...
+try:
+    ssh.connect(
+        host=HOST,
+        user=USER,
+        config_file=CONFIG_FILE,
+    )
+except LibsshConfigParseException as exc:
+    print(f'Failed to parse SSH config: {exc!s}')
+except LibsshSessionException as exc:
+    print(f'Failed to connect: {exc!s}')
```
webknjaz:
```suggestion
except LibsshConfigParseException as ssh_config_load_error:
    print(f'Failed to load the SSH config: {ssh_config_load_error!s}')
except LibsshSessionException as ssh_session_connection_error:
    print(f'Failed to connect: {ssh_session_connection_error!s}')
```

webknjaz: Though, are we sure `LibsshSessionException` is a *connection error*? The name implies a generic session handling problem, no mentions of the connection specifically in the docstring or the name. @Jakuje was it meant like that?

tests/unit/session_test.py
```diff
@@ -26,3 +32,43 @@ def test_session_connection_refused(free_port_num):
...
+        LibsshConfigParseException,
+        match=r'Failed to parse SSH config:',
+    ):
+        session.parse_config('/nonexistent/path/ssh_config')
+
+
+def test_parse_config_valid_file_succeeds():
+    """parse_config() with a valid config file does not raise."""
+    session = Session(host='somehost')
+    with tempfile.NamedTemporaryFile(
```
webknjaz: Use `tmp_path` instead. No need for manual tmp dir management.

tests/unit/session_test.py
```diff
@@ -26,3 +32,43 @@ def test_session_connection_refused(free_port_num):
...
+        mode='w',
+        suffix='.config',
+        delete=False,
+    ) as f:
+        f.write('Host *\n')
+        path = f.name
+    try:
+        session.parse_config(path)
+    except LibsshConfigParseException:
+        pytest.skip(
```
webknjaz, line 78: This doesn't seem justified

tests/unit/session_test.py
KB-perByte: I find this much cleaner to write and maintain in the longer run, any specific reason to do things otherwise?

webknjaz: Yes, these are `pytest` based tests and we should be using native built-in helpers that are expected to be seen in such tests. Otherwise, you'd be forcing every reader to figure out how your DIY thing works and try to guess why you chose an exotic structure instead of what's expected.

src/pylibsshext/session.pyx
```diff
@@ -542,6 +558,37 @@ cdef class Session(object):
...
+            if isinstance(filename, str):
+                b_filename = filename.encode("utf-8")
+            else:
+                b_filename = filename
+            c_filename = b_filename
+
+        rc = libssh.ssh_options_parse_config(self._libssh_session, c_filename)
+        if rc != libssh.SSH_OK:
+            raise LibsshConfigParseException(
+                "Failed to parse SSH config: %s" % self._get_session_error_str(),
```
webknjaz: FTR legacy str formatting is slower, which is why today it's better to use f-strings (due to performance, though, I'd recommend extracting the error into a var for readability too):
```suggestion
            underlying_libssh_error = self._get_session_error_str()
            raise LibsshConfigParseException(
                f"Failed to parse SSH config: {underlying_libssh_error!s}",
```

tests/unit/session_test.py
```diff
@@ -2,7 +2,10 @@
 
 import pytest
```
webknjaz, line 7:
```suggestion
import typing as _t

import pytest
```

src/pylibsshext/session.pyx
```diff
@@ -542,6 +558,37 @@ cdef class Session(object):
...
+        :param filename: Path to the SSH config file, or ``None`` for defaults.
+
+        :raises LibsshConfigParseException: If parsing fails.
+        """
+        cdef int rc
+        cdef bytes b_filename
+        cdef const char *c_filename = NULL
+
+        if filename is not None:
+            if isinstance(filename, str):
```
webknjaz: Pretty sure it's a good idea to also accept a native `pathlib.Path` — it's a much better API to have. I'd even argue that unicode and byte strings should be rejected forcing a unified API for the callers.

tests/unit/session_test.py
webknjaz:
```suggestion
import typing as _t
from pathlib import Path

import pytest
```

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+
+
+def test_parse_config_nonexistent_raises():
+    """parse_config() with nonexistent file raises LibsshConfigParseException."""
+    session = Session(host='somehost')
+    with pytest.raises(
+        LibsshConfigParseException,
+        match=r'Failed to parse SSH config:',
+    ):
+        session.parse_config('/nonexistent/path/ssh_config')
```
webknjaz: A properly isolated test must create a 100% non-existent path by injecting the `tmp_path` fixture and passing a name under that dir — this is what guarrantees the file non-existence. Otherwise, it's possible to break this test by creating a file at this path prior to invoking the test runner.

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+
+
+def test_parse_config_valid_file_succeeds(tmp_path):
```
webknjaz: You still want to go for accepting multiple types, this should be parametrized:
```suggestion
@pytest.mark.parametrize(
    'path_arg_type',
    (bytes, str, Path),
    ids=str,
)
def test_parse_config_valid_file_succeeds(
    path_arg_type: _t.Type[bytes] | _t.Type[str] | _t.Type[Path],
    tmp_path: Path,
) -> None:
```

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+    """parse_config() with a valid config file does not raise."""
+    session = Session(host='somehost')
+    config_file = tmp_path / 'ssh_config'
+    config_file.write_text('Host *\n', encoding='utf-8')
+    try:
+        session.parse_config(str(config_file))
+    except LibsshConfigParseException:
+        pytest.skip(
+            'libssh rejected minimal config file (path expansion or parsing is strict)',
+        )
```
webknjaz, line 80: Then, you'd use the param.
```suggestion
    session.parse_config(path_arg_type(config_file))
```

Although, I think I'd prefer a better API that would reject anything non-pathlib.

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+
+
+def test_connect_config_file_none_no_parse():
+    """connect(config_file=None) does not call parse_config; fails at connect."""
+    session = Session(host='test.nonexistent.example.invalid')
+    with pytest.raises(LibsshSessionException, match=r'ssh connect failed:'):
+        session.connect(
+            host='test.nonexistent.example.invalid',
```
webknjaz, line 91: does the host arg need to be duplicated here if it's already in the session?

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+    except LibsshConfigParseException:
+        pytest.skip(
+            'libssh rejected minimal config file (path expansion or parsing is strict)',
+        )
+
+
+def test_connect_config_file_none_no_parse():
+    """connect(config_file=None) does not call parse_config; fails at connect."""
+    session = Session(host='test.nonexistent.example.invalid')
+    with pytest.raises(LibsshSessionException, match=r'ssh connect failed:'):
```
webknjaz: Best be most specific with error matching:
```suggestion
    expected_error_regex = r'^ssh connect failed: .*test.nonexistent.example.invalid.*$'
    with pytest.raises(LibsshSessionException, match=expected_error_regex):
```

tests/unit/session_test.py
```diff
@@ -26,3 +29,36 @@ def test_session_connection_refused(free_port_num):
...
+    config_file.write_text('Host *\n', encoding='utf-8')
+    try:
+        session.parse_config(str(config_file))
+    except LibsshConfigParseException:
+        pytest.skip(
+            'libssh rejected minimal config file (path expansion or parsing is strict)',
+        )
+
+
+def test_connect_config_file_none_no_parse():
```
webknjaz, line 83: Seems like `test_parse_config_nonexistent_raises` is practically the same and they could be merged though parametrize with different error + host + file in params.

src/pylibsshext/session.pyx
```diff
@@ -272,11 +271,10 @@ cdef class Session(object):
 
         # Parse SSH config after host is set (so Host blocks match) but before connect,
         # so that options from the config (e.g. HostName, User, Port, ProxyCommand) are
-        # applied to this connection. Only when config_file is passed explicitly.
-        with _ctx.suppress(KeyError):
```
webknjaz: Unclear why you reverted this — the CM bit was good, the `None` double-check still isn't.

webknjaz (changes requested):
I'll wait until here's responses to the questions I'd already asked before I leave any more feedback.

*UPD (06.03.2026):* @KB-perByte and I had a 1:1 call the other day to clear some things up and synchronize the understanding. This is now waiting for his changes which will likely happen next week as I believe we're on the same page now. Meanwhile, I'll finish up #808 and cut a release right after (hopefully today!) to make the logging improvements available early.

docs/changelog-fragments/790.feature.rst
```diff
@@ -0,0 +1,6 @@
+Added support for using a custom SSH config file when connecting. You can now pass
+``config_file`` to ``Session.connect()`` to parse a specific config file (e.g. for
+``Host`` aliases, ``ProxyCommand``, or per-host settings) before the connection is
+established. See :ref:`Connecting with a custom SSH config <custom-ssh-config>` for examples.
+
+-- by :user:`KB-perByte`
```
webknjaz: Since this includes bits of #479, let's remember to credit the contributor:
```suggestion
-- by :user:`donnerhacke` and :user:`KB-perByte`
```

docs/changelog-fragments/790.feature.rst
webknjaz, line 1: Same as with the byline, add a symlink mentioning the other PR:
```console
$ (cd docs/changelog-fragments/ && ln -sf {790,479}.feature.rst && git add 479.feature.rst)
$ git commit -m '📝 Link PR #479 to PR #790 change note'
```

webknjaz (commented):
I did another quick pass and realized that another thing I mentioned earlier somewhere hasn't been followed up — there's an earlier PR #479 that somebody took time to file. Even though it remains unmerged, let's be nice and credit the contributor.

Inline, I've left change note-specific instructions but it'd be good to add a co-authored-by trailer to the commit changing `libssh.pxd` following https://hynek.me/til/easier-crediting-contributors-github/.

src/pylibsshext/errors.pyx
```diff
@@ -34,6 +34,10 @@ cdef class LibsshSessionException(LibsshException):
     pass
 
 
+cdef class LibsshConfigParseException(LibsshSessionException):
+    """Raised when SSH config file parsing fails."""
```
webknjaz: Just realized that this docstring references some external logic that may change over time in a disconnected manner. That behavior description doesn't belong here.
The docstring should instead explain what the exception represents. How about
```suggestion
    """A failure to parse or load an SSH configuration file."""
```
?

src/pylibsshext/session.pyx
```diff
@@ -256,13 +260,25 @@ cdef class Session(object):
...
         for key in kwargs:
             if (key in OPTS_MAP or key in OPTS_DIR_MAP) and (kwargs[key] is not None):
                 self.set_ssh_options(key, kwargs[key])
 
+        # Parse SSH config after host is set (so Host blocks match) but before connect,
+        # so that options from the config (e.g. HostName, User, Port, ProxyCommand) are
+        # applied to this connection. Only when config_file is passed and non-None.
+        path = kwargs.get('config_file')
+        if path is not None:
+            self.parse_config(path)
```
webknjaz: Why can't this be
```suggestion
        self.parse_config(path)
```
/
```suggestion
        self.load_config_into_session(path)
```
/
```suggestion
        self._parse_config(path)
```
/
```suggestion
        self._load_config_into_session(path)
```
?

src/pylibsshext/session.pyx
webknjaz: One more point to consider — it's probably best to either allow positional- or keyword-only args but not both. For single-arg APIs it's usually good to go w/ pos-only — this way, the callers would only have to think of the method name and the data they are passing in and not how it's referenced internally.

```python
import pathlib as _pth
```

Here's a few suggestions on how to improve the signature based on our call with variants for public and private APIs, with the above idea in mind and an API naming improvement to possibly entertain:

Public variants:
```suggestion
    def parse_config(self, filename: _pth.Path | None = None, /) -> None:
```

```suggestion
    def load_config_into_session(self, filename: _pth.Path | None = None, /) -> None:
```

The private way:
```suggestion
    def _parse_config(self, filename: _pth.Path | None, /) -> None:
```

```suggestion
    def _load_config_into_session(self, filename: _pth.Path | None, /) -> None:
```

> [!tip]
> As I'm reading more into this, I realized that it's a bit unclear what the semantics of the `None` sentinel is here. The docstring mentions it as a default but does passing `None` mean that `libssh` will attempt to read the standard config that it knows of? Additionally, does this API *need* a default? Not sure about the public variants but if it turns into a private method, I'd definitely recommend against defaults in favor of forcing the callers (that'd be us!) to pass something in.

KB-perByte:
@webknjaz Ithe CI is not stable, I'll get back to this PR again later. The failures are unrelated.
I could address most of the revirew comments.

Jakuje:
https://github.com/ansible/pylibssh/actions/runs/22965988172/job/66671296544?pr=790

>   urllib.error.HTTPError: HTTP Error 429: Too Many Requests

sounds like the downloads are failing due to some rate limiting ...

But the failure in the following pipeline sounds related:

https://github.com/ansible/pylibssh/actions/runs/22965988172/job/66671296430?pr=790#step:6:802

>   /project/tests/unit/session_test.py:54: in <module>
>       path_arg_type: type[bytes] | type[str] | type[Path],
>   E   TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'types.GenericAlias'

The error in https://github.com/ansible/pylibssh/actions/runs/22965988172/job/66671296510?pr=790 sounds related to changes in this PR too:

>   src/pylibsshext/session.pyx:562:71: Expected an identifier, found '/'

KB-perByte:
@Jakuje checking the same, Cython doesn't support PEP 570's positional-only parameter syntax (the / in the function signature) getting em fixed

KB-perByte:
https://github.com/ansible/pylibssh/actions/runs/22989405548/job/66747015702?pr=790#step:6:802 I am not 100% sure what would the right way to fix these is, checking for the possible fixes. 
```

  ==================================== ERRORS ====================================
  _________________ ERROR collecting tests/unit/session_test.py __________________
  /project/tests/unit/session_test.py:54: in <module>
      path_arg_type: type[bytes] | type[str] | type[Path],
  E   TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'types.GenericAlias'
          LibsshConfigParseException = <class 'pylibsshext.errors.LibsshConfigParseException'>
          LibsshSessionException = <class 'pylibsshext.errors.LibsshSessionException'>
          Path       = <class 'pathlib.Path'>
          Session    = <class 'pylibsshext.session.Session'>
          __builtins__ = <builtins>
          __cached__ = '/project/tests/unit/__pycache__/session_test.cpython-39.pyc'
          __doc__    = 'Tests suite for session.'
          __file__   = '/project/tests/unit/session_test.py'
          __loader__ = <_pytest.assertion.rewrite.AssertionRewritingHook object at 0x7fca12a26ca0>
          __name__   = 'session_test'
          __package__ = ''
```

webknjaz:
> add annotations for 2.9

What's this mysterious "2.9"?

packaging/rpm/ansible-pylibssh.spec
```diff
@@ -1,11 +1,11 @@
 %global canonical_project_name ansible-pylibssh
 %global pypi_name %canonical_project_name
-%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:\\.:_:g')
+%global normalized_core_packaging_metadata_project_name %(echo '%canonical_project_name' | sed --regexp-extended 's:-:_:g;s:[.]:_:g')
```
webknjaz, line 3: It's possible that the fix will only be needed on this line. The error was caused by https://github.com/ansible/pylibssh/commit/c94f365b5e27397076ecbe09329036262b8adb29 somehow, it seems, not sure why. But line 8 only got changes to the var name so it's unlikely that it needs changing. I'll look into it further — it's surprising that my smoke test script doesn't fail on this locally, nor do the GHA jobs, it seems — looks like Packit might have different versions of lua or something in its env compared to the regular Fedora and UBI images.

docs/_samples/ssh_config_connect.py
```diff
@@ -0,0 +1,45 @@
...
+        config_file=CONFIG_FILE,
+    )
+except LibsshConfigParseException as ssh_config_load_error:
+    print(f'Failed to load the SSH config: {ssh_config_load_error!s}')
+except LibsshSessionException as ssh_session_connection_error:
+    print(f'Failed to connect: {ssh_session_connection_error!s}')
+
+# Alternative: call parse_config() explicitly (e.g. with a pathlib.Path)
+# ssh = Session(host=HOST)
+# ssh.parse_config(Path(CONFIG_FILE))  # or str/bytes
```
webknjaz: Why do you insist on bytes of unguaranteed encoding in the API?

src/pylibsshext/session.pyx
```diff
@@ -115,6 +121,8 @@ cdef class Session(object):
         if self._libssh_session is NULL:
             raise MemoryError
         self._opts = {}
+        if host is not None:
+            self.set_ssh_options('host', host)
```
webknjaz: This looks new.. How is it related?

tests/unit/session_test.py
```diff
@@ -26,3 +33,52 @@ def test_session_connection_refused(free_port_num):
...
+        session.parse_config(str(nonexistent))
+
+
+@pytest.mark.parametrize(
+    'path_arg_type',
+    (bytes, str, Path),
+    ids=str,
+)
+def test_parse_config_valid_file_succeeds(
+    path_arg_type: type[bytes | str | Path],
```
webknjaz: This doesn't sound right. It shouldn't be a type of a union but a union of types, I think. Also, the `bytes` controversy in the API stands.

tests/unit/session_test.py
webknjaz: (meaning the solution is https://github.com/ansible/pylibssh/pull/790/files#r2878365547)

src/pylibsshext/session.pyx
Jakuje: this looks like the `host` argument was never handled through the `kwargs` as the rest of the options. It should land likely at least as a separate bugfix commit, I think.

packaging/rpm/ansible-pylibssh.spec
webknjaz, line 1: I'm not seeing this surfacing in other PRs. So just drop the commit that added it when *rebasing* the next time.

src/pylibsshext/session.pyx
webknjaz: Yeah, that's what I thought. A separate bug rather than a feature which this one is. And a separate change note.

webknjaz:
FYI, I'm looking at adding linting for requiring not creating foxtrot merges automatically.

KB-perByte:
Apologies for the delay in activity. I am trying to progress multiple items in various repos.
It takes me quite some time to get back to review comments.

We would need to prioritize this PR, to get the broken functionality fixed wrt to netcommon.
We can have another iteration of the fixes if those are not must-have.

webknjaz:
@KB-perByte please perform an actual rebase instead of adding new merge commits to the branch.

webknjaz:
@KB-perByte hey, could you use interactive rebase to absorb automatic reformatting commits into respective functional commits and also squash relative functional commits, reshaping the commit sequence in the branch to tell a more comprehensive story with a better story of what each commit (change) effect is transparently?

This will probably end up being like 4 commits instead of 23 — tests, functional (runtime code) implementation, tests and changelog, once polished.

Also, please check out all the inline threads and respond in there so we could decide how to proceed from there.

easpeagle:
What's it going to take to get this one going again... this feature is a rather important one. Especially since ansible.netcommon.libssh already calls this out as a valid feature.

https://docs.ansible.com/projects/ansible/latest/collections/ansible/netcommon/libssh_connection.html#parameter-config_file

PR #791, opened 2026-01-28
Update macOS VMs in CI/CD to v15 on Intel arch

komaldesai13:
##### SUMMARY

CI run fails due to `macos-13` getting dropped by GitHub, this PR updates `runner-vm-os` to `macos-15-intel` to address that.

##### ISSUE TYPE

- Maintenance Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -579,11 +579,11 @@ jobs:
     strategy:
       matrix:
         runner-vm-os:
-        - macos-13
+        - macos-15-intel
     uses: ./.github/workflows/reusable-cibuildwheel.yml
     with:
       runner-vm-os: ${{ matrix.runner-vm-os }}
-      timeout-minutes: 9
+      timeout-minutes: 10
       source-tarball-name: >-
         ${{ needs.pre-setup.outputs.sdist-artifact-name }}
       dists-artifact-name: >-
@@ -1168,7 +1168,7 @@ jobs:
         - "3.10"
         - 3.9
         runner-vm-os:
-        - macos-13
+        - macos-15-intel
         dist-type:
         - binary
         - source
--- a/docs/changelog-fragments/791.contrib.rst
+++ b/docs/changelog-fragments/791.contrib.rst
@@ -1,0 +1,2 @@
+Added "macos-15-intel" as vm runner as "macos-13" is not being supported anymore
+The change includes giving the macOS build job more time to complete as the new version is a little slower -- by :user:`komaldesai13`
--- a/docs/changelog-fragments/791.contrib.rst
+++ b/docs/changelog-fragments/791.contrib.rst
@@ -1,2 +1,3 @@
-Added "macos-15-intel" as vm runner as "macos-13" is not being supported anymore
-The change includes giving the macOS build job more time to complete as the new version is a little slower -- by :user:`komaldesai13`
+Now that the ``macos-13`` runner VM image has been decommissioned, the CI and CD jobs have been migrated to use ``macos-15-intel`` -- by :user:`komaldesai13`.
+
+This includes giving the macOS build job more time to complete as the new version is a little slower.


docs/changelog-fragments/791.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added "macos-15-intel" as vm runner as "macos-13" is not being supported anymore
+-- by :user:`komaldesai13`
```
webknjaz:
```suggestion
Replaced the decommissioned `macos-13` runner VM image with `macos-15-intel` in CI and CD jobs
-- by :user:`komaldesai13`.
```

webknjaz (changes requested):
Plz also replace it in the test jobs below and grep the workflow for any other mentions of this VM version.

Additionally, raise the timeout to 10 minutes in the build job — looks like the new runners might be slower.

.github/workflows/ci-cd.yml
```diff
@@ -1199,7 +1199,7 @@ jobs:
     - pre-setup  # transitive, for accessing settings
     runs-on: ubuntu-latest
 
-    timeout-minutes: 1
+    timeout-minutes: 10
```
webknjaz: This job doesn't need so much time. It just verifies metadata and is very fast.

I saw the job that builds macOS wheels timing out because it was unable to complete in 9 minutes. That one does need the timeout to be raised a bit. On the line 586.

```suggestion
    timeout-minutes: 1
```

docs/changelog-fragments/791.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Replaced the decommissioned `macos-13` runner VM image with `macos-15-intel` in CI and CD jobs
+-- by :user:`komaldesai13`.
```
webknjaz:
```suggestion
-- by :user:`komaldesai13`.

The change includes giving the macOS build job more time to complete as the new version is a little slower.
```

webknjaz (changes requested):

webknjaz:
Could you also combine related commits, giving them a good commit message?

docs/changelog-fragments/791.contrib.rst
webknjaz, line 1: @komaldesai13 it's best to accept the suggested changes as is and preview what it looks like. Here's a PR preview: https://ansible-pylibssh--791.org.readthedocs.build/en/791/changelog/#changelog. With your rephrasing, the rendering is broken since RST merges the lines. Plus there's no periods in sentences.

We should be careful to write change notes that follow the same style. These previews I've set up will help you with judging that. The goal is to have consistent style across the change log. You can check RTD builds after pushing PR updates but you can also execute something like `tox r -qq -e build-docs` and check it locally even sooner.

Additionally, spell known abbreviations with capital letters as sequences like "vm" aren't actual words. Though, in most cases it's best to avoid abbreviations and use full words, which reduces cognitive load for less technical readers. Note that RST differs from Markdown: for example, double backticks are used for inline code.

The change log style differs from Git commits since this is a combination of changes that has happened since the previous release, combined in one doc. It's more high-level and often uses something like the past tense or the present simple with the reference to what's changed. While Git commits typically use the imperative mood and focus on the motivation and decisions conserning other developers/contributors but scoped to as smaller change. So in change logs we avoid jargon and make them presentable to a wider audience. In both cases, we should avoid being vague about things and be specific.

As for the byline — we either add it at the end of the sentence and then add a period right after or we put it on a separate lines (with a separator in RST) and don't add a full stop since it's more standalone and isn't a sentence.

docs/changelog-fragments/791.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added "macos-15-intel" as vm runner as "macos-13" is not being supported anymore
+The change includes giving the macOS build job more time to complete as the new version is a little slower -- by :user:`komaldesai13`
```
webknjaz: This will give the change log a more consistent look
```suggestion
Now that the ``macos-13`` runner VM image has been decommissioned, the CI and CD jobs have been migrated to use ``macos-15-intel`` -- by :user:`komaldesai13`.

This includes giving the macOS build job more time to complete as the new version is a little slower.
```

webknjaz (changes requested):

webknjaz (approved):

PR #795, opened 2026-02-03
Add two scripts to generate universal2 artifacts for macOS

komaldesai13:
##### SUMMARY

This patch adds two shell scripts for building OpenSSL and libssh for use in the Python packaging process.

1) `build-libssh-macos.sh`: which takes openssl version, libssh version and macos arch as input and generate artifacts.

2) `merge-universal2.sh`: merges x86 and arm64 artifacts and generates a universal2 libssh build for macOS.

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```


webknjaz:
@komaldesai13 you can select just one type in the PR description, not all three. Let's describe such things accurately so the readers aren't left guessing what this is all about.

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+set -e
```
webknjaz: A good mode is
```suggestion

set -eEuo pipefail
```

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Prompt for build parameters
+read -p "libssh version [0.10.6]: " LIBSSH_VERSION
+LIBSSH_VERSION=${LIBSSH_VERSION:-0.10.6}
+
+read -p "OpenSSL version [3.0.13]: " OPENSSL_VERSION
+OPENSSL_VERSION=${OPENSSL_VERSION:-3.0.13}
+
+read -p "Architecture (arm64/x86_64) [$(uname -m)]: " ARCH
+ARCH=${ARCH:-$(uname -m)}
```
webknjaz: These should be just CLI args. Although the OpenSSL version should be sourced from https://github.com/ansible/pylibssh/blob/devel/build-scripts/manylinux-container-image/openssl-version.sh. We should eventually implement sourcing the `libssh` version from a single source too (perhaps, parse it from the `pyproject.toml`).

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+
+OUTPUT_DIR="artifacts/${ARCH}"
+mkdir -p ${OUTPUT_DIR}
```
webknjaz: I wouldn't make these dirs in arbitrary locations in-tree. Maybe in `build/` that's gitignored: https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/.gitignore#L190.

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+
+WORK_DIR=$(mktemp -d)
+cd ${WORK_DIR}
+
+echo "Building in ${WORK_DIR}"
+echo "libssh: ${LIBSSH_VERSION}, OpenSSL: ${OPENSSL_VERSION}, Arch: ${ARCH}"
+
+export MACOSX_DEPLOYMENT_TARGET="10.13"
```
webknjaz: Looks like upstream has different targets for the artifacts: https://github.com/pyca/infra/blob/fb1f9b3b26f94d384b1392c2388771a4f53b6fa4/.github/workflows/build-macos-openssl.yml#L31C23-L31C44. Can we mimic that?

packaging/macos/build-libssh-macos.sh
webknjaz: We probably need to store the absolute path somewhere so it could be used to set env vars similar to https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/pyproject.toml#L236-L242.

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ "${ARCH}" = "arm64" ]; then
+    OPENSSL_TARGET="darwin64-arm64-cc"
+else
+    OPENSSL_TARGET="darwin64-x86_64-cc"
+fi
+
+./Configure ${OPENSSL_TARGET} --prefix=${WORK_DIR}/deps no-shared no-tests
+make -j$(sysctl -n hw.ncpu) > /dev/null
+make install_sw > /dev/null
+cd ..
```
webknjaz, line 61: Can't we extend/reuse https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/build-scripts/manylinux-container-image/install_openssl.sh for this?

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+# Build OpenSSL
+echo "Downloading OpenSSL..."
+curl -sL https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz | tar xz
+cd openssl-${OPENSSL_VERSION}
+
+if [ "${ARCH}" = "arm64" ]; then
+    OPENSSL_TARGET="darwin64-arm64-cc"
+else
+    OPENSSL_TARGET="darwin64-x86_64-cc"
+fi
```
webknjaz: Looks like this could be
```suggestion
OPENSSL_TARGET="darwin64-${ARCH}-cc"
```

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
+    -DOPENSSL_ROOT_DIR=${WORK_DIR}/deps \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DWITH_EXAMPLES=OFF \
+    -DWITH_SERVER=OFF \
+    -DUNIT_TESTING=OFF \
+    > /dev/null
+
+make -j$(sysctl -n hw.ncpu) > /dev/null
+make install > /dev/null
```
webknjaz, line 84: Try reusing https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/build-scripts/manylinux-container-image/install_libssh.sh

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+
+# Package artifacts
+cd ${WORK_DIR}/deps
+ARTIFACT_NAME="libssh-${LIBSSH_VERSION}-openssl-${OPENSSL_VERSION}-macos-${ARCH}"
+mkdir -p ${ARTIFACT_NAME}/{lib,include}
```
webknjaz: Verbosity is useful in scripts.
```suggestion
mkdir -pv ${ARTIFACT_NAME}/{lib,include}
```

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+cmake .. \
+    -DCMAKE_INSTALL_PREFIX=${WORK_DIR}/deps \
+    -DCMAKE_OSX_ARCHITECTURES=${ARCH} \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
+    -DOPENSSL_ROOT_DIR=${WORK_DIR}/deps \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DWITH_EXAMPLES=OFF \
+    -DWITH_SERVER=OFF \
+    -DUNIT_TESTING=OFF \
+    > /dev/null
```
webknjaz: Let's avoid preventing people from understanding what's happening.

```suggestion
```

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+./Configure ${OPENSSL_TARGET} --prefix=${WORK_DIR}/deps no-shared no-tests
+make -j$(sysctl -n hw.ncpu) > /dev/null
+make install_sw > /dev/null
+cd ..
+
+# Build libssh
+echo "Downloading libssh..."
+curl -sL https://www.libssh.org/files/$(echo ${LIBSSH_VERSION} | cut -d. -f1-2)/libssh-${LIBSSH_VERSION}.tar.xz | tar xJ
+cd libssh-${LIBSSH_VERSION}
+mkdir build && cd build
```
webknjaz: Prefer `pushd`/`popd`

packaging/macos/build-libssh-macos.sh
```diff
@@ -0,0 +1,79 @@
...
+curl -sL https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz | tar xz
+cd openssl-${OPENSSL_VERSION}
+
+if [ "${ARCH}" = "arm64" ]; then
+    OPENSSL_TARGET="darwin64-arm64-cc"
+else
+    OPENSSL_TARGET="darwin64-x86_64-cc"
+fi
+
+./Configure ${OPENSSL_TARGET} --prefix=${WORK_DIR}/deps no-shared no-tests
```
webknjaz, line 58: Looks like `CFLAGS` should be set too: https://github.com/pyca/infra/blob/fb1f9b3b26f94d384b1392c2388771a4f53b6fa4/.github/workflows/build-macos-openssl.yml#L27

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
...
+    echo "Error: Input directories don't exist"
+    exit 1
+fi
+
+mkdir -p ${OUTPUT_DIR}/{lib,include}
+
+echo "Merging ${ARM64_DIR} and ${X86_64_DIR} into ${OUTPUT_DIR}"
+
+# Copy headers from arm64 (should be identical)
+cp -r ${ARM64_DIR}/include/* ${OUTPUT_DIR}/include/
```
webknjaz: This would be a bit more elegant and doesn't require pre-creating the includes dir:
```suggestion
# Copy headers from arm64 (should be identical)
cp -rv "${ARM64_DIR}"/include "${OUTPUT_DIR}"/
```

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
...
+ARM64_DIR=$1
+X86_64_DIR=$2
+OUTPUT_DIR=$3
+
+if [ ! -d "${ARM64_DIR}" ] || [ ! -d "${X86_64_DIR}" ]; then
+    echo "Error: Input directories don't exist"
+    exit 1
+fi
+
+mkdir -p ${OUTPUT_DIR}/{lib,include}
```
webknjaz: Don't forget to quote vars to prevent problems with whitespaces.
```suggestion
mkdir -pv "${OUTPUT_DIR}"/lib
```

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -e
```
webknjaz:
```suggestion

set -eEuo pipefail
```

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
...
+
+mkdir -p ${OUTPUT_DIR}/{lib,include}
+
+echo "Merging ${ARM64_DIR} and ${X86_64_DIR} into ${OUTPUT_DIR}"
+
+# Copy headers from arm64 (should be identical)
+cp -r ${ARM64_DIR}/include/* ${OUTPUT_DIR}/include/
+
+# Merge libraries using lipo
+for lib in ${ARM64_DIR}/lib/*.a; do
```
webknjaz: Are they just `.a` or `.so` / `.dylib` too?
```suggestion
for lib in "${ARM64_DIR}"/lib/*.a; do
```

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
...
+    libname=$(basename ${lib})
+    if [ -f "${X86_64_DIR}/lib/${libname}" ]; then
```
webknjaz, line 34: This seems simpler:
https://github.com/pyca/infra/blob/fb1f9b3b26f94d384b1392c2388771a4f53b6fa4/.github/workflows/build-macos-openssl.yml#L91-L92. Do we really need a loop and a conditional block here? How many libs are there?

packaging/macos/merge-universal2.sh
webknjaz: Also, https://github.com/pyca/infra/blob/fb1f9b3b26f94d384b1392c2388771a4f53b6fa4/.github/workflows/build-macos-openssl.yml#L89-L90 handles a `pkgconfig/` too.

packaging/macos/merge-universal2.sh
```diff
@@ -0,0 +1,36 @@
...
+        echo "Merging ${libname}"
+        lipo -create ${ARM64_DIR}/lib/${libname} ${X86_64_DIR}/lib/${libname} \
+            -output ${OUTPUT_DIR}/lib/${libname}
+    fi
+done
+
+echo "Done. Universal2 artifacts in ${OUTPUT_DIR}/"
+file ${OUTPUT_DIR}/lib/*.a | head -3
```
webknjaz: Why 3? what's the purpose of hiding the others? We should probably store the base dir path for later integration with the build system on the Python side.

webknjaz:
@komaldesai13 could you see if it'd be possible to replace https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/pyproject.toml#L246 with your scripts and run `cibuildwheel` locally? Perhaps, `brew uninstall libssh` before doing that. Pay attention at whether your scripts leave Git in a dirty state or create untracked files in the working directory. We should strive to unify compiling these components in the same manner so the version numbers could be updated once in the repository and influence both macOS and GNU/Linux build scripts.

packaging/macos/build-libssh-macos.sh
komaldesai13: well , there are several differences in command we use for macos , @webknjaz do we want to extend install_openssl.sh and install_libssh.sh or i can create something like install_openssl_macos.sh and install_libssh_macos.sh

webknjaz: I'm not sure. I'd need to see it in action or at least tell me of specific differences. If possible, attach comments to lines you're referring to for context.

webknjaz:
@KB-perByte let's not create merge commits in this branch — we'll have to remember to clean them up, which may complicate the process depending on @komaldesai13's experience with using Git.

packaging/macos/build-libssh-macos.sh
komaldesai13: done

packaging/macos/merge-universal2.sh
komaldesai13: @webknjaz i  have got just `.a`  files  for  `arm64` but this is not tested in arm64 machine as i have x86 machine and i am unable to spin aws vm also form arm64 it has got some `dedicated host not available` error  , i just stimulated facts , any suggestion how to test on actual arm64

webknjaz: I wouldn't worry about testing locally for now. I recommend focusing on making it integrated w/ x86_64 such that running `cibuildwheel` is able to pick it up. Then, we'll be able to see how it behaves in the CI, including arm64.

packaging/macos/merge-universal2.sh
komaldesai13: addressed and remobed` head -3`

packaging/macos/build-libssh-macos.sh
komaldesai13: addressed

packaging/macos/build-libssh-macos.sh
komaldesai13: addressed

packaging/macos/build-libssh-macos.sh
komaldesai13: Differences:
  1. Platform targets: The Linux script uses different OpenSSL config options, while macOS needs darwin64-arm64-cc or darwin64-x86_64-cc
  2. Installation path: Linux installs to /opt/pyca/cryptography/openssl, but macOS needs a temporary WORK_DIR since it packages everything into a tarball

however ,we can extract the OpenSSL SHA256 checksum verification logic from install_openssl.sh

packaging/macos/merge-universal2.sh
komaldesai13: addressed

packaging/macos/merge-universal2.sh
komaldesai13: there are 3 libs but we have added condition and loop for robustness

webknjaz:
@komaldesai13 best avoid adding merge commits into this branch.

packaging/macos/build-libssh-macos.sh
webknjaz: Sounds like the differences are minimal and a few things can be put into variables w/o the need to duplicate way too much.

packaging/.DS_Store
webknjaz, line 1: why didn't gitignore exclude this?

.DS_Store
webknjaz, line 1: huh?

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -17,6 +17,43 @@
...
+def _load_macos_build_config():
+    """Load macOS build configuration from .macos_build_config.py if it exists."""
+    config_path = Path('.macos_build_config.py')
+    if not config_path.exists():
+        return {}
+
+    try:
+        config = {}
+        with Path(config_path).open() as f:
+            exec(f.read(), config)
```
webknjaz, line 30: Ouch.. this is a straight way to a CVE (timing attack)

packaging/pep517_backend/_cython_configuration.py
webknjaz, line 1: We shouldn't update anything in the build backend. Set any flags outside, just like with other platforms.

pyproject.toml
```diff
@@ -242,38 +242,17 @@ LDFLAGS = "'-L$(cat /root/.static-deps-path)/lib64' '-L$(cat /root/.static-deps-
...
-  # - name: Build libssh
-  #   env:
-  #     CFLAGS: -I/usr/local/opt/openssl/include
-  #     LDFLAGS: -L/usr/local/opt/openssl/lib
-  #   run: |
-  #     cmake ..
-  #     make
-  #     make install/strip
-  #   working_directory: libssh/build
+  "bash {project}/packaging/macos/cibuildwheel-before-build.sh",
```
webknjaz, line 248: I'd expect the lib to be pre-built in a separate workflow and stored as an artifact. Similar to manylinux containers. The whole point is to avoid running expensive dep rebuilds in every single CI workflow run.

packaging/rpm/ansible-pylibssh.spec
webknjaz, line 1: entirely unrelated

wheelhouse/ansible_pylibssh-0.1.dev2095+gfe80be0ec.d20260305-cp312-cp312-macosx_10_13_universal2.whl
webknjaz, line 1: another leftover?

pyproject.toml
```diff
@@ -200,7 +200,7 @@ profile = "True"
     showcontent = true
 
 [tool.cibuildwheel]
-build-frontend = "build"
+build-frontend = "pip"
```
webknjaz, line 203: this doesn't seem reasonable

packaging/macos/merge-universal2.sh
webknjaz: I don't think it's robust to accidentally pick up libs that happen to end up in the dir by a wildcard, which can lead to a [TOC/TOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) vulnerability.

Moreover, it obscures the expectations which hurts maintainability.

packaging/macos/config.sh
webknjaz, line 1: We don't need yet another config with duplicate data in it — it's better to use the single source of configuration.

packaging/pep517_backend/_cython_configuration.py
komaldesai13: removed this logic

packaging/pep517_backend/_cython_configuration.py
komaldesai13: using [tool.cibuildwheel.macos.environmen] inside pyproject.yaml now

pyproject.toml
komaldesai13: addressed , changed script to use pre-built artifcats

packaging/macos/merge-universal2.sh
komaldesai13: removed loop here and added specific files

packaging/macos/build-libssh-macos.sh
komaldesai13: Created shared build-scripts/install_openssl.sh , reusing https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/build-scripts/manylinux-container-image/install_openssl.sh

pyproject.toml
komaldesai13: moved build-frontend it back to build

pyproject.toml
```diff
@@ -248,44 +248,19 @@ LDFLAGS = "'-L$(cat /root/.static-deps-path)/lib64' '-L$(cat /root/.static-deps-
...
-  #   run: |
-  #     cmake ..
-  #     make
-  #     make install/strip
-  #   working_directory: libssh/build
+  "bash {project}/build-scripts/macos/cibuildwheel-before-build.sh",
 ]
 
 [tool.cibuildwheel.macos.environment]
-# delocate (bundled with cibuildwheel v3) enforces that the wheel's
```
webknjaz, line 284: Why is all of this lost?

pyproject.toml
```diff
@@ -248,44 +248,19 @@ LDFLAGS = "'-L$(cat /root/.static-deps-path)/lib64' '-L$(cat /root/.static-deps-
...
-  # - name: Build libssh
-  #   env:
-  #     CFLAGS: -I/usr/local/opt/openssl/include
-  #     LDFLAGS: -L/usr/local/opt/openssl/lib
-  #   run: |
-  #     cmake ..
-  #     make
-  #     make install/strip
-  #   working_directory: libssh/build
+  "bash {project}/build-scripts/macos/cibuildwheel-before-build.sh",
```
webknjaz, line 254: Not sure how I feel about building the deps as a part of the wheel build process. The whole point of the pre-cached artifacts is that they are built separately and don't waste resources every time the project is being build.

.gitignore
```diff
@@ -198,10 +198,14 @@ parts/
 sdist/
 var/
 wheels/
+wheelhouse/
```
webknjaz, line 201: This part is coming from a centralized template, I wouldn't put changes in the middle of it unless the upstream got the same changes.

.gitignore
```diff
@@ -198,10 +198,14 @@ parts/
...
 pip-wheel-metadata/
 share/python-wheels/
 *.egg-info/
 .installed.cfg
+
+# macOS build artifacts
+.macos-static-deps-path*
```
webknjaz, line 208: Why would it be here?

.gitignore
```diff
@@ -210,6 +214,8 @@ MANIFEST
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
 *.spec
+# Exception: Allow RPM spec files
+!packaging/rpm/*.spec
```
webknjaz, line 218: This is definitely unrelated. Plus, it's unnecessary since the file is already being tracked.

webknjaz: I guess it may make sense to have this but not globally — scope it to the respective folder. Though, this must be a standalone change as it has absolutely nothing to do with the goal of this PR and only distracts from whatever that goal is.

.gitignore
```diff
@@ -396,3 +402,6 @@ $RECYCLE.BIN/
 
 # setuptools-scm write_to:
 /src/pylibsshext/_scm_version.py
+
+# macOS
+.DS_Store
```
webknjaz, line 407: This mustn't exist in the project. The contributors must add this to their global gitignore as this is something that is coming from the contrib system and is unrelated to the project.

build-scripts/install_openssl.sh
```diff
@@ -0,0 +1,141 @@
...
+    # Build using all available cores
+    NPROC=$(nproc 2>/dev/null || echo 4)
+fi
+
+echo "Building OpenSSL..."
+make -j${NPROC}
+
+echo "Installing OpenSSL to ${INSTALL_PREFIX}..."
+# Install only software and SSL directories (skip docs)
+make install_sw install_ssldirs
```
webknjaz, line 139: Why is this needed?

komaldesai13: @webknjaz ,
 this has reusing logic of [this](https://github.com/ansible/pylibssh/blob/3498943c86bd46043b490b004931e56faaaf4937/build-scripts/manylinux-container-image/install_openssl.sh ) script , but creating generic script for macos and manylinux , instead enhancing manylinux script

webknjaz: That was needed in the context of containers. With macOS, we probably need to think of an in-tree location for the built lib cache. I'd reuse PyCA's ideas, of course, instead of reinventing the wheel.

webknjaz: Or maybe out of the tree.. Anyway, there's separation between building the deps and using them. For using them (while building the wheels), I'd expect us to pre-download the deps like so: https://github.com/pyca/cryptography/blob/43897cbe22d304a93d6e8736fd386516baa9781d/.github/workflows/wheel-builder.yml#L212-L220. I originally expected that this PR would only contain a minimal patch for building the deps and storing them as GHA artifacts but not embed that into the wheel building process just yet.

It seems like we may want to think of a mechanism for pointing to the place where the pre-built deps live. The manylinux config just sets the respective env vars and the macOS one should too. But I was hoping to avoid growing the scope indefinitely.

build-scripts/macos/README.md
```diff
@@ -0,0 +1,145 @@
...
+     --notes "Pre-built static libraries for macOS wheel builds"
+   ```
+
+2. **Upload artifacts:**
+   ```bash
+   gh release upload libssh-v0.12.0-openssl-3.5.0 \
+     build-scripts/macos/build/arm64/libssh-*.tar.gz \
+     build-scripts/macos/build/arm64/libssh-*.tar.gz.sha256 \
+     build-scripts/macos/build/x86_64/libssh-*.tar.gz \
+     build-scripts/macos/build/x86_64/libssh-*.tar.gz.sha256
```
webknjaz, line 98: Just saw this. We should definitely not publish releases with other projects in here.

PR #797, opened 2026-02-12
Update libssh to 0.12.0

Jakuje:
##### SUMMARY
First step of updating to new feature (& security) release of libssh released this week.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
This brings in some of the fixes we discussed over last months that made it to libssh upstream.

PR #798, opened 2026-02-12
Update to libssh 0.12.0 in platform-specific wheels

Jakuje:
##### SUMMARY
Update to libssh 0.12.0 in platform-specific wheels

Follow-up from #797

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/798.contrib.rst
+++ b/docs/changelog-fragments/798.contrib.rst
@@ -1,3 +1,2 @@
 Updated the bundled version of libssh to 0.12.0 in platform-specific
 wheels published on PyPI -- by :user:`Jakuje`.
-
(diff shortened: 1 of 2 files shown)


Jakuje:
Note, that the libssh 0.12 brings in the PQC key exchange algorithms, which is something we are certainly interested in having in ansible. While configuration management is not the top priority target for harvest know & decrypt later, its something we will need for compliance sooner or later.

Jakuje:
This is now blocked on getting the pipeline in `devel` to pass so we have the right images

webknjaz:
I'll take a look tomorrow. Thanks!

webknjaz (approved):
Thanks!

PR #801, opened 2026-02-17
Update build images to include recent ubi9 and current Fedora versions

Jakuje:
##### SUMMARY
This increases test coverage with current libssh on platforms we need to support pylibssh.

##### ISSUE TYPE
 - CI/CD

### DIFF
--- a/docs/changelog-fragments/801.contrib.rst
+++ b/docs/changelog-fragments/801.contrib.rst
@@ -1,0 +1,2 @@
+Added Fedora and ubi9 images to CI/CD pipeline that are relevant in 2026
+and removed EOL Fedora versions -- by :user:`Jakuje`.
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -909,10 +909,9 @@ jobs:
     strategy:
       matrix:
         target-container:
-        - tag: fedora:39
-        - tag: fedora:40
-        - tag: fedora:41
         - tag: fedora:42
+        - tag: fedora:43
+        - tag: fedora:44
         - tag: ubi9/ubi:9.0.0
           registry: registry.access.redhat.com
         - tag: ubi9/ubi:9.1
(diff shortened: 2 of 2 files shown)


Jakuje:
Hmm. For some reason, the UBI10 does not have python3-cython, while it should be in CRB.  Not sure if it needs to be added somehow specifically in RH infra to get it working.  So I will drop the ubi10 part and just get the Fedoras updated.

Jakuje:
For ubi10 filled #802

Jakuje:
@webknjaz the ubuntu lint is broken now?

webknjaz:
That's because `setuptools` dropped `pkg_resources`. I fixed this in `awx-plugins` if you want to check out the commit. I'll fix it here later otherwise.

webknjaz:
Fixed via #803. I'll click `Rebase` in this PR for you.

docs/changelog-fragments/801.contrib.rst
```diff
@@ -0,0 +1 @@
+Add current Fedora and ubi9 images to CI/CD pipeline -- by :user:`Jakuje`.
```
webknjaz: Could you be more specific. "Current" is timestamp-dependent, after all. This will be read in the future when it won't be true anymore.

src/pylibsshext/includes/sftp.pxd
```diff
@@ -1,5 +1,5 @@
 #
-# This file is part of the pylibssh library
+# This file is part of the ansible-pylibssh library
```
webknjaz: Could you move this correction into a separate PR?

docs/changelog-fragments/801.contrib.rst
Jakuje: Done

src/pylibsshext/includes/sftp.pxd
Jakuje: Sure. In #804.

docs/changelog-fragments/801.contrib.rst
```diff
@@ -0,0 +1 @@
+Add Fedora and ubi9 images to CI/CD pipeline that are relevenat in 2026 -- by :user:`Jakuje`.
```
webknjaz: Non-imperative mood plz

webknjaz: Also "relevant".

webknjaz (approved):

PR #805, opened 2026-02-23
Add recent UBI 10 images

### LINKED ISSUE #802: Add ubi10 CI images
##### SUMMARY

It looks like the crb in ubi10 does not have python3-cython which prevents adding this version. Some WIP branch where I tried to add the rest of the stuff is here:

https://github.com/Jakuje/pylibssh/tree/ubi10 / #805

##### ISSUE TYPE
- CI/CD

##### ADDITIONAL INFORMATION

> [!caution]
>
> ##### External/downstream blocker
>
> https://redhat.atlassian.net/browse/RHEL-151535
>
> ##### Okay to abandon RHEL10 support via UBI10 in the upstream CI
>
> On Aug 12, 2026 if downstream isn't interested in us testingthis scenario.

Jakuje:
##### SUMMARY

Adds UBI10 images to CI

Fixes: #802

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

This will help with testing against more different configurations.

But it currently does not work as it looks like the ubi10 does not have python3-cython.

> [!caution]
>
> ##### External/downstream blocker
>
> https://redhat.atlassian.net/browse/RHEL-151535
>
> ##### Okay to abandon RHEL10 support via UBI10 in the upstream CI
>
> On Aug 12, 2026 if downstream isn't interested in us testingthis scenario.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -1028,7 +1028,10 @@ jobs:
         dnf-plugins-core
         rpm-build
         rpmdevtools
-        rpmlint
+
+    - name: Install rpmlint except for ubi10, which is broken
+      if: ${{ ! contains(matrix.target-container.tag, 'ubi10') }}
+      run: dnf install -y rpmlint
 
     - name: Create rpmbuild directory structure
       run: rpmdev-setuptree
@@ -1054,6 +1057,7 @@ jobs:
         merge-multiple: true
 
     - name: Lint the RPM spec file
+      if: ${{ ! contains(matrix.target-container.tag, 'ubi10') }}
       run: rpmlint packaging/rpm/"${PROJECT_NAME}".spec
 
     - name: Copy sdist to the sources dir
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -958,6 +958,12 @@ jobs:
           registry: registry.access.redhat.com
         - tag: ubi9/ubi:9.7
           registry: registry.access.redhat.com
+        - tag: ubi9/ubi:9.8
+          registry: registry.access.redhat.com
+        - tag: ubi10/ubi:10.1
+          registry: registry.access.redhat.com
+        - tag: ubi10/ubi:10.2
+          registry: registry.access.redhat.com
         - tag: centos/centos:stream9
           registry: quay.io
         - tag: centos/centos:stream10
@@ -1008,8 +1014,9 @@ jobs:
         dnf install -y
         https://dl.fedoraproject.org/pub/epel/epel{,-next}-release-latest-9.noarch.rpm
 
-    - name: Install epel for centos 10
-      if: contains(matrix.target-container.tag, 'centos:stream10')
+    - name: Install epel for centos 10 / ubi 10
+      if: contains(matrix.target-container.tag, 'centos:stream10') ||
+          contains(matrix.target-container.tag, 'ubi10')
       run: >-
         dnf install -y
         https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
@@ -1065,8 +1072,8 @@ jobs:
         python3-py
         python3-tomli
 
-    - name: Install static test dependencies missing from all UBIs
-      if: contains(matrix.target-container.tag, 'ubi')
+    - name: Install static test dependencies missing from UBI9
+      if: contains(matrix.target-container.tag, 'ubi9')
       run: >-
         rpm
         -ivh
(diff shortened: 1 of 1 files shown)


Jakuje:
For the record, I opened https://issues.redhat.com/browse/RHEL-151535 to add the python3-cython to ubi10. Its in ubi9 so I do not see a reason why it should not be in ubi10.

In the meantime, we might be able to have centos:10stream build image, but it does not have the minor versions so it would get more limited use.

webknjaz:
> Its in ubi9 so I do not see a reason why it should not be in ubi10.

I faced similar situations in the past and was told that nobody needs a dep outside the subscription :man_shrugging: Hope this time we'll get a better outcome.

webknjaz:
> I opened https://issues.redhat.com/browse/RHEL-151535

Looks like I don't have access to that. Any news?

Jakuje:
> > I opened https://issues.redhat.com/browse/RHEL-151535
> 
> Looks like I don't have access to that. Any news?

No update there. That makes me thing that we should just go ahead with the c10s

webknjaz:
> with the c10s

what's that?

Jakuje:
> > with the c10s

> what's that?

Centos 10 Stream -- basically latest RHEL Y-stream version. Nicely described on this year FOSDEM talk:

https://fosdem.org/2026/schedule/event/8WES7N-centos-mythbusters/

webknjaz:
@Jakuje could you try rebasing to see if this got fixed?

Jakuje:
I do not think so. The https://redhat.atlassian.net/browse/RHEL-151535 was not touched since I opened it. But I can rebase.

webknjaz:
I've left a few warm words in there, in the spirit of transparency… If they don't respond in like a week or so, we'll consider that env unsupported.

> [!note]
>
> N.B. I might've mentioned this in the past or not, I don't remember.
> But this entire thing of me trying to have a spec file, packit, rhel/ubi/epel, fedora, centos et al in the CI was just me being curious of how hard could it be to maintain maximum upstream compat w/ downstreams and being nice to them in general. And if they aren't interested, I don't think we should be either (beyond the contrib-required infra that we ourselves need for dev env and stuff). I've had great experiences with the downwnstream maintainers like Arch, Fedora and Gentoo, as a Python project upstream maintainer, where they would make an effort to be on the same page and have a good collab. OTOH, we've had UBI folks saying "no" to us asking to expose an RPM when I was first starting to put the `ansible-pylibssh` infra together. I think I wasn't able to properly frame the need in terms of a "business case" back then, though due to the lack of experience, perhaps this time we'll be able to show the upside of having respective build deps exposed for our upstream testing to help their downstream packaging in the long run :man_shrugging:
>
> But if not — I think we should take a serious look at the sustainability side of this  project's infra since the side of the CI matrix is sort of a side effect of my idealist worldview and does not reflect the actual resources we've got to keep up with it.

> [!tip]
>
> One possibility could be reducing the matrix and instead of having each of 9.0.0, 9.1, 9.0, 9.2, 9.3, 9.4, 9.5, 9.6 and 9.7, we could just have min+max (9.0+9.7), at least on the PR runs.

Jakuje:
I did some more fight with the following outcomes:
 * ~the problem is not missing cython package from ubi, but missing python3-cython metapackage~
 * since the last time trying the `rpmlint` became uninstallable because python3-zstandard moved from epel to rhel, but was not included in ubi. Reporeted internally.

The current version goes up to the rpm dependencies evaluation and fails on the following:
```
No matching package to install: 'python3dist(cython) >= 3'
```
These requirements are generated from the python requirements, but the missing package is breaking this. If I will have success with fixing the rmplint, I will try to nudge also the python3-cython similarly.

Edit: The problem is really missing unversioned python3-Cython package. Created a MR to add that after a bit discussion that this is the right direction. After it will be there (and repositories will be regenerated) we should unblock this MR (and drop the top commit excluding the rpmlint).

.github/workflows/ci-cd.yml
```diff
@@ -958,6 +958,12 @@ jobs:
           registry: registry.access.redhat.com
         - tag: ubi9/ubi:9.7
           registry: registry.access.redhat.com
+        - tag: ubi9/ubi:9.8
```
webknjaz, line 961: Let's not mix in unrelated stuff.

.github/workflows/ci-cd.yml
```diff
@@ -958,6 +958,12 @@ jobs:
...
+          registry: registry.access.redhat.com
+        - tag: ubi10/ubi:10.1
+          registry: registry.access.redhat.com
+        - tag: ubi10/ubi:10.2
+          registry: registry.access.redhat.com
```
webknjaz, line 966: looks like these remain broken

.github/workflows/ci-cd.yml
Jakuje: Will add this separately.

.github/workflows/ci-cd.yml
Jakuje: Yes. But there is light in the end of the tunnel

webknjaz: Dunno. Personally, I'm not invested in this anymore. UBI creates a lot of maintenance burden.

.github/workflows/ci-cd.yml
Jakuje: #876

PR #807, opened 2026-02-27
fixes scp operation for cisco iosxe

KB-perByte:
##### SUMMARY

Related changes https://github.com/ansible-collections/ansible.netcommon/pull/759/changes

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/src/pylibsshext/scp.pyx
+++ b/src/pylibsshext/scp.pyx
@@ -48,14 +48,22 @@ cdef class SCP:
         remote_dir_b, filename_b = os.path.split(remote_file_b)
         if not remote_dir_b:
             remote_dir_b = b"."
+        # Device filesystems (e.g. Cisco IOS flash:) often expect full path for scp -t
+        elif remote_dir_b.endswith(b":") and b"/" not in remote_dir_b:
+            remote_dir_b = remote_dir_b + b"/"
 
         with open(local_file, "rb") as f:
             file_stat = os.fstat(f.fileno())
             file_size = file_stat.st_size
             file_mode = file_stat.st_mode & 0o777
 
+            # For device-style paths (e.g. flash:/file), server may expect full path for scp -t
+            if b":" in remote_file_b and remote_dir_b != b".":
+                scp_location = remote_file_b
+            else:
+                scp_location = remote_dir_b
             # Create the SCP session in write mode
-            scp = libssh.ssh_scp_new(self._libssh_session, libssh.SSH_SCP_WRITE, remote_file_b)
+            scp = libssh.ssh_scp_new(self._libssh_session, libssh.SSH_SCP_WRITE, scp_location)
             if scp is NULL:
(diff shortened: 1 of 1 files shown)


webknjaz:
I imagine you'll fill out the context later and add the tests et al?

src/pylibsshext/scp.pyx
```diff
@@ -48,14 +48,22 @@ cdef class SCP:
         remote_dir_b, filename_b = os.path.split(remote_file_b)
         if not remote_dir_b:
             remote_dir_b = b"."
+        # Device filesystems (e.g. Cisco IOS flash:) often expect full path for scp -t
+        elif remote_dir_b.endswith(b":") and b"/" not in remote_dir_b:
+            remote_dir_b = remote_dir_b + b"/"
```
Jakuje, line 53: Is this an issue just for cisco or adding a trailing slash will not hurt other paths?

If this is only for Cisco, we should do that only if we are sure the remote is cisco. For example by checking the banner:

https://api.libssh.org/master/group__libssh__session.html#ga144bb84d1b1a23f3b520a42a8fc6e56a

Inside of libssh we have now a flag, but it is not accessible through the API so some similar feature/bug flag needs to be implemented also in pylibssh:

https://gitlab.com/libssh/libssh-mirror/-/blob/master/src/misc.c#L1616-1620

src/pylibsshext/scp.pyx
```diff
@@ -48,14 +48,22 @@ cdef class SCP:
...
 
         with open(local_file, "rb") as f:
             file_stat = os.fstat(f.fileno())
             file_size = file_stat.st_size
             file_mode = file_stat.st_mode & 0o777
 
+            # For device-style paths (e.g. flash:/file), server may expect full path for scp -t
+            if b":" in remote_file_b and remote_dir_b != b".":
```
Jakuje, line 61: I do not think we can easily test the device-style paths in CI, but again the `:` is a valid character for a file name in Linux so if this is a workaround only for Cisco, it would make sense to do this only if we detect cisco servers (see the other comment).

KB-perByte: Noted, this and the above comment, I spend a good amount of time figuring out what is really breaking cisco iosxe platform and could reach here. I'll be happy to discss and move the implementation to the expected correct place.

ganeshrn (commented):
` Device filesystems (e.g. Cisco IOS flash:) often expect full path for scp -t`

can't this be handled in the calling function, in this case it can he the cisco ios plugin itself. The API can be update to document the expected file path format and additional throw an error is the path is not correct.

KB-perByte:
Closing this PR as this is not needed.
The issue is handled in netcommon.

PR #808, opened 2026-03-05
📦 Bump build env pins to their last versions

### LINKED ISSUE #799: Fix (Test)PyPI rejecting sdist uploads
@Jakuje reported CI/CD failing on merge. It turned out that automatic publishing to PyPI was partly failing — it'd upload 20 wheels but fail on sdist.

I've looked at a few last releases on TestPyPI and they apparently lack sdists: https://test.pypi.org/project/ansible-pylibssh/1.3.1.dev50/#files.
The last one with an sdist was on Oct 21. This raises a suspicion that needs investigating.

The CI/CD runs have indeed started failing matching the timeline of missing sdist releases: https://github.com/ansible/pylibssh/actions/workflows/ci-cd.yml?query=branch%3Adevel+event%3Apush.

Our DumbPyPI index received an sdist so it is not getting lost in the GHA artifacts: https://ansible.github.io/pylibssh//simple/ansible-pylibssh/.

So it's likely that PyPI somehow chokes on sdists and we'll need to figure out why. We may be hitting an overall limit per-release. Though our releases contain roughly 50 MB of artifacts so this is unlikely.

Action items:
* [x] Enable verbose output in `pypi-publish`
* [x] Enable printing out the hashes in `pypi-publish`
* [x] Decide what's next
* [ ] The sdist base name needs to be normalized, it seems. This likely means having to update the `setuptools` pin

webknjaz:
This will fix #799 by making use of a newer version of `setuptools` in CI that follows modern source distribution file base name normalization rules according to newer metadata version spec.

PyPI and TestPyPI now enforces that sdists must have the metadata version of 2.2 or higher. And metadata version 2.1 and higher have name format restrictions formalized that weren't there earlier [[1]].

[1]: https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-format

##### SUMMARY

SSIA.

##### ISSUE TYPE

- Packaging Pull Request

##### ADDITIONAL INFORMATION

N/A

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -308,13 +308,17 @@ jobs:
         FILE_APPEND_MODE = 'a'
 
         whl_file_prj_base_name = '${{ env.PROJECT_NAME }}'.replace('-', '_')
-        sdist_file_prj_base_name = whl_file_prj_base_name.replace('.', '_')
+        sdist_file_prj_base_name = (
+            whl_file_prj_base_name.
+            replace('.', '_').
+            lower()
+        )
 
         with Path(environ['GITHUB_OUTPUT']).open(
                 mode=FILE_APPEND_MODE,
         ) as outputs_file:
             print(
-                "sdist=${{ env.PROJECT_NAME }}-${{
+                f"sdist={sdist_file_prj_base_name}-${{
                     steps.request-check.outputs.release-requested == 'true'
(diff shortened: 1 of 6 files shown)


webknjaz:
Looks like the macOS runner provisioning is facing a global outage: https://github.com/community/maintainers/discussions/723.

PR #719, opened 2025-05-09
🧪 Pre-build armv7l images for building wheels

webknjaz:
These are relatively recent upstream and only exist for glibc 2.31+.

##### SUMMARY

$sbj.

##### ISSUE TYPE

- Maintenance Pull Request
- Packaging Pull Request

##### ADDITIONAL INFORMATION

Blocks #648.

PR #704, opened 2025-04-23
Indicate support for Python 3.13

Qalthos:
##### SUMMARY

Also drop support for 3.8

##### ISSUE TYPE

- Feature Pull Request

PR #809, opened 2026-03-09
Add Python 3.13 and 3.14 wheel builds

### LINKED ISSUE #699: No binary wheel for py313 published on pypi

### LINKED ISSUE #781: [TODO] Fix coverage collection for Cython modules
Some time between https://app.codecov.io/github/ansible/pylibssh/commit/ccc7dc544e18654b4c6a073b46481624bc0e9aff (https://github.com/ansible/pylibssh/commit/ccc7dc544e18654b4c6a073b46481624bc0e9aff) and https://app.codecov.io/github/ansible/pylibssh/commit/3c2c36a2fb05825287637ccebca616f42b4dd5c3 (https://github.com/ansible/pylibssh/commit/3c2c36a2fb05825287637ccebca616f42b4dd5c3) Codecov forgot about the existance of Cython modules entirely.
Need to bisect that and see what happened.

cidrblock:
##### SUMMARY

Add Python 3.13 and 3.14 wheel builds to unblock downstream consumers
(e.g. `ansible.netcommon` via `ansible-pylibssh`).

**Per-file change map:**

| File | Change | Why |
|------|--------|-----|
| `setup.cfg` | Add 3.13/3.14 classifiers | Direct: declare supported versions |
| `packaging/pep517_backend/_backend.py` | Cython version bounds per Python | Direct: `Cython >= 3.1.0` for 3.14, `>= 3.0.11` for 3.13 (3.0.x has known build failures on 3.14) |
| `.github/workflows/reusable-cibuildwheel.yml` | cibuildwheel v2.17.0 → v3.3.1 | Direct: v3 is the minimum version that supports cp314 wheel building |
| `.github/workflows/ci-cd.yml` | Add 3.13/3.14 to test/build matrices | Direct: test and build wheels for new versions |
| `.github/workflows/ci-cd.yml` | Add `*t-*` to `CIBW_SKIP` in 3 jobs | Cascade (cibuildwheel v3): free-threaded builds are [no longer experimental](https://cibuildwheel.pypa.io/en/v3.1.2/changelog/#v310) and built by default; Cython extensions don't set `Py_mod_gil_not_used` |
| `.github/workflows/ci-cd.yml` | sdist name uses `f"sdist={sdist_file_prj_base_name}"` | Cascade (setuptools): PEP 625 normalization changed the sdist filename from hyphens to underscores |
| `.github/workflows/ci-cd.yml` | RPM source copy uses `env:` block | Cascade (setuptools): rename PEP 625-normalized sdist to match RPM spec's `Source0`; also moves GHA expressions out of `run:` to avoid script injection |
| `pyproject.toml` | Skip `*t-*`, `armv7l`, `riscv64`; set `MACOSX_DEPLOYMENT_TARGET` | Cascade (cibuildwheel v3): `*t-*` — see above; `armv7l`/`riscv64` — `CIBW_ARCHS_LINUX=all` in the odd-arches job now attempts these since v3 has working support, but no libssh container images exist; `MACOSX_DEPLOYMENT_TARGET=15.0` — `delocate` 0.13.0 [now errors](https://github.com/pypa/cibuildwheel/pull/1766) on deployment-target mismatch; brew-installed `openssl@3` (3.6.1) on `macos-15-intel` has a minimum target of 15.0 ([CI evidence](https://github.com/ansible/pylibssh/actions/runs/22876699147/job/66370234887)) |
| `.coveragerc` | Add `core = ctrace` | Cascade (Python 3.14): Cython coverage plugin doesn't support 3.14's `sys.monitoring` tracer; `ctrace` avoids the issue (cf. [coveragepy#2099](https://github.com/coveragepy/coveragepy/issues/2099)) |
| `requirements-build.txt` | Re-compiled lockfile | Cascade (Cython bump): updating Cython from 3.0.12 to 3.1.8 required re-running `pip-compile`; this incidentally resolved `setuptools` from 68.2.2 to 82.0.1, which introduced PEP 625 sdist normalization |
| `.packit.yml` | Hardcode `ansible_pylibssh-{version}` | Cascade (setuptools): `{upstream_pkg_name}` resolves to hyphens but the sdist directory is now underscores per PEP 625 |
| `packaging/rpm/ansible-pylibssh.spec` | Add `%{sdist_name}` macro, update `%autosetup` | Cascade (setuptools): extracted source directory now uses underscores per PEP 625 |
| `docs/changelog-fragments/809.packaging.rst` | Changelog fragment | Direct |
| `docs/changelog-fragments/809.contrib.rst` | Changelog fragment | Direct |

**Note on setuptools / PEP 625 scope:**
Re-running `pip-compile` for the Cython update incidentally resolved `setuptools` from 68.2.2 to 82.0.1 (no dependency forces this — `setuptools-scm` only requires `>= 61`). The newer setuptools introduced PEP 625 sdist filename normalization (underscores), which cascaded to `.packit.yml`, the RPM spec, and CI workflow changes. Constraining setuptools to the old version would avoid these changes but deliberately hold back standard-compliant behavior. If preferred, these can be split into a separate preparatory PR.

Fixes #781

##### ISSUE TYPE
- Feature Pull Request

##### ADDITIONAL INFORMATION

**Motivation:**
Collections depending on `ansible-pylibssh` (e.g. via `ansible.netcommon`) fail to install on Python 3.13+ because no pre-built wheels are available, and building from source requires `libssh` development headers that are typically not installed.

**Notes:**
- Custom manylinux images rebuild weekly from latest PyPA base images, which already include Python 3.13/3.14 interpreters. No Dockerfile changes needed.
- Free-threaded builds (`cp314t`) are skipped per [cibuildwheel v3.1.0 guidance](https://cibuildwheel.pypa.io/en/v3.1.2/changelog/#v310): "you have to skip it explicitly with `'cp31?t-*'` if you don't support it yet."
- macOS wheels require macOS 15.0+ (Sequoia) — `delocate` 0.13.0 enforces deployment-target consistency and the brew-installed `openssl@3` (3.6.1) bottle on `macos-15-intel` has a minimum target of 15.0 ([CI error log](https://github.com/ansible/pylibssh/actions/runs/22876699147/job/66370234887)).
- `armv7l`/`riscv64` skip entries are defensive: `CIBW_ARCHS_LINUX=all` was already set in the odd-arches job; cibuildwheel v3 has working support for these arches, so builds now fail loudly rather than silently. No container images with libssh exist for them.

**Test plan:**
- [x] CI builds cp313 and cp314 wheels successfully on Linux (manylinux)
- [x] CI builds cp313 and cp314 wheels successfully on macOS
- [x] Smoke tests pass for 3.13 and 3.14 on all platforms
- [x] Existing 3.9–3.12 wheels still build and pass tests (no regression)
- [x] sdist builds succeed across all Python versions in the matrix
- [x] RPM builds succeed on Fedora and UBI containers
- [x] Packit SRPM, rpm-build, and testing-farm checks succeed
- [x] manylinux [odd] QEMU wheel builds (aarch64, ppc64le, s390x)


requirements-build.txt
webknjaz, line 1: This is a constraint file (aka lock file), it must not contain non-pinned entries. Ever.

.packit.yml
```diff
@@ -31,7 +31,7 @@ actions:
 # - 5DE3E0509C47EA3CF04A42D34AEE18F83AFDEB23  # GH web UI commit signing key
 
 archive_root_dir_template: >-
-  {upstream_pkg_name}-{version}
+  ansible_pylibssh-{version}
```
webknjaz, line 34: This may be wrong, still looking into it.

.github/workflows/reusable-cibuildwheel.yml
```diff
@@ -108,7 +108,7 @@ jobs:
         platforms: all
 
     - name: Build wheels
-      uses: pypa/cibuildwheel@v2.17.0
+      uses: pypa/cibuildwheel@v3.3.1
```
webknjaz, line 111: This will likely explode due to a bunch of deps that need to produce compatible artifacts first.

.github/workflows/ci-cd.yml
```diff
@@ -995,7 +1004,9 @@ jobs:
       run: >-
         cp -v
         'dist/${{ needs.pre-setup.outputs.sdist-artifact-name }}'
-        ~/rpmbuild/SOURCES/
+        ~/'rpmbuild/SOURCES/${{ env.PROJECT_NAME }}-${{
```
webknjaz: Avoid GHA interpolation due to security considerations. Use bash env var expansion instead.

webknjaz (changes requested):
FYI, a part of this is being worked on in #808, the other part requires preparatory work like #795, #744 and some others. Once you're done experimenting here, make sure to send atomic PRs for every individual change being made. I'm definitely not going to accept a bunch of unrelated changes smashed together .

requirements-build.txt
cidrblock: Fixed — re-ran `pip-compile` against the pinned top-level deps (`cython==3.1.8`, `expandvars==0.9.0`, `setuptools-scm==8.1.0`, `wheel==0.37.1`, `setuptools==68.2.2`) under Python 3.9 to regenerate proper exact pins for all transitive deps:

- `packaging==26.0`
- `tomli==2.4.0`
- `typing-extensions==4.15.0`

All entries are now `==` pinned. See 8415de2.

.github/workflows/ci-cd.yml
cidrblock: Fixed — moved the `needs.pre-setup.outputs.*` expressions into an `env:` block and switched the `run:` block to use `${VARIABLE}` shell expansion. See 8415de2.

cidrblock:
> FYI, a part of this is being worked on in #808, the other part requires preparatory work like #795, #744 and some others. Once you're done experimenting here, make sure to send atomic PRs for every individual change being made. I'm definitely not going to accept a bunch of unrelated changes smashed together .

The PR summary has been updated to map changes to the addition of python3.13/3.14. There are no unrelated changes in this PR.

.packit.yml
cidrblock: The archive_root_dir_template change from {upstream_pkg_name}-{version} to ansible_pylibssh-{version} is a direct consequence of the setuptools bump to 82.0.1.
Modern setuptools (>= ~69.3) normalizes sdist filenames per PEP 625 — the tarball is now named ansible_pylibssh-{version}.tar.gz and the directory inside it is ansible_pylibssh-{version}/ (underscores). The Packit template variable {upstream_pkg_name} resolves to ansible-pylibssh (hyphens, from upstream_package_name in .packit.yml line 88), which no longer matches the actual directory that setuptools creates inside the tarball.
Changing upstream_package_name itself to underscores was considered but risks side effects in other Packit contexts (downstream package name, spec file path, etc.) where the hyphenated name is expected.
If there's a better way to handle this — e.g., a Packit template variable that applies PEP 625 normalization, or a preferred approach for this project — happy to adjust.

webknjaz:
How about following the PR template, though?

pyproject.toml
```diff
@@ -225,6 +225,8 @@ manylinux-ppc64le-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_ppc64le:libss
 manylinux-s390x-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_s390x:libssh-v0.12.0"
 manylinux-x86_64-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_x86_64:libssh-v0.12.0"
 skip = [
+  "*-manylinux_armv7l",  # no custom container image with libssh for this arch
```
webknjaz, line 233: Which is why this is a blocker.

pyproject.toml
```diff
@@ -274,6 +276,9 @@ before-build = [
   #   working_directory: libssh/build
 ]
 
+[tool.cibuildwheel.macos.environment]
+MACOSX_DEPLOYMENT_TARGET = "15.0"
```
webknjaz, line 288: This should be lower depending on the target runtime. Also, unrelated, of course.

docs/changelog-fragments/809.packaging.rst
```diff
@@ -0,0 +1,3 @@
+Added Python 3.13 and 3.14 wheel builds by bumping
+``cibuildwheel`` to v3.3.1 and ``Cython`` to 3.1.8
```
webknjaz: Missing links through RST roles, missing change notes for infra changes and packaging changes.

.github/workflows/ci-cd.yml
```diff
@@ -609,6 +609,7 @@ jobs:
           *-musllinux_*
           *-win32
           *_arm64
+          *t-*
```
webknjaz, line 614: no public notes on the need/effect/etc

.github/workflows/reusable-cibuildwheel.yml
webknjaz: Also this is a huge jump, the updates should be incremental and initially not mixed up with test infra changes.

.packit.yml
webknjaz: Alright, this is yet to be fact-checked. But not directly releated.

pytest.ini
```diff
@@ -54,6 +54,9 @@ filterwarnings =
   # Ref: https://github.com/pytest-dev/pytest-cov/issues/557
   ignore:The --rsyncdir command line argument and rsyncdirs config variable are deprecated.:DeprecationWarning
 
+  # Cython coverage plugin does not yet support Python 3.14's sys.monitoring
+  ignore:Plugin file tracers.*aren't supported with SysMonitor:coverage.exceptions.CoverageWarning
```
webknjaz: cherry-pick https://github.com/aio-libs/aiohttp/commit/eb8aefc5f8108d0d130fcaf38049511b324b4d6a.patch instead

pyproject.toml
cidrblock: Could you clarify what the specific blocker is here? The `*t-*` skip is intentional — Cython extensions don't set `Py_mod_gil_not_used`, so free-threaded imports fail with `RuntimeWarning`. Added inline comments in `pyproject.toml` and all three `CIBW_SKIP` blocks in `ci-cd.yml` to document this.

If you're referring to a prerequisite PR that needs to land first, happy to rebase once it's merged.

pyproject.toml
cidrblock: This is a direct consequence of the cibuildwheel v2→v3 upgrade: `delocate` 0.13.0 (bundled with cibuildwheel v3) enforces deployment-target consistency, and the brew-installed `libssh`/`libcrypto` on the `macos-15-intel` runner are built against macOS 15.0. Setting it lower causes `delocate` to fail with a deployment-target mismatch.

Happy to investigate building `libssh` from source with a lower target in a follow-up, but for the brew-installed path this is the minimum that works.

docs/changelog-fragments/809.packaging.rst
cidrblock: Updated — added RST cross-reference roles (`:gh:`, `:pep:`, etc.) to the packaging fragment and created a separate `809.contrib.rst` fragment for the CI/infra changes.

.github/workflows/ci-cd.yml
cidrblock: Added inline YAML comments above all three `CIBW_SKIP` blocks in `ci-cd.yml` and in `pyproject.toml` explaining that `*t-*` skips free-threaded builds because Cython extensions do not set `Py_mod_gil_not_used`, causing `RuntimeWarning` on import.

.github/workflows/reusable-cibuildwheel.yml
cidrblock: cibuildwheel v3.0.0 is the minimum version that supports `cp314` wheel building — there is no intermediate v2.x release that adds Python 3.14 support. The jump from v2 to v3 is required to achieve the goal of this PR. Incremental updates within the v2.x line wouldn't bring us closer to cp314 support.

The cascading changes (free-threaded skips, new arch skips, `delocate` deployment-target enforcement) are all direct consequences of the major version bump and would need to be addressed regardless of whether we landed at v3.0.0 or v3.3.1.

.packit.yml
cidrblock: This is a direct cascade from the `setuptools` bump (68.2.2 → 82.0.1), which is itself required by the cibuildwheel v3 upgrade. Modern `setuptools` (≥ ~69.3) normalizes sdist filenames per PEP 625 — the tarball directory is now `ansible_pylibssh-{version}` (underscores), but Packit's `{upstream_pkg_name}` resolves to `ansible-pylibssh` (hyphens). Without this fix, Packit SRPM builds fail because the extracted directory name no longer matches.

Happy to split it out if there's a way to land it independently without breaking the Packit build.

pytest.ini
cidrblock: Done — added `core = ctrace` to `.coveragerc` under `[run]` and removed the `pytest.ini` warning filter. This also resolves the coverage collection issue on Python 3.14 noted in the "Known issue" section of the PR description.

cidrblock:
Updated the PR description to follow the template structure (SUMMARY / ISSUE TYPE / ADDITIONAL INFORMATION). The detailed breakdown of the causal chain is retained under ADDITIONAL INFORMATION since the cascading dependencies between changes warrant the context.

Also pushed code changes addressing the other review comments:
- Replaced the `pytest.ini` warning filter with `core = ctrace` in `.coveragerc` (per the aiohttp approach)
- Added inline comments documenting the `*t-*` skip rationale in all three `CIBW_SKIP` blocks and in `pyproject.toml`
- Updated `809.packaging.rst` with RST cross-reference roles and created a separate `809.contrib.rst` for CI/infra changes

pyproject.toml
cidrblock: Apologies, I misread which line this was on.

The `armv7l`/`riscv64` skip entries are new but purely defensive — cibuildwheel v3 expanded its default Linux arch set to include them, so without the skips the build would attempt (and fail for) architectures the project has no container images for. They maintain the existing behavior of not building for these arches; they don't add architecture support.

Actually adding `armv7l`/`riscv64` support (container images, CI jobs, etc.) is out of scope for a Python 3.13/3.14 PR and can be handled separately.

pyproject.toml
cidrblock: On further review, the brew-installed Intel bottles for libssh are built for Sonoma (macOS 14) — there is no Sequoia Intel bottle. Lowered MACOSX_DEPLOYMENT_TARGET to 14.0 so the wheels remain compatible with macOS Sonoma users.

cidrblock: To clarify why this is related: under cibuildwheel v2, `delocate` did not enforce deployment-target consistency, so `MACOSX_DEPLOYMENT_TARGET` wasn't needed. Under cibuildwheel v3, `delocate` 0.13.0 does enforce it — without this setting, the macOS wheel build fails with a deployment-target mismatch between the wheel and the brew-installed dylibs. This is a direct cascade from the v2→v3 upgrade, not an independent change.

.github/workflows/reusable-cibuildwheel.yml
webknjaz: Which is why #719 #744 #725 #704 #648 have been difficult to deal with. There's a 200-comment thread on slack dedicated to what needs to happen with smaller steps if you're interested in helping out.

I don't see how we'll be able to get such multi-effect changes in as a big chunk. It's possible to make smaller changes and it seems that this will fall on me to do the work.

I'd see if #648 can be completed first, for example.

cidrblock:
@webknjaz — Updated the PR description with a per-file change map showing each file's change and its justification (direct vs cascade). Also added an honest note about the setuptools/PEP 625 scope: the bump from 68.2.2 to 82.0.1 was incidental from re-running `pip-compile` for the Cython update (no dependency forces it — `setuptools-scm` only requires `>= 61`). Happy to split the PEP 625 normalization changes into a separate preparatory PR if that's preferred.

Other updates since last review:
- ~`MACOSX_DEPLOYMENT_TARGET` lowered from 15.0 to 14.0 (brew Intel bottles target Sonoma, not Sequoia)~
- `core = ctrace` in `.coveragerc` replaces the `pytest.ini` warning filter (per your suggestion to follow the aiohttp approach)
- Inline comments added for `*t-*` skip rationale
- Changelog fragments updated with RST roles and split into `809.packaging.rst` + `809.contrib.rst`
- Commits squashed to a single commit

pyproject.toml
cidrblock: Correction: 14.0 doesn't work. CI confirmed that while `libssh` itself targets 14.0, the brew-installed `openssl@3` (3.6.1) bottle on `macos-15-intel` has a minimum target of **15.0**:

```
delocate.libsana.DelocationError: Library dependencies do not satisfy target MacOS version 14.0:
libcrypto.3.dylib has a minimum target of 15.0
```

[Full CI log](https://github.com/ansible/pylibssh/actions/runs/22876699147/job/66370234887)

Reverted to `MACOSX_DEPLOYMENT_TARGET=15.0`. To support older macOS, `libssh` and `openssl` would need to be built from source with a lower deployment target rather than using brew bottles.

cidrblock: Back to 15, explanation with links in the PR summary.

docs/changelog-fragments/809.contrib.rst
```diff
@@ -0,0 +1,11 @@
+Updated CI infrastructure for Python 3.13 and 3.14 support
+-- by :user:`cidrblock`.
+
+Bumped :gh:`pypa/cibuildwheel` from v2.17.0 to v3.3.1, ``setuptools``
```
webknjaz: There's a `:pypi:` role for published projects.

docs/changelog-fragments/809.contrib.rst
```diff
@@ -0,0 +1,11 @@
...
+from 68.2.2 to 82.0.1, and re-compiled ``requirements-build.txt`` to
+resolve transitive pin conflicts. Adapted the CI workflow and Packit
+config for :pep:`625` sdist filename normalization (underscores) caused
+by the ``setuptools`` upgrade. Configured ``core = ctrace`` in
+``.coveragerc`` to work around the Cython coverage plugin not supporting
```
webknjaz: `:file:`

docs/changelog-fragments/809.contrib.rst
```diff
@@ -0,0 +1,11 @@
...
+Python 3.14's ``sys.monitoring`` tracer
```
webknjaz: This is linkable through intersphinx too.

docs/changelog-fragments/809.contrib.rst
cidrblock: Done — added `:pypi:` to extlinks in `docs/conf.py` and updated both changelog fragments to use `:pypi:` for `cibuildwheel`, `Cython`, `setuptools`, and `delocate`. GitHub issue refs (e.g. `coveragepy/coveragepy#2099`) kept as `:gh:`.

docs/changelog-fragments/809.contrib.rst
cidrblock: Fixed — `.coveragerc` now uses `:file:` and `requirements-build.txt` as well.

docs/changelog-fragments/809.contrib.rst
cidrblock: Fixed — `sys.monitoring` now uses `:mod:` for intersphinx linking.

.packit.yml
cidrblock: Closing as addressed. Hardcoded `ansible_pylibssh-{version}` to match PEP 625 normalization from the setuptools upgrade. Explanation and cascade documented in the per-file change map in the PR description.

.github/workflows/reusable-cibuildwheel.yml
cidrblock: Closing as addressed. Acknowledged the prerequisite work tracked in #648, #704, #719, #725, #744. The cibuildwheel v2→v3 bump is required for cp314 support — v3.0.0 is the minimum version. All cascading changes are documented in the per-file change map in the PR description.

pyproject.toml
cidrblock: Closing as addressed. The armv7l/riscv64 skip entries are defensive — cibuildwheel v3 expanded its default arch set, so without the skips the build fails for architectures with no container images. Not adding new architecture support. Documented in inline code comments and the PR description.

pyproject.toml
cidrblock: Closing as addressed. CI confirmed 15.0 is required — the brew-installed `openssl@3` (3.6.1) `libcrypto.3.dylib` has a minimum target of 15.0. Reverted to `MACOSX_DEPLOYMENT_TARGET=15.0` with inline comment explaining the `delocate` constraint. CI evidence linked in the PR description.

docs/changelog-fragments/809.packaging.rst
cidrblock: Closing as addressed. Added RST roles (:pypi:, :pep:, :file:, :mod:) and split into separate 809.packaging.rst and 809.contrib.rst fragments. Added :pypi: extlink to docs/conf.py.

.github/workflows/ci-cd.yml
cidrblock: Closing as addressed. Added YAML comments above all three CIBW_SKIP blocks documenting the *t-* free-threaded build skip rationale, and inline comments in pyproject.toml.

pytest.ini
cidrblock: Closing as addressed. Replaced the pytest.ini warning filter with core = ctrace in .coveragerc per the aiohttp approach. This also resolves coverage collection on Python 3.14.

docs/changelog-fragments/809.contrib.rst
cidrblock: Closing as addressed. Added :pypi: extlink to docs/conf.py and updated both changelog fragments to use :pypi: for published packages.

docs/changelog-fragments/809.contrib.rst
cidrblock: Closing as addressed. Changed .coveragerc and requirements-build.txt to use :file: role.

docs/changelog-fragments/809.contrib.rst
cidrblock: Closing as addressed. Changed sys.monitoring to use :mod: for intersphinx linking to the Python docs.

.github/workflows/ci-cd.yml
```diff
@@ -590,6 +590,8 @@ jobs:
         ${{ needs.pre-setup.outputs.dists-artifact-name }}
       # CIBW_CONFIG_SETTINGS: `with-cython-tracing` — for coverage collection
       # CIBW_ARCHS_MACOS=all x86_64 arm64 universal2
+      # NOTE: *t-* skips free-threaded builds — Cython extensions do not
+      # NOTE: set Py_mod_gil_not_used, causing RuntimeWarning on import.
```
webknjaz, line 594: This needs an issue, I think + a link there.

.github/workflows/ci-cd.yml
```diff
@@ -752,7 +763,13 @@ jobs:
         store-sdist-to-artifact:
         - false
         include:
-        - python-version: 3.12
+        - python-version: "3.14"
```
webknjaz, line 766: Keep the entries unquoted following the existing preferred style.

packaging/rpm/ansible-pylibssh.spec
webknjaz, line 1: I worked out a proper solution for this in #808 so I'll go with that (will split it). But will note that it's also solved similarly here.

.packit.yml
```diff
@@ -20,7 +20,7 @@ actions:
     --sourcedir packaging/rpm/ansible-pylibssh.spec
     '
   - tox -e build-dists -qq -- --sdist
-  - sh -c 'echo dist/"${PACKIT_PROJECT_NAME_VERSION}".tar.gz'
+  - sh -c 'echo dist/ansible_pylibssh-"${PACKIT_PROJECT_VERSION}".tar.gz'
```
webknjaz, line 23: I worked out a correct solution for this in #808. The right thing to do here is updating the upstream package name, after all. I haven't found any evidence of this breaking other contexts as you implied earlier (provided that it's accompanied with a change to the spec that's needed anyway).

docs/changelog-fragments/809.contrib.rst
```diff
@@ -0,0 +1,11 @@
+Updated CI infrastructure for Python 3.13 and 3.14 support
+-- by :user:`cidrblock`.
+
+Bumped :pypi:`cibuildwheel` from v2.17.0 to v3.3.1, :pypi:`setuptools`
+from 68.2.2 to 82.0.1, and re-compiled :file:`requirements-build.txt` to
+resolve transitive pin conflicts. Adapted the CI workflow and Packit
+config for :pep:`625` sdist filename normalization (underscores) caused
+by the ``setuptools`` upgrade. Configured ``core = ctrace`` in
```
webknjaz, line 8:
```suggestion
by the :pypi:`!setuptools` upgrade. Configured ``core = ctrace`` in
```

docs/conf.py
```diff
@@ -143,6 +143,7 @@
     'pr': (f'{github_repo_url}/pull/%s', 'PR #%s'),
     'commit': (f'{github_repo_url}/commit/%s', '%s'),
     'gh': (f'{github_url}/%s', 'GitHub: %s'),
+    'pypi': ('https://pypi.org/project/%s', '%s'),
```
webknjaz, line 146: :thinking: I didn't realize `sphinx-issues` isn't yet integrated here. Will have to add it later.

packaging/pep517_backend/_backend.py
```diff
@@ -335,7 +335,8 @@ def get_requires_for_build_wheel(
 
     """
     c_ext_build_deps = [
-        'Cython >= 3.0.11; python_version >= "3.13"',  # Ideally >= 3.1.0
+        'Cython >= 3.1.0; python_version >= "3.14" and python_version < "3.15"',
+        'Cython >= 3.0.11; python_version >= "3.13" and python_version < "3.14"',
```
webknjaz, line 339: You removed the comment but didn't address it. This should be updated to `>= 3.1.0`. IIRC, there was no stable release (only betas) at the time of the last change (and maybe some UBI incompatibility). But I don't see why this needs to be kept separate from the line above. AFAIR, that higher version is the one fixing a few bugs w/ Python 3.13 and/or FT. So we definitely want it as long as it doesn't cause issues.

pyproject.toml
```diff
@@ -204,7 +204,12 @@ build-frontend = "build"
 # container-engine = "podman"  # FIXME?
 test-requires = "pytest pytest-cov pytest-xdist"
 test-command = "python -Im pytest -m smoke --no-cov {project}/tests"
-skip = "pp*"
+skip = [
+  "pp*",  # FIXME: we don't ship these currently but could
```
webknjaz, line 208: Not so sure about this:
* In my other projects with C-extensions, there's pure-python implementations of the same APIs as those provided by the extensions and so they are able to ship `any`-tagged wheels with just that as a fallback. This is a powerful technique but won't work here as the entire point of this library is being a wrapper around `libssh`. We probably don't have to maintain (or depend on) a pure-python implemnentation.
* PyPy's kinda being sunset, AFAIK — it's being phased out from a bunch of upstream projects I know of, and even the core devs say that it's provided a lot of insight into JIT over the years but with CPython getting its own JIT nobody's really working on PyPy. It's unlikely that past Python 3.11 compat there will be a new stream of PyPy.

So the comment should reflect that it'll never be supported.

pyproject.toml
```diff
@@ -204,7 +204,12 @@ build-frontend = "build"
...
+  # Cython extensions do not set Py_mod_gil_not_used; free-threaded builds
+  # fail with RuntimeWarning on import:
```
webknjaz, line 210: I'd phrase it along the lines of "supporting FT requires a dedicated effort and it hasn't yet been enabled nor tested in our Cython infra). Keep the GIL notes, of course, but perhaps clarify what they mean — does this relate to our config of Cython or just using a Cython version that's pre-FT?

pyproject.toml
```diff
@@ -225,6 +230,8 @@ manylinux-ppc64le-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_ppc64le:libss
 manylinux-s390x-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_s390x:libssh-v0.12.0"
 manylinux-x86_64-image = "ghcr.io/ansible/pylibssh-manylinux_2_28_x86_64:libssh-v0.12.0"
 skip = [
+  "*-manylinux_armv7l",  # no custom container image with libssh for this arch
```
webknjaz, line 233: Just realized that this one is actually untrue. There's an image for this.

pyproject.toml
```diff
@@ -274,6 +281,12 @@ before-build = [
   #   working_directory: libssh/build
 ]
 
+[tool.cibuildwheel.macos.environment]
+# delocate (bundled with cibuildwheel v3) enforces that the wheel's
+# deployment target matches the bundled dylibs. The brew-installed
```
webknjaz, line 286: This probably nees to explain that Homebrew no longer supports selecting a user-defined target, which is the underlying reason for why this is so difficult. The actual target needs to be different depending on the platform version.

requirements-build.txt
webknjaz, line 1: This will be disregarded in favor of a separate bump. Looks like you've also used a different `pip-tools` version that changed the header.

docs/conf.py
webknjaz, line 1: This is something I'd mention in another contrib note. Though, I guess I'll add `sphinx-issues` instead.

webknjaz:
Looks like there's no atomic commits even. This will make handling the patch more difficult.

.github/workflows/ci-cd.yml
cidrblock: Noted. The maintainer has indicated he will arrive at 3.13/3.14 support through his own sequencing. A tracking issue for free-threaded build support can be raised when that work encounters this specific constraint.

.github/workflows/ci-cd.yml
cidrblock: Acknowledged. The maintainer has chosen to arrive at 3.13/3.14 support through a different sequence — quoting style can be matched to the existing convention in that work.

packaging/rpm/ansible-pylibssh.spec
cidrblock: Acknowledged. The PEP 625 normalization was solved independently in both PRs. #808 has landed with an alternate approach.

.packit.yml
cidrblock: Acknowledged. #808 has landed with `upstream_package_name` updated directly. The concern about side effects in other Packit contexts was precautionary — glad it was confirmed safe.

docs/changelog-fragments/809.contrib.rst
cidrblock: Noted. This syntax is specific to the custom extlink added here as a workaround — per your other comment, `sphinx-issues` will replace it.

docs/conf.py
cidrblock: The `:pypi:` extlink was added here to satisfy the review request for RST roles in the changelog fragments. `sphinx-issues` would be a cleaner long-term solution.

packaging/pep517_backend/_backend.py
cidrblock: The 3.13 constraint was intentionally kept at `>= 3.0.11` — the minimum working version — to avoid an unnecessary bump. Consolidating both to `>= 3.1.0` is a reasonable simplification, but was deferred here in favor of the incremental approach requested elsewhere in this review.

pyproject.toml
cidrblock: The `FIXME` comment was carried forward unchanged from the existing codebase. Changing it to a policy statement about PyPy support would have been scope creep on a Python 3.13/3.14 PR.

pyproject.toml
cidrblock: The comment accurately describes the technical constraint — `Py_mod_gil_not_used` is not set by Cython extensions. The suggested rephrasing can be addressed in the maintainer's future PR for this work.

pyproject.toml
cidrblock: Acknowledged — the comment wording was inaccurate. The skip itself is still appropriate for this PR's scope; enabling armv7l builds would be scope creep. The comment can be corrected when armv7l support is added.

pyproject.toml
cidrblock: The inline comment already explains the `delocate` enforcement and the `openssl@3` bottle target. The broader Homebrew limitation around user-defined deployment targets is useful context and can be expanded in the maintainer's follow-up work.

requirements-build.txt
cidrblock: Acknowledged. #808 has landed with its own lockfile bump. The header change was from `pip-compile` output — not an intentional modification.

docs/conf.py
cidrblock: Understood. The extlink was a stopgap; `sphinx-issues` is the better path.

cidrblock:
All review comments have been addressed or acknowledged. The maintainer has indicated he will arrive at Python 3.13/3.14 support through his own sequencing, with portions of this work already absorbed into #808. Closing this PR to avoid further conflict with that approach.

The technical findings from this PR remain available as reference:
- cibuildwheel v3.3.1 is required for cp314 wheel building
- `*t-*` free-threaded builds must be skipped until Cython sets `Py_mod_gil_not_used`
- `MACOSX_DEPLOYMENT_TARGET=15.0` is required due to `delocate` 0.13.0 enforcement against brew-installed `openssl@3`
- `core = ctrace` in `.coveragerc` resolves Python 3.14 coverage compatibility
- Cython `>= 3.1.0` is required for Python 3.14; `>= 3.0.11` for Python 3.13

Thank you for the detailed review.

docs/conf.py
webknjaz: xref #818

pyproject.toml
webknjaz, line 208: In #821, I discovered that having this triggers a warning in `cibuildwheel` v3+ which has a flipped default for PyPy and has been overlooked in this patch.

.github/workflows/ci-cd.yml
```diff
@@ -752,7 +763,13 @@ jobs:
...
+          runner-vm-os: ubuntu-22.04
+          store-sdist-to-artifact: false
+        - python-version: "3.13"
+          runner-vm-os: ubuntu-22.04
+          store-sdist-to-artifact: false
+        - python-version: "3.12"
           runner-vm-os: ubuntu-22.04
           store-sdist-to-artifact: true
```
webknjaz, line 774: FTR, this should've gone into the main matrix as 2 lines as they are smoke tests and all of the factors are already set up there and don't need repeating. Only the highest version (3.14) would be configured differently to save the artifact for further use in the following CI jobs.

webknjaz: Even better — it'd be enough to just update 3.12 to 3.14 here as the oldest supported Python versions are the smoke test and there's really no need to test everything in between sdist-wise. This was added mainly as a regression guardrails, IIRC, and was just a single job at some point until something broke.

packaging/pep517_backend/_backend.py
```diff
@@ -335,7 +335,8 @@ def get_requires_for_build_wheel(
 
     """
     c_ext_build_deps = [
-        'Cython >= 3.0.11; python_version >= "3.13"',  # Ideally >= 3.1.0
+        'Cython >= 3.1.0; python_version >= "3.14" and python_version < "3.15"',
```
webknjaz, line 338: Leaving the crusial dep missing under Python 3.15 is a practice highly discouraged by PyPA. There must be no upper bound in such cases — it's quite a cruel thing to do. Otherwise, it's  a straight way towards backtracking and/or breaking future runtime users ruthlessly.

docs/changelog-fragments/809.packaging.rst
```diff
@@ -0,0 +1,12 @@
+Added Python 3.13 and 3.14 wheel builds
+-- by :user:`cidrblock`.
+
+Bumped :pypi:`cibuildwheel` from v2.17.0 to v3.3.1 (required for
+``cp313``/``cp314`` wheel building) and :pypi:`Cython` from 3.0.12
+to 3.1.8 (Cython 3.0.x has known build failures with Python 3.14).
```
webknjaz, line 6: No proof of this. I confirmed that Cython 3.1 is needed for free-threading support. With FT disabled, there's no justification for setting a bar higher than it needs to be.

PR #812, opened 2026-03-10
[DNM] Leftover experiments w/ PEP 621 / 639

webknjaz:
The commits are yet to be moved to own PRs.


webknjaz:
:bulb: Current status: downstream support is poor; but it's possible to build a converter/transformer/transpiler that would work when building an sdist (in the in-tree build backend) and generate a *very minimal* compat `setup.cfg` + replace a few things in `pyproject.toml`'s `[build-system]` requires (or maybe dynamically downgrade deps in the in-tree build backend). This way, the upstream should be able to use all the fancy bells and whistles while letting downstreams work with what's available.

PR #830, opened 2026-03-20
Add Centos 10 Stream build

Jakuje:
##### SUMMARY
This is alternative to #805, which attempted to add ubi10, but failed on missing python-cython from the ubi images.

They should be though in the centos 10 stream.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -954,6 +954,8 @@ jobs:
           registry: registry.access.redhat.com
         - tag: ubi9/ubi:9.7
           registry: registry.access.redhat.com
+        - tag: centos/centos:stream10
+          registry: quay.io
 
     runs-on: ubuntu-latest
 
@@ -988,6 +990,13 @@ jobs:
       run: |
         dist_tag=$(rpm --eval '%{?dist}')
         echo "dist-tag=${dist_tag}" >> "${GITHUB_OUTPUT}"
+
+    - name: Install epel and enable CRB repository for centos
+      if: contains(matrix.target-container.tag, 'centos')
+      run: >-
+        dnf config-manager --set-enabled crb &&
+        dnf install -y
+        https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
 
     - name: Install build tooling
       run: >-
--- a/docs/changelog-fragments/830.contrib.rst
+++ b/docs/changelog-fragments/830.contrib.rst
@@ -1,0 +1,1 @@
+Added Centos 10 Stream image to the CI/CD pipeline -- by :user:`Jakuje`.


.github/workflows/ci-cd.yml
```diff
@@ -1039,6 +1048,7 @@ jobs:
         python3-pluggy
         python3-py
         python3-tomli
+        python3-cython
```
webknjaz: @Jakuje this looks redundant — this step is skipped for centos so I don't see how it'd belong in the PR.
```suggestion
```

webknjaz:
Other than the redundant UBI dep step, I've been thinking whether it's possible to replace all UBI9s w/ CentOS too (in a separate PR, of course) — if we can pull it off, the CI definition could be simplified.

webknjaz:
@Jakuje also, would you mind setting up commit signing (GPG or SSH keys) in your Git config? We've got a mandate to have it across the @ansible org. I've asked for an exception for now but this may block future PRs so it's best to solve this earlier if possible.

.github/workflows/ci-cd.yml
Jakuje: Thanks! Will remove. This is the leftover from the ubi10 attempt

Jakuje:
> @Jakuje also, would you mind setting up commit signing (GPG or SSH keys) in your Git config? We've got a mandate to have it across the @ansible org. I've asked for an exception for now but this may block future PRs so it's best to solve this earlier if possible.

Are there some specific requirements for that Key type, strength, storage? I would love to set something up (again) but using ssh keys for that does not feel right and working with OpenPGP is quite pain, especially when I want to have it on dedicated hardware.

Jakuje:
Seems like the centos 9 stream does not really need anything special either and adding it is as straightforward as this:

https://github.com/Jakuje/pylibssh/commit/1fd0714ed1cf085408690f8e5e2058cec7c2d7ac

So I can push it either here or into separate follow-up PR, whatever you prefer

webknjaz:
> Are there some specific requirements for that Key type, strength, storage? I would love to set something up (again) but using ssh keys for that does not feel right and working with OpenPGP is quite pain, especially when I want to have it on dedicated hardware.

I'm not aware of any. GitHub's branch protection rules only check for "is signed by a key you have verified in your GH account", IIUC.



> Seems like the centos 9 stream does not really need anything special either and adding it is as straightforward as this:
> 
> [Jakuje@1fd0714](https://github.com/Jakuje/pylibssh/commit/1fd0714ed1cf085408690f8e5e2058cec7c2d7ac)
> 
> So I can push it either here or into separate follow-up PR, whatever you prefer

A standalone PR is better. I was thinking more along the lines of removing UBI9 in favor of CentOS to reduce maintenance overhead.

webknjaz (approved):

.github/workflows/ci-cd.yml
webknjaz: Marking as resolved.

PR #831, opened 2026-03-20
Add CentOS 9 Stream to CI too

Jakuje:
##### SUMMARY
Follow-up from #830 with signed commit.

I still find a value in the ubi9 build images, but I am ok to drop them for simplifying the infrastructure.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -954,6 +954,8 @@ jobs:
           registry: registry.access.redhat.com
         - tag: ubi9/ubi:9.7
           registry: registry.access.redhat.com
+        - tag: centos/centos:stream9
+          registry: quay.io
         - tag: centos/centos:stream10
           registry: quay.io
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -991,10 +993,20 @@ jobs:
         dist_tag=$(rpm --eval '%{?dist}')
         echo "dist-tag=${dist_tag}" >> "${GITHUB_OUTPUT}"
 
-    - name: Install epel and enable CRB repository for centos
+    - name: Install enable CRB repository for centos
```
webknjaz: "Install enable"?

Jakuje: Fixed he wording.

docs/changelog-fragments/830.contrib.rst
```diff
@@ -1 +1 @@
-Added Centos 10 Stream image to the CI/CD pipeline -- by :user:`Jakuje`.
+Added Centos 9 Stream image to the CI/CD pipeline -- by :user:`Jakuje`.
```
webknjaz: Did you edit this by mistake?

Jakuje: Indeed! I should focus more to avoid committing random changes. Thanks for double-checking that! Fixed now.

Jakuje:
And there is also my ssh key signature attached to the commit.

docs/changelog-fragments/830.contrib.rst
webknjaz: No problem :)

webknjaz:
> I still find a value in the ubi9 build images, but I am ok to drop them for simplifying the infrastructure.

Well, if *you* need it, I'm not going to push for removal. It'd simplify the infra a bit but I think I could start with a few other things that would be good improvements in the general CI structure.

webknjaz (approved):
tnx

Jakuje:
> Well, if you need it, I'm not going to push for removal. It'd simplify the infra a bit but I think I could start with a few other things that would be good improvements in the general CI structure.

I really do not need it. If the CI could be simplified and then the efforts of maintaining it used elsewhere, I think its worth dropping it.

PR #848, opened 2026-07-23
packit: Correctly find generated tarball

Jakuje:
##### SUMMARY
The checkout in CI should be clean so this should be the only file in the dist/ directory.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION

the logs from recent runs
```
Before this change, the packit was failing to find the tarball, due to different setuptools_scm versions at different stages (see the built generated 7 characters hash, while the packit expected 9 characters git hash:

    Successfully built ansible_pylibssh-1.4.1.dev8+g2a52685.tar.gz
    Command: sh -c echo dist/"${PACKIT_PROJECT_NAME_VERSION}".tar.gz
    dist/ansible_pylibssh-1.4.1.dev8+g2a5268548.tar.gz

```

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -24,8 +24,9 @@ actions:
   get-current-version:
   - >-
     bash -c "
-      python3 -m setuptools_scm  | sed 's/.*Guessed Version\s\+//g'
-    "
+      python3 -Im setuptools_scm \
+        | sed 's/.*Guessed Version\s\+//g' \
+        | sed -E 's/\+g([0-9a-f]{7})[0-9a-f]*/+g\1/'"
 
 # allowed_gpg_keys:
 # - 5DE3E0509C47EA3CF04A42D34AEE18F83AFDEB23  # GH web UI commit signing key
--- a/docs/changelog-fragments/848.contrib.rst
+++ b/docs/changelog-fragments/848.contrib.rst
@@ -1,0 +1,2 @@
+Fixed Packit CI caused by ``setuptools_scm`` version mismatch
+in build environment -- by :user:`Jakuje`.


.packit.yml
```diff
@@ -20,7 +20,7 @@ actions:
     --sourcedir packaging/rpm/ansible-pylibssh.spec
     '
   - tox -e build-dists -qq -- --sdist
-  - sh -c 'echo dist/"${PACKIT_PROJECT_NAME_VERSION}".tar.gz'
+  - sh -c 'ls dist/*.tar.gz'
```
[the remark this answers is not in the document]
Jakuje: The checkout is clean and the pipeline should generate just one file. I believe this is ok assumption. If there will be more files, it will fail somewhere down the pipeline.

Jakuje:
We need to drop the `testing-farm:fedora-42-x86_64` from the required jobs otherwise we will not be able to make the CI pass.

komaldesai13 (approved):

.packit.yml
webknjaz: I don't like that this solution is too permissive. The previous one was there as a sanity check for the contract between the components and removing it will just allow for the components to drift apart more.
This is probably caused by different versions of setuptools-scm under tox and externally or something like that. We could try cutting a few chars off and then having a wildcard. Or at least using the name in the match.

@Jakuje have you seen the most recent version of the spec file?

Jakuje: I can try to do some manual modification of the names, but I feel like it will just explode next time when something gets updated somewhere. But I can do that.

webknjaz: Let's at least try to keep a part of the original guard in place, while allowing some flexibility in the future local version fragment surprises.

Jakuje: Adjusted. Please take a look

.github/workflows/ci-cd.yml
```diff
@@ -935,7 +935,6 @@ jobs:
     strategy:
       matrix:
         target-container:
-        - tag: fedora:42
```
webknjaz, line 938: Could you send this in separately? Or re-frame the PR and the change note + commits?

.packit.yml
```diff
@@ -24,8 +24,9 @@ actions:
   get-current-version:
   - >-
     bash -c "
-      python3 -m setuptools_scm  | sed 's/.*Guessed Version\s\+//g'
-    "
+      python3 -m setuptools_scm \
```
webknjaz: There's nothing to escape because of that `>-` in YAML (you're actually trying to escape a whitespace which you probably shouldn't).
```suggestion
      python3 -Im setuptools_scm
```

.github/workflows/ci-cd.yml
Jakuje: moved to #852

.packit.yml
Jakuje: Good point. Removed the escaping

Jakuje: Hmm. Without escaping the CI fails:
```
2026-07-28 16:16:23.032 logging.py        INFO   bash: -c: line 3: syntax error near unexpected token `|'
2026-07-28 16:16:23.033 logging.py        INFO   bash: -c: line 3: `    | sed 's/.*Guessed Version\s\+//g''
2026-07-28 16:16:23.033 commands.py       ERROR  Command "bash -c \n  python3 -m setuptools_scm\n    | sed 's/.*Guessed Version\\s\\+//g'\n    | sed -E 's/\\+g([0-9a-f]{7})[0-9a-f]*/+g\\1/'" failed.
2026-07-28 16:16:23.033 commands.py       DEBUG  Command stderr: bash: -c: line 3: syntax error near unexpected token `|'
```
Regardless of the `-I`.

Jakuje: I guess the escaping is still needed for the bash then.

webknjaz: Ah, that's because bash receives a multi-line script, then.

webknjaz (approved):

PR #860, opened 2026-07-30
Revert "Bump redhat-actions/buildah-build from 2.13 to 3"

webknjaz:
Reverts ansible/pylibssh#853 in hopes to recover the CI/CD health.

### DIFF
--- a/.github/workflows/build-manylinux-container-images.yml
+++ b/.github/workflows/build-manylinux-container-images.yml
@@ -99,7 +99,7 @@ jobs:
         --install all
     - name: Build the image with Buildah
       id: build-image
-      uses: redhat-actions/buildah-build@v3
+      uses: redhat-actions/buildah-build@v2.13
       with:
         arch: ${{ env.QEMU_ARCH }}
         image: ${{ env.FULL_IMAGE_NAME }}


webknjaz:
@Jakuje you know what..? I seem to recall the same container getting updates the last time:
> ```
> Trying to pull docker.io/tonistiigi/binfmt:latest...
> ...
> Error: OCI runtime error: crun: unknown version specified
> Error: Process completed with exit code 126.
> ```

https://github.com/ansible/pylibssh/actions/runs/30573194011/job/90974868335?pr=860#step:3:20

webknjaz:
Found a few mentions of possible GHA Ubuntu VM updates or golang toolchain mismatch or something being "flaky" there over the past day. But also, the step that's failing is before the invocation of the buildah-build action — it's in the podman call. Will have to look deeper tomorrow.

Jakuje:
Checking with Copilot suggests that we should use oci:false with the `redhat-actions/buildah-build`, but I did not investigate it further. Sounds like its not related to the version of the action as we can see here.

webknjaz:
Yeah, in the past the fix was moving to from the `multiarch/qemu-user-static` image to `tonistiigi/binfmt`: [`9ccbb8a` (#713)](https://github.com/ansible/pylibssh/pull/713/changes/9ccbb8a0b31890ef72d715da03564b2d0774cb9c) which seems to have also been prompted by a change in the GHA-provided `ubuntu-24.04` VM: https://github.com/actions/runner-images/issues/11662. It required a QEMU bump with a fix in the container image we were using but since it was no longer maintained, we had to move to another image.

So I'm curious if we now have to wait for a new release in https://github.com/tonistiigi/binfmt.

webknjaz:
https://github.com/podman-container-tools/podman/discussions/22068#discussioncomment-9145858 seems to suggest that the version of `crun` is outdated (mismatched) somewhere (maybe in the container image, maybe on the host?)

I also seem to recall seeing something like this in ansible-core-ci (which is running Fedora VMs among other things on AZP). I'll try searching those logs and some Slack convos for more clues..

webknjaz:
@Jakuje so they've deployed a bad image and rolled it back: https://github.com/actions/runner-images/issues/14473#issuecomment-5141000369. We're good for now. They'll probably take this into account in the next VM update and… not break it, I hope.

---

For history, here's one workaround posted in there:
```
OCI_RUNTIME=$(command -v crun)
podman --runtime "$OCI_RUNTIME" run ...
```
*(https://github.com/actions/runner-images/issues/14473#issuecomment-5124639423)*

Another workaround was to point podman to a matching `crun` binary when two are installed but the wrong (old) one was being picked up due to being priorities via PATH lookup, it seems: https://github.com/ansible/molecule/pull/4672.patch.

Not sure which one I like more, but it's good to have these for future reference since this isn't the first time we're catching a bug of the sort.

PR #862, opened 2026-08-03
sftp: Various improvements in the synchronous SFTP

Jakuje:
##### SUMMARY

 * Improve doc texts
 * Reformat long lines
 * Avoid memory leaks by usimg try/finally blocks
 * Avoid leaking memory from SFTP attributes

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/862.bugfix.rst
+++ b/docs/changelog-fragments/862.bugfix.rst
@@ -1,0 +1,8 @@
+Fixed several potential issues with the SFTP implementation:
+
+ * Fixed corner case where large write was truncated by libssh to match
+   server requested limits.
+ * Handled corner case where short write to local file would not write
+   all requested data.
+
+-- by :user:`Jakuje`
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -71,6 +71,8 @@ cdef class SFTP:
         """
         cdef sftp.sftp_file rf
         cdef const char* c_buf
+        cdef size_t offset
+
         with open(local_file, "rb") as f:
             remote_file_b = remote_file
             if isinstance(remote_file_b, unicode):
@@ -91,12 +93,13 @@ cdef class SFTP:
                     ),
                 )
             read_buffer = f.read(SFTP_MAX_CHUNK)
+            offset = 0
 
-            while read_buffer != b"":
+            while len(read_buffer) - offset > 0:
                 c_buf = read_buffer
-                length = len(read_buffer)
-                written = sftp.sftp_write(rf, c_buf, length)
-                if written != length:
+                length = len(read_buffer) - offset
+                written = sftp.sftp_write(rf, c_buf + offset, length)
+                if written <= 0:
                     sftp.sftp_close(rf)
                     raise LibsshSFTPException(
                         "Writing to remote file [%s] failed with error [%s]"
@@ -105,7 +108,12 @@ cdef class SFTP:
                             self._get_sftp_error_str(),
                         )
                     )
+                elif written < length:
+                    offset += written
+                    continue
                 read_buffer = f.read(SFTP_MAX_CHUNK)
+                offset = 0
+
             sftp.sftp_close(rf)
 
     def get(self, remote_file, local_file):
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -176,21 +176,19 @@ cdef class SFTP:
                             ),
                         )
 
-                    bytes_written = f.write(read_buffer[:file_data])
-                    if bytes_written and file_data != bytes_written:
-                        sftp.sftp_close(rf)
-                        raise LibsshSFTPException(
-                            "Number of bytes [%s] read from remote file [%s]"
-                            " does not match number of bytes [%s] written to"
-                            " local file [%s] due to error [%s]"
-                            % (
-                                file_data,
-                                remote_file,
-                                bytes_written,
-                                local_file,
-                                self._get_sftp_error_str(),
-                            ),
-                        )
+                    bytes_written = 0
+                    while bytes_written < file_data:
+                        chunk_written = f.write(read_buffer[bytes_written:file_data])
+                        if chunk_written == 0:
+                            sftp.sftp_close(rf)
+                            raise LibsshSFTPException(
+                                "Unable to write [%s] bytes to local file [%s]"
+                                % (
+                                    file_data,
+                                    local_file,
+                                ),
+                            )
+                        bytes_written += chunk_written
         finally:
             if read_buffer is not NULL:
                 PyMem_Free(read_buffer)
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -5,10 +5,13 @@ import uuid
 
 import pytest
 
-from pylibsshext.sftp import SFTP_MAX_CHUNK
+from pylibsshext.errors import LibsshSFTPException
+
+from pylibsshext import __libssh_version__, sftp
 
 
 SMALL_PAYLOAD = 32
+HUGE_CHUNK = 1024 * 1024
 
 
 @pytest.fixture
@@ -35,8 +38,14 @@ def sftp_session(ssh_client_session):
         pytest.param(
             # 1B larger than chunk size in sftp to make sure we exercise
             # at least two rounds of reading/writing
-            SFTP_MAX_CHUNK + 1,
+            sftp.SFTP_MAX_CHUNK + 1,
             id='large-payload',
+        ),
+        pytest.param(
+            # 1B larger than any limits advertised by SFTP servers
+            # passing this directly on wire will not be handled by server
+            HUGE_CHUNK + 1,
+            id='huge-payload',
         ),
     ),
 )
@@ -122,3 +131,32 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+LIBSSH_VERSION_TUPLE = tuple(map(int, __libssh_version__.split('.')))
+xfail_on_libssh_below0_11_as_crashing = pytest.mark.xfail(
+    LIBSSH_VERSION_TUPLE < (0, 11, 0),
+    reason='libssh < 0.11.0 does not partition large writes.',
+    strict=True,
+    raises=LibsshSFTPException,
+)
+
+
+@xfail_on_libssh_below0_11_as_crashing  # pragma: no cover
+@pytest.mark.parametrize(
+    'transmit_payload',
+    (HUGE_CHUNK + 1,),
+    indirect=True,
+    ids=('huge-payload',),
+)
+def test_put_chunk_partition(
+    dst_path,
+    src_path,
+    sftp_session,
+    transmit_payload,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Check the large write is split into chunks by libssh."""
+    monkeypatch.setattr(sftp, 'SFTP_MAX_CHUNK', HUGE_CHUNK)
+    sftp_session.put(str(src_path), str(dst_path))
+    assert dst_path.read_bytes() == transmit_payload


src/pylibsshext/sftp.pyx
```diff
@@ -24,7 +24,9 @@ from pylibsshext.errors cimport LibsshSFTPException
 from pylibsshext.session cimport get_libssh_session
 
 
-SFTP_MAX_CHUNK = 32_768  # 32kB
+# 32kB is safe value if we can not determine better value from the server.
+# for example using limits@openssh.com
```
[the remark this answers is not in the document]
Jakuje: i think this should have been a comma.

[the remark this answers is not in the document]
webknjaz: yep, so accept the suggestion

src/pylibsshext/sftp.pyx
```diff
@@ -58,32 +60,43 @@ cdef class SFTP:
             self._libssh_sftp_session = NULL
 
     def put(self, local_file, remote_file):
+        """
+        Upload local file to remote server
```
webknjaz: Docstring summary lines are customary full sentences with a trailing period at the end. If it fits in one line, double-tripple quotes can also be collapsed. PEP 257.

src/pylibsshext/sftp.pyx
```diff
@@ -58,32 +60,43 @@ cdef class SFTP:
...
-                            remote_file,
-                            self._get_sftp_error_str(),
-                        )
-                    )
+                raise LibsshSFTPException(
+                    "Opening remote file [%s] for write failed with error [%s]"
+                    % (remote_file, self._get_sftp_error_str())
+                )
+            try:
                 buffer = f.read(SFTP_MAX_CHUNK)
```
webknjaz: Urgh.. While not *directly* related, `buffer` is a builtin function name in Python3. This re-assigns it to bytes (while it's typed as a callable returning a `memoryview()` in the global context, IIRC). This would be something to fix indendendently.

src/pylibsshext/sftp.pyx
webknjaz: Note that it's a good idea to also add `:param:` lists and maybe call out acceptable types in the docstrings or the type annotations. I think this would be in the scope of the doc string addition change.

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed various issues with the existing SFTP operations that could
```
webknjaz: This doesn't seem specific enough. I think the generic nature of this sentence is a direct consequence of stuffing of at least three conceptually different changes in the same commit (and a PR for that matter). I recommend that they are split for ease of discussion. I've noticed a few things that could be submitted in atomic chunks/PRs:
* code formatting — these would be ones of the easiest to accept and merge quickly; unlikely to need a change note
* rephrasing of code comments and docstrings — these may add some value but will block the rest of the PR until their shortcomings are fixed properly; probably doesn't need a news fragment either
* insufficiently explained behavior changes — it looks like there's a C-API exposed and called, and some code reshuffled, which is then claimed to be somehow related to memory leaking in a generic way without a proof provided — neither verbal, nor in the form of tests; this would need a more comprehensive explanation on multiple levels, including the PR description and the changelog fragment, the commit message, we may need to rebase this on top of the previous two so that with a cleaner diff it's more obvious which changes are related to leaking and not formatting. I have a feeling that we may end up using a better CM-based abstraction once that's clear.
* I think we should strongly consider adding test coverage for cases that supposedly induce the leaks
* we'll need a follow-up to get rid of the `buffer` builtin shadowing confusion too

Jakuje: > I think we should strongly consider adding test coverage for cases that supposedly induce the leaks

Do you have some architectural suggestion how to check for the leaks from python? Gemini suggested `memray` but I do not see that in epel/rhel so getting this integrated into the CI sounds non-trivial. That would be for the deterministic leaks from SFTP attributes.

The other potential leaks come from the exceptions raised during remote read/write, which are unlike and that would demonstrate even harder/impossible., which is why I did not introduce any tests here.

src/pylibsshext/sftp.pyx
Jakuje: Pushed this one too as its touching the same code paths and I would like to avoid creating (and rebasing) stacked PRs)

Jakuje:
Annoying the copilot is right about the recent comments so I guess I will have to include one or two more changes.

The separate changes are now split into separate commits.

src/pylibsshext/sftp.pyx
```diff
@@ -58,32 +61,55 @@ cdef class SFTP:
...
+                while read_buffer != b"":
+                    length = len(read_buffer)
+                    written = sftp.sftp_write(rf, PyBytes_AS_STRING(read_buffer), length)
+                    if written < 0:
+                        raise LibsshSFTPException(
+                            "Writing to remote file [%s] failed with error [%s]"
+                            % (remote_file, self._get_sftp_error_str()))
+                    elif written < length:
+                        read_buffer = read_buffer[written:]
+                        continue
```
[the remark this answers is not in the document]
Jakuje: sftp_write can not return 0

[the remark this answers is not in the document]
Jakuje: but better be safe

docs/changelog-fragments/862.doc.rst
```diff
@@ -0,0 +1,2 @@
+Added documentation for SFTP ``get()`` and ``put()``
```
webknjaz: This wasn't really added to the docs but to the docstrings, which are sorta different. So let's say docstrings, not docs. Also, reference objects like they are accessible from Python — `SFTP.get()`, not just `get()`.

tests/unit/sftp_test.py
```diff
@@ -115,3 +118,31 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+LIBSSH_VERSION_TUPLE = tuple(map(int, __libssh_version__.split('.')))
+xfail_if_old_libssh = pytest.mark.xfail(
+    LIBSSH_VERSION_TUPLE < (0, 11, 0),
+    reason='libssh < 0.11.0 does not partition large writes.',
+    strict=True,
```
webknjaz, line 140: Could you add `raises=` to show how *exactly* this would crash on older `libssh`?

tests/unit/sftp_test.py
```diff
@@ -21,23 +25,22 @@ def sftp_session(ssh_client_session):
...
+    * 0 is for empty file.
+    * 32 is arbitrary small value.
+    * SFTP_MAX_CHUNK + 1 (32kB + 1B) is meant to be 1B larger than the chunk
+      size used in :file:`sftp.pyx` to make sure we exercise at least two
+      rounds of reading/writing.
+    * HUGE_CHUNK + 1 (1MB + 1B) is meant to be 1B larger than any limits
+      advertised by the SFTP servers and it should enforce partitioning on the
+      libssh side (if it gets that large buffer) and returning partial writes.
     """
-    payload_len = request.param
```
webknjaz, line 45: Keep this on a separate line and add an "is int" assertion after for type narrowing when we have proper type checking enabled.

tests/unit/sftp_test.py
```diff
@@ -21,23 +25,22 @@ def sftp_session(ssh_client_session):
...
-    size used in :file:`sftp.pyx` to make sure we exercise at least two rounds of
-    reading/writing.
+    * 0 is for empty file.
+    * 32 is arbitrary small value.
+    * SFTP_MAX_CHUNK + 1 (32kB + 1B) is meant to be 1B larger than the chunk
+      size used in :file:`sftp.pyx` to make sure we exercise at least two
+      rounds of reading/writing.
+    * HUGE_CHUNK + 1 (1MB + 1B) is meant to be 1B larger than any limits
+      advertised by the SFTP servers and it should enforce partitioning on the
+      libssh side (if it gets that large buffer) and returning partial writes.
```
webknjaz: Why are these in a docsting? These are clearly params, and they are already explained in IDs. It'd make more sense to attach more context next to `ids=` per line.

Jakuje: We had these when we had less of them. Right, we could add them to the comments if you prefer

tests/unit/sftp_test.py
```diff
@@ -41,11 +40,7 @@ def transmit_payload(request: pytest.FixtureRequest) -> bytes:
       advertised by the SFTP servers and it should enforce partitioning on the
       libssh side (if it gets that large buffer) and returning partial writes.
     """
-    payload_len = request.param
-    random_bytes = [
-        ord(random.choice(string.printable)) for _ in range(payload_len)
```
webknjaz: So your commit message is talking about some mysterious performance optimization which may be fair but keeps the details secret. It'd be nice to reveal some more of the behind-the-scenes of your motivation here. #638 seems to have attempted to preserve the data printable, but I haven't found any evidence wrt why. It's possible you might've wanted any pytest report errors to be visually inspectable with diff easily comparable between the vars. I'm not sure.

Whatever the reasoning is, we have to make sure there's a trail of the transparently documented justification written down somewhere. I'd like to see it in the commit message. [Commit messages *are* important.](https://xcancel.com/nedbat/status/1283147492616556544)

docs/changelog-fragments/862.bugfix.rst
webknjaz, line 1: Why do some of the list items end with trailing periods while others don't?

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+LIBSSH_VERSION_TUPLE = tuple(map(int, __libssh_version__.split('.')))
+xfail_if_old_libssh = pytest.mark.xfail(
+    LIBSSH_VERSION_TUPLE < (0, 11, 0),
+    reason='libssh < 0.11.0 does not partition large writes.',
+    strict=True,
+    raises=LibsshSFTPException,
+)
+
+
+@xfail_if_old_libssh
```
webknjaz: Since this is only used once, it's not mandatory to put this into a var. OTOH, it might be useful if we can come up with a good explanatory name.

In some projects, I also automatically auto-exclude `@pytest.mark.xfail` from coverage [but here I forgot for some reason](https://github.com/ansible/pylibssh/blob/521cc01656b3de56b847aba8f87f7881e806c36a/.coveragerc#L16-L22). One I fix it later, the standard syntax will work, but custom aliases will keep needing a comment like this:
```suggestion
@xfail_on_libssh_below_0_11_as_crashing  # pragma: no cover
```

tests/unit/sftp_test.py
webknjaz: Marking as resolved.

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+@pytest.mark.parametrize(
+    'transmit_payload',
+    (HUGE_CHUNK + 1,),
+    indirect=True,
+    ids=['huge-payload'],
```
webknjaz: We usually prefer tuples around here
```suggestion
    ids=('huge-payload',),
```

tests/unit/sftp_test.py
```diff
@@ -1,13 +1,18 @@
 """Tests suite for sftp."""
 
 import random
-import string
 import uuid
 
 import pytest
 
+from pylibsshext.errors import LibsshSFTPException
 from pylibsshext.sftp import SFTP_MAX_CHUNK
```
webknjaz: Best
```suggestion
from pylibsshext import sftp
```
So you'd be able to patch module-level stuff

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+    indirect=True,
+    ids=['huge-payload'],
+)
+def test_put_chunk_partition(
+    dst_path,
+    src_path,
+    sftp_session,
+    transmit_payload,
+    monkeypatch,
+):
```
webknjaz:
```suggestion
) -> None:
```

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+    """Check the large write is split into chunks by libssh."""
+    monkeypatch.setattr('pylibsshext.sftp.SFTP_MAX_CHUNK', HUGE_CHUNK)
```
webknjaz: With https://github.com/ansible/pylibssh/pull/862/changes#r3730119315, you can patch module-level vars
```suggestion
    monkeypatch.setattr(sftp, 'SFTP_MAX_CHUNK', HUGE_CHUNK)
```

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+
+
+@xfail_if_old_libssh
+@pytest.mark.parametrize(
+    'transmit_payload',
+    (HUGE_CHUNK + 1,),
+    indirect=True,
+    ids=['huge-payload'],
+)
+def test_put_chunk_partition(
```
webknjaz, line 152: I'm not *entirely* sure if this belongs among unit tests, honestly. You're trying to influence something in the external thing, right?
Could you explain more about what you're trying to achieve here?

tests/unit/sftp_test.py
webknjaz: They are still there, still w/o the comments, still with the detached explanation in the docstring. Docstrings are there to describe what the function body does, not what's in some disconnected thing like a decorator.

tests/unit/sftp_test.py
```diff
@@ -115,3 +121,32 @@ def test_put_existing(
...
+    dst_path,
+    src_path,
+    sftp_session,
+    transmit_payload,
+    monkeypatch,
```
webknjaz: Could you add typing here and in all the other newly added signatures?
```suggestion
    monkeypatch: pytest.MonkeyPatch,
```

src/pylibsshext/sftp.pyx
webknjaz: > I would like to avoid creating (and rebasing) stacked PRs)

That's not something avoidable. I'm not sure we'll be able to agree on so many changes going into the same PR altogether. Adding more things might be fine temporarily as long as we understand that they'll have to be picked apart into reviewable+maintainable patches submitted separately and optimized for readability and the maintainer-in-the-loop process. Agentic coding is great, but it nowadays is still heavily tailored for the contributor-in-the-loop-and-the-maintainer-out.

I'm sometimes leaving TODOs for the future as I review things and then some agents that are out of my control start piling up more things into the PR, pushing them farther away from being merged. This also includes force-pushes at arbitrary points in time (GH loosing and misattributing in-flight inline review comments isn't helping here either).

To reduce the agentic disruptiveness, I recommend having controlled rebase (Git and reshape in general) behavior, plus scoping things it does, with postponing other stuff into Git worktrees, separate PRs etc.

A good workflow for working with Git here is
1) adding incremental commits during review cycles, this way, GH review UI doesn't get messed up, and it's easier to see what changed between the iterations, including tracking conversations in the inline threads
2) once the generic shape is agreed upon, the PR should go into a phase where it doesn't get any new content but could get its Git branch shape brushed up (like squashing related commits together) — it's helpful if (1) had very small commits so they could be attached to other commits at this point and there would be no need to split commits with too much stuff before combining parts of them with others; this is when a force-push is acceptable and maybe a rebase on top of the main/devel branch could happen
3) in some cases, multiple unrelated chains of changes end up in a commit during exploration, then (2) turns into multiple iterations where some of those atomic commits get pushed into separate branches with new PRs open; this may require more coordination, yes, but also results in more thoughtful and maintainable outcomes, which is what we're after

----

So I would argue that adding more changes that touch these lines, this may create more friction and delays as we'll still have to go through all these things, since I'm reviewing every single change and each of them will influence blocking this entire PR and then us having to figure out how to split things out into small chunks anyway. [It's rather tiring](https://ep2026.europython.eu/session/the-human-in-the-loop-is-tired), and [I'm kinda trying to work out ways to break through the isolation in collab](https://ep2026.europython.eu/session/defending-open-source-from-ai-slop-a-maintainer-s-practical-guide).

----

One more thing — if Copilot or whatever robot is right about a thing in the repo, it's still okay to ignore it and record it as an issue for the future instead. Or submit as a separate PR. This approach will definitely get us closer to merging PRs sooner by simply reducing the cognitive load of having to dig through patches that are bigger than they need to be *right now*.

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Fixed several potential issues with the SFTP implementation:
+ * Memory leak of SFTP attributes during ``SFTP.get()`` downloads.
```
webknjaz: It's a bit unclear what the exact leaks are, though.

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,9 @@
...
+ * Potential memory leaks of file descriptors on error paths
+ * Fixed corner case where large write is truncated by libssh to match
+   server requested limits.
+ * Handle corner case where short write to local file would not write
```
webknjaz: These all seem to be using different writing styles inconsistent with each other and overall, eh?

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,9 @@
...
+   all requested data
+
+-- by :user:`Jakuje`.
```
webknjaz: If the byline is detached on a separate line like this, it doesn't need a trailing period.

tests/unit/sftp_test.py
```diff
@@ -21,23 +26,24 @@ def sftp_session(ssh_client_session):
 
 
 @pytest.fixture(
-    params=(32, SFTP_MAX_CHUNK + 1),
-    ids=('small-payload', 'large-payload'),
+    params=(0, 32, SFTP_MAX_CHUNK + 1, HUGE_CHUNK + 1),
+    ids=('empty-payload', 'small-payload', 'large-payload', 'huge-payload'),
```
webknjaz: Plz multiline all these entries.
Also, can this be extracted into a separate PR? Sounds like this would  influence existing tests, right? Would the huge payload need to be marked as xfail similarly in other places too? Or just that partition test? Is it possible to wrap the params entries with `pytest.param`? Do fixtures allow that syntax?

I think this would in general make a good standalone PR.

tests/unit/sftp_test.py
```diff
@@ -21,23 +26,24 @@ def sftp_session(ssh_client_session):
...
+      advertised by the SFTP servers and it should enforce partitioning on the
+      libssh side (if it gets that large buffer) and returning partial writes.
     """
     payload_len = request.param
-    random_bytes = [
-        ord(random.choice(string.printable)) for _ in range(payload_len)
-    ]
-    return bytes(random_bytes)
+    assert isinstance(payload_len, int)
+    return random.randbytes(payload_len)
```
webknjaz, line 56: Refactoring whatever this is would be another standalone PR.

src/pylibsshext/sftp.pyx
```diff
@@ -17,14 +17,16 @@
 
 from posix.fcntl cimport O_CREAT, O_RDONLY, O_TRUNC, O_WRONLY
 
-from cpython.bytes cimport PyBytes_AS_STRING
```
webknjaz: This seems like something prompted by #819 but doesn't belong here, right?

src/pylibsshext/sftp.pyx
```diff
@@ -17,14 +17,16 @@
...
 from cpython.mem cimport PyMem_Free, PyMem_Malloc
 
 from pylibsshext.errors cimport LibsshSFTPException
 from pylibsshext.session cimport get_libssh_session
 
 
-SFTP_MAX_CHUNK = 32_768  # 32kB
+# The maximal SFTP chunk size we attempt to transfer in a single SFTP packet.
```
```suggestion
# The maximum SFTP chunk size we attempt to transfer in a single SFTP packet.
```

webknjaz:
Alright, so let's move the following into separate PRs to reduce the noise in the diff here:
* https://github.com/ansible/pylibssh/commit/b7051028fbd16770f1b2e47569b61e3195290165, no change note needed
* https://github.com/ansible/pylibssh/commit/eb5c1d1894acd4d9d629b86d2f98c7c4bd7bf463 + https://github.com/ansible/pylibssh/commit/b97fcaf0169b8c60032e602fcd40919b6061558b could probably be both in the same PR, change note not mandatory
* https://github.com/ansible/pylibssh/commit/9396c93e6e520d1469ea8fabc780fd0b56dff557, no change log entry necessary
* https://github.com/ansible/pylibssh/commit/cbe324ab07357d93be32f6647e606cf7cc92d292, optional news entry, mandatory proper git commit message context (https://github.com/ansible/pylibssh/pull/862/changes#r3730036108 / https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79)

With those merged in, we'll have a cleaner PR here. Still not sure what to do with other things that seem like too many changes. But maybe we'll see more things to pick out later.

src/pylibsshext/sftp.pyx
webknjaz: (include this into that commit I asked to move to another PR)

docs/changelog-fragments/862.bugfix.rst
webknjaz: @Jakuje I don't have experience w/ `memray` but if something isn't present in epel/rhel but could be installed in *any* other CI env we have available, that would be enough for me — one job in testing something still counts. I wouldn't care if every single downstream can't repeat the same. We'd still have some sort of proof of the scenario behaving certain way.

Depending on what the leaks are, you could try using the `gc` module, freeze the object collection, do something, count objects, invoke manual collection, count them again, re-enable the GC, compare the numbers: https://github.com/ansible/pylibssh/blob/521cc01656b3de56b847aba8f87f7881e806c36a/tests/unit/channel_test.py#L212-L220 / https://github.com/aio-libs/aiomysql/pull/493/changes#diff-5871d4c569454db5e625383975462132da0bd03d32df145d8d72d8fafd86d952R227. At least for the objects that the Python runtime tracks/manages.

I'm not entirely sure what you mean by "leaks from SFTP attributes". Do you mean like C-extension side things that are invisible to Python? Low-level C stuff?

tests/unit/sftp_test.py
Jakuje: This was from one of the copilot reviews. While we did short random data, it was ok to generate them byte by byte, but at the scale of creating 1MB chunks, it no longer looks like a good idea.
I guess printable characters were used for debugging. In the past, there were some issues with uploads/downloads that corrupted files and in that case, being able to review the content of printable characters sounds easier than reviewing binary data (even though escaped).
I think we no longer have these issues (though the payload is printed on errors).

tests/unit/sftp_test.py
Jakuje: The libssh since 0.11.0 splits large chunks to digestible SFTP packets. This test case is for the case, where the `sftp_write` returns smaller value than the length of the buffer provided. This is valid behavior (even though the current code does not handle it):
https://github.com/ansible/pylibssh/blob/devel/src/pylibsshext/sftp.pyx#L75

The newer libssh versions also have dynamic bounds based on what is advertised by the sftp server (through limits@openssh.com), so this could be in theory a smaller value than our default chunk size (32k) and it could fail here. But instead of trying to mock the sftp server and libssh to hit this code path, I removed (bumped) the pylibssh limit to force the large buffer up to the libssh layer. I guess this should go to the commit message.

docs/changelog-fragments/862.bugfix.rst
Jakuje: What specific information would you like to read here? During SFTP.get() we allocated attributes, but never free them. One structure per file.

src/pylibsshext/sftp.pyx
Jakuje: Yes and no. Its related to #819, but it was prompted by other copilot comment that pointed out some needless copies of buffers and their handling after my initial approach of solving the partial writes. Removing this first in separate PR/commit might make sense though.

Jakuje:
> Alright, so let's move the following into separate PRs to reduce the noise in the diff here:
> 
>     * [b705102](https://github.com/ansible/pylibssh/commit/b7051028fbd16770f1b2e47569b61e3195290165), no change note needed

#869

>     * [eb5c1d1](https://github.com/ansible/pylibssh/commit/eb5c1d1894acd4d9d629b86d2f98c7c4bd7bf463) + [b97fcaf](https://github.com/ansible/pylibssh/commit/b97fcaf0169b8c60032e602fcd40919b6061558b) could probably be both in the same PR, change note not mandatory

#870

>     * [9396c93](https://github.com/ansible/pylibssh/commit/9396c93e6e520d1469ea8fabc780fd0b56dff557), no change log entry necessary

#871

>     * [cbe324a](https://github.com/ansible/pylibssh/commit/cbe324ab07357d93be32f6647e606cf7cc92d292), optional news entry, mandatory proper git commit message context (https://github.com/ansible/pylibssh/pull/862/changes#r3730036108 / https://gist.github.com/webknjaz/cb7d7bf62c3dda4b1342d639d0e78d79)

#872

I will try to split some more self-contained changes

src/pylibsshext/sftp.pyx
Jakuje: Included in the separate PR #870

tests/unit/sftp_test.py
Jakuje: Included in the separate PR #872.

tests/unit/sftp_test.py
Jakuje: Done this in separate PR #873 (the empty files).

For the huge-payload, it will influence other tests, but it will not xfail in them as the capping is handled by the pylibssh too (to SFTP_MAX_CHUNK) and larger SFTP packet is never crated.

docs/changelog-fragments/862.bugfix.rst
Jakuje: I guess I will just take the sftp attributes to separate object which will take care of the memory lifetime. In separate PR.

Jakuje: -> #874

src/pylibsshext/sftp.pyx
Jakuje: Anyway moved to #875. Further changes will be built on top of these once merged.

docs/changelog-fragments/862.bugfix.rst
Jakuje: The `sftp_stat()` function allocates the `sftp_attributes` structure (with `malloc()`) and returns pointer to the python side, but never freed with `sftp_attributes_free()`. I do not think this would be caught by the gc, but I really did not investigate

tests/unit/sftp_test.py
Jakuje: Done.

tests/unit/sftp_test.py
Jakuje: Moved

tests/unit/sftp_test.py
```diff
@@ -23,25 +28,18 @@ def sftp_session(ssh_client_session):
     params=(
         0,
         32,
-        SFTP_MAX_CHUNK + 1,
+        sftp.SFTP_MAX_CHUNK + 1,
```
webknjaz: @Jakuje could you see if this syntax works in a separate PR?
```suggestion
       pytest.param(
          0,  # empty file
          id='empty-payload',
       ),
       pytest.param(
           32,  # arbitrary small value
           id='small-payload',
       ),
       pytest.param(
           sftp.SFTP_MAX_CHUNK + 1,  # 1B larger than chunk size in sftp
           id='large-payload',
       ),
```
you could also couple some docstring cleanup w/ this. WDYT?

webknjaz:
@Jakuje looks like this also needs a rebase + commit cleanup + `PyBytes_AS_STRING`-related stuff cleanup + the change notes/PR description update

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,9 @@
+Fixed several potential issues with the SFTP implementation:
+
+ * Fixed potential memory leaks of file descriptors on error paths.
```
webknjaz: This one is still a bit vague. Needs more specificity about those mysterious error paths.

tests/unit/sftp_test.py
Jakuje: #877

docs/changelog-fragments/862.bugfix.rst
```diff
@@ -0,0 +1,10 @@
+Fixed several potential issues with the SFTP implementation:
+
+ * Fixed potential memory leaks of file descriptors on error paths during
```
webknjaz: This should probably drop the word “potential” as long as it describes things that can actually happen and aren't fictional.

Jakuje: Dropped.

PR #870, opened 2026-08-10
Improve SFTP doc strings

Jakuje:
##### SUMMARY
Minor doc strings improvements taken from #862.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -24,7 +24,10 @@ from pylibsshext.errors cimport LibsshSFTPException
 from pylibsshext.session cimport get_libssh_session
 
 
-SFTP_MAX_CHUNK = 32_768  # 32kB
+# The maximum SFTP chunk size we attempt to transfer in a single SFTP packet.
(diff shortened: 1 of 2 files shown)


webknjaz (approved):

src/pylibsshext/sftp.pyx
```diff
@@ -58,6 +61,15 @@ cdef class SFTP:
             self._libssh_sftp_session = NULL
 
     def put(self, local_file, remote_file):
```
webknjaz, line 63: It'd be good to add type annotations to these signatures too. Maybe even instead of the `:type:` entries in the docstrings.

Jakuje: As mentioned in https://github.com/ansible/pylibssh/pull/874#discussion_r3782071491 I believe the type annotations with OR/Union can not really be enforced in runtime the cpython code so they end up having the same effect as the doc strings. We can argue which one is better and what types we should actually accept, but that would be for separate topic, which should change/add/fix everything in one PR.

PR #874, opened 2026-08-10
sftp: Introduce stat() method and provide Attributes class

Jakuje:
##### SUMMARY
The stat was used only internally so far and managing its life-cycle was problematic leading to memory leaks.

This change is introducing separate stat() API and Attributes class carying all the SFTP attributes visible to the caller and managing underlying C-structure lifetime.

##### ISSUE TYPE
- Feature Pull Request

### DIFF
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -15,6 +15,8 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 
+import os
+
 from posix.fcntl cimport O_CREAT, O_RDONLY, O_TRUNC, O_WRONLY
 
 from cpython.mem cimport PyMem_Free, PyMem_Malloc
@@ -27,6 +29,8 @@ from pylibsshext.session cimport get_libssh_session
 # The value 32kB is a safe fallback when we cannot determine better value from
 # the server, for example using limits@openssh.com (since libssh 0.11.0).
 SFTP_MAX_CHUNK = 32_768
+
+_NO_ATTR_ERR = "The attribute {attr_name} is not available"
 
 
 MSG_MAP = {
@@ -59,6 +63,49 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
+    @property
+    def version(self) -> int:
+        """
+        The SFTP protocol version.
+
+        The most common protocol version is version 3 implemenbed by OpenSSH. The libssh
+        implements also some parts of version 4 which is partially incompatible with the
+        previous versions.
+        """
+        return sftp.sftp_server_version(self._libssh_sftp_session)
+
+    def stat(self, remote_path: str | os.PathLike[str]) -> Attributes:
+        """
+        Requests information about remote file or directory.
+
+        Requests information about file or directory from the remote server.
+        This information includes its size, owner, group, type and many more.
+
+        :param remote_path: The remote file or directory to find information about
+
+        :raises LibsshSFTPException: If operation failed.
+
+        :return: Remote file or directory attributes
+        """
+        cdef sftp.sftp_attributes attrs
+
+        remote_path_b = os.fspath(remote_path)
+        if isinstance(remote_path_b, str):
+            remote_path_b = remote_path_b.encode("utf-8")
+        else:
+            raise TypeError(f"Expected str or os.PathLike, got {type(remote_path).__name__}")
+
+        attrs = sftp.sftp_stat(self._libssh_sftp_session, remote_path_b)
+        if attrs is NULL:
+            raise LibsshSFTPException(
+                "Failed to stat the remote file [%s]. Error: [%s]"
+                % (
+                    remote_path,
+                    self._get_sftp_error_str(),
+                ),
+            )
+        return Attributes._from_ptr(attrs, self.version)
+
     def put(self, local_file, remote_file):
         """
         Upload local file to remote server.
@@ -87,7 +134,7 @@ cdef class SFTP:
                     "Opening remote file [%s] for write failed with error [%s]"
                     % (
                         remote_file,
-                        self._get_sftp_error_str()
+                        self._get_sftp_error_str(),
                     ),
                 )
             read_buffer = f.read(SFTP_MAX_CHUNK)
@@ -120,21 +167,12 @@ cdef class SFTP:
         """
         cdef sftp.sftp_file rf
         cdef char *read_buffer = NULL
-        cdef sftp.sftp_attributes attrs
 
         remote_file_b = remote_file
         if isinstance(remote_file_b, unicode):
             remote_file_b = remote_file.encode("utf-8")
 
-        attrs = sftp.sftp_stat(self._libssh_sftp_session, remote_file_b)
-        if attrs is NULL:
-            raise LibsshSFTPException(
-                "Failed to stat the remote file [%s]. Error: [%s]"
-                % (
-                    remote_file,
-                    self._get_sftp_error_str(),
-                ),
-            )
+        attrs = self.stat(remote_file)
         file_size = attrs.size
 
         rf = sftp.sftp_open(self._libssh_sftp_session, remote_file_b, O_RDONLY, sftp.S_IRWXU)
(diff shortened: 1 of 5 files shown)


src/pylibsshext/sftp.pyx
```diff
@@ -57,6 +57,32 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
+    def stat(self, remote_path) -> Attributes:
```
webknjaz: We can add a type annotation right here instead of the docstring:
```suggestion
    def stat(self, remote_path: str | bytes) -> Attributes:
```

src/pylibsshext/sftp.pyx
```diff
@@ -57,6 +57,32 @@ cdef class SFTP:
...
+        """Requests information about remote file or directory
+
+        Requests information about file or directory from the remote server.
+        This information includes its size, owner, group, type and many more.
+
+        :param remote_path: The remote file or directory to find information about
+        :type remote_path: str or bytes
```
```suggestion
```

src/pylibsshext/sftp.pyx
```diff
@@ -57,6 +57,32 @@ cdef class SFTP:
...
+
+        :raises LibsshSFTPException: If operation failed.
+
+        :return: Remote file or directory attributes
+        :rtype: Attributes
```
webknjaz: It's already in the type annotation
```suggestion
```

src/pylibsshext/sftp.pyx
```diff
@@ -57,6 +57,32 @@ cdef class SFTP:
...
+        """
+        cdef sftp.sftp_attributes attrs
+
+        remote_path_b = remote_path
+        if isinstance(remote_path_b, unicode):
```
```suggestion
        if isinstance(remote_path_b, str):
```

src/pylibsshext/sftp.pyx
```diff
@@ -57,6 +57,32 @@ cdef class SFTP:
...
+            remote_path_b = remote_path.encode("utf-8")
+        attrs = sftp.sftp_stat(self._libssh_sftp_session, remote_path)
+        if attrs is NULL:
+            raise LibsshSFTPException(
+                "Failed to stat the remote file [%s]. Error: [%s]"
+                % (remote_path, self._get_sftp_error_str()))
```
```suggestion
                % (
                    remote_path,
                    self._get_sftp_error_str(),
                ),
            )
```

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +162,78 @@ cdef class SFTP:
...
+    @property
+    def owner(self):
+        return self.attrs.owner.decode('utf-8') if self.attrs.owner else None
+
+    @property
+    def group(self):
+        return self.attrs.group.decode('utf-8') if self.attrs.group else None
+
+    @property
+    def flags(self): return self.attrs.flags
```
webknjaz: Plz add docstrings and type annotations into all public attributes

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +162,78 @@ cdef class SFTP:
...
+cdef class Attributes:
+    @staticmethod
+    cdef Attributes _from_ptr(sftp.sftp_attributes ptr):
+        """Internal C-factory method to wrap a raw C pointer into a Python object."""
+        cdef Attributes obj = Attributes.__new__(Attributes)
+        obj.attrs = ptr
+        return obj
+
+    def __dealloc__(self):
+        if self.attrs is not NULL:
```
webknjaz: This reads better if you return early and dedent the rest.
```suggestion
    def __dealloc__(self) -> None:
        if self.attrs is NULL:
            return
```

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +162,78 @@ cdef class SFTP:
...
+            sftp.sftp_attributes_free(self.attrs)
+            self.attrs = NULL
+
+    @property
+    def name(self):
+        return self.attrs.name.decode('utf-8') if self.attrs.name else None
```
webknjaz: This looks like a weird API with a property that returns `str | None`. It's best to always return a `str` and be clear if something's not there by raising a `LookupError` instead of silently converting an empty string (or bytestring??) into `None`.

tests/unit/sftp_test.py
```diff
@@ -115,3 +116,23 @@ def test_put_existing(
...
+    attrs = sftp_session.stat(str(src_path))
+    assert attrs.size == len(transmit_payload)
+
+
+def test_stat_non_existing(
+    dst_path,
+    sftp_session,
+):
+    """Check the stat raises exception on non-existing file."""
+    error_msg = r'^Failed to stat the remote file \[.*\]. Error: \[File doesn\'t exist\]$'
```
webknjaz: Can't this check be more specific?
```suggestion
    error_msg = fr'^Failed to stat the remote file \[{dst_path}\]. Error: \[File doesn\'t exist\]$'
```

tests/unit/sftp_test.py
```diff
@@ -115,3 +116,23 @@ def test_put_existing(
...
+    with pytest.raises(LibsshSFTPException, match=error_msg):
+        sftp_session.stat(str(dst_path))
```
webknjaz: Why can't this API accept a pathlib object? Also, the test should check both input types — str and pathlib.Path.

src/pylibsshext/sftp.pyx
webknjaz: Though, we should really think of the high-level APIs here:
```suggestion
    def stat(self, remote_path: str | bytes | os.PathLike[str] | os.PathLike[bytes]) -> Attributes:
```

webknjaz: On the second thought, isn't this API supposed to be high-level? It's probably a bad idea to accept bytes and guess the encoding to be utf-8, hoping for the best. The sane thing to do is to have the caller pass in objects with the encoding that they would already know to be correct.

```suggestion
    def stat(self, remote_path: str | os.PathLike[str]) -> Attributes:
```

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +162,78 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class Attributes:
```
webknjaz, line 241: It feels like this entire class isn't getting test coverage it deserves. We should definitely fix this.

tests/unit/sftp_test.py
```diff
@@ -115,3 +116,23 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+def test_stat_existing(
```
webknjaz: I'd add some tests with `@pytest.mark.parametrize` exercising more of the new props, especially since they have nested attr access and branches inside.

src/pylibsshext/sftp.pyx
```diff
@@ -140,3 +162,78 @@ cdef class SFTP:
...
+    @staticmethod
+    cdef Attributes _from_ptr(sftp.sftp_attributes ptr):
+        """Internal C-factory method to wrap a raw C pointer into a Python object."""
```
webknjaz: Plz start w/ a verb for methods and use active voice / imperative mood for callables.

src/pylibsshext/sftp.pyx
webknjaz, line 241: Also add a class docstring plz.

webknjaz (commented):
Here's some initial high-level feedback on this patch you should be able to act on.

src/pylibsshext/sftp.pyx
Jakuje: After discussing the type signatures in cython with gemini, it suggested that the Cython does not really have a way to enforce the Union/OR types at runtime so it will use generic python `object` and the effect would be just like a type declaration in the docstring.

Regarding the PathLike, I am not much convinced the remote paths should be PathLike as they might have different directory separators and many other differences between systems.

src/pylibsshext/sftp.pyx
Jakuje: why?

tests/unit/sftp_test.py
Jakuje: This is being consistent with the get()/put() methods, which do the same.
I think we can accept PathLike for local files, but for remote files, I have some reservations.

src/pylibsshext/sftp.pyx
Jakuje: Added.

Jakuje: Right. Added.

tests/unit/sftp_test.py
Jakuje: Applied.

src/pylibsshext/sftp.pyx
Jakuje: Adjusted. Thanks!

src/pylibsshext/sftp.pyx
Jakuje: Added.

src/pylibsshext/sftp.pyx
Jakuje: Right. We no longer care about python2. I kept unicode to be systematic with existing code, but it really does not make sense to keep it. I will adjust also the other places in other PR.

src/pylibsshext/sftp.pyx
webknjaz: @Jakuje that doesn't really matter. Type annotations in Python mainly exist for type checkers — this is like linting that the callers can use when they'll call this API. So when you accept these, you'll still need to do a runtime check and conversion to whatever type the low-level API needs. And maybe error out on weird types. My feedback is mainly from the PoV of the high-level API accessible from Python here and the expectations an average pythonista would have when calling it.

So what exactly are you concerned about and is this new API you're adding going to be considered high-level by its end-users or not?

src/pylibsshext/sftp.pyx
webknjaz: Mainly just to match the style in your other formatting PR. It's minor otherwise.

tests/unit/sftp_test.py
webknjaz: @Jakuje the idea is that the method will still have to normalize it internally.

src/pylibsshext/sftp.pyx
Jakuje: Read through the rest of the PRs so I agree. I will change these,

src/pylibsshext/sftp.pyx
Jakuje: Good point. Yes, this should be the high-level API (even though it started internally and small :)). I will add these.

tests/unit/sftp_test.py
Jakuje: Good point. I will include these.

src/pylibsshext/sftp.pyx
```diff
@@ -60,6 +60,42 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
+    @property
+    def version(self) -> int:
+        """
+        The SFTP protocol version
```
webknjaz: Use full sentences with trailing periods for summary lines of docstrings per PEP 257.
```suggestion
        The SFTP protocol version.
```

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class Attributes:
+    """
+    SFTP Attributes providing information about remote file
```
```suggestion
    SFTP Attributes providing information about remote file.
```
same

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
...
+
+    Note, that not all the information is always available (see the doc string describing availability).
+    Unavailable properties raise LookupError exception.
+    """
+
+    @staticmethod
+    cdef Attributes _from_ptr(sftp.sftp_attributes ptr, int version):
+        """Create a new object from a raw C pointer"""
```
```suggestion
        """Create a new object from a raw C pointer."""
```
same

src/pylibsshext/sftp.pyx
Jakuje: Though we use stat() internally in get() which gets bytes already so accepting bytes is probably the only reasonable way to keep that working reliably (rather than converting the possible bytes on input of get() to string artificially here and then back to bytes inside of stat().

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
...
+        if self.attrs is NULL:
+            return
+
+        sftp.sftp_attributes_free(self.attrs)
+        self.attrs = NULL
+
+    @property
+    def name(self) -> str:
+        """
+        The file name
```
```suggestion
        The file name.
```
same

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
...
+        This is set only since SFTP protocol version 4 or with OpenSSH.
+        """
+        if self.attrs.flags & sftp.SSH_FILEXFER_ATTR_OWNERGROUP == 0 or not self.attrs.group:
+            raise LookupError()
+        return self.attrs.group.decode('utf-8')
+
+    @property
+    def is_regular(self) -> bool:
+        """
+        True if the file is a regular file
```
webknjaz: Same. The docstrings aren't code comments. They have different audience and writing style. Think end-users (external callers) vs. core devs (fellow contributors).
```suggestion
        Attribute representing whether the object is a regular file or not.
```

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
...
+        self.attrs = NULL
+
+    @property
+    def name(self) -> str:
+        """
+        The file name
+
+        Note, that this attribute is not set when Attributes come from stat().
+        """
+        if not self.attrs.name:
```
webknjaz: This is still equal to `if self.attrs.name == 0 or self.attrs.name == '' or self.attrs.name is None:`. But you probably just meant

```suggestion
        if self.attrs.name is None:
```

right?

Jakuje: no. The attrs is a c structure so it will be NULL pointer.

Jakuje: so likely `self.attrs.name is NULL` would be better

webknjaz: Also, what about the `if self.attrs is NULL:` case? Because in such a situation, the control flow won't reach the following raise but would explode earlier.

src/pylibsshext/sftp.pyx
```diff
@@ -161,3 +193,241 @@ cdef class SFTP:
...
+            raise LookupError()
```
webknjaz: In Python, you don't actually need `()` when raising exceptions w/ no args:
```suggestion
            raise LookupError
```

*However*, I'd recommend writing unique error messages with good DX that could also be used for testing.

You could have some global message template constant and use it like

```suggestion
            raise LookupError(_NO_ATTR_ERR.format(attr_name='name'))
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +128,55 @@ def test_put_existing(
...
+
+    # file type
+    assert attrs.is_regular
+    props = ('is_dir', 'is_symlink', 'is_special', 'is_unknown')
+    for prop in props:
+        assert not getattr(attrs, prop)
+
+    # These would work only with SFTPv4 (not implemented in OpenSSH)
+    # or with other operation than stat
+    props = (
```
webknjaz: This must either go to `@pytest.mark.parametrize` or be wired via `pytest-subtest`.

tests/unit/sftp_test.py
```diff
@@ -122,3 +128,55 @@ def test_put_existing(
...
+        'group',
+        'atime64',
+        'atime_nseconds',
+        'createtime',
+        'createtime_nseconds',
+        'mtime64',
+        'mtime_nseconds',
+    )
+    for prop in props:
+        with pytest.raises(LookupError):
```
webknjaz: `match=` is a must.

tests/unit/sftp_test.py
```diff
@@ -122,3 +128,55 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+def test_stat_existing(  # noqa: WPS218
+    src_path,
+    transmit_payload,
+    sftp_session,
+):
```
webknjaz: typing plz?

tests/unit/sftp_test.py
```diff
@@ -122,3 +128,55 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+def test_stat_existing(  # noqa: WPS218
```
webknjaz, line 142: why?

src/pylibsshext/sftp.pyx
webknjaz: I'd do the conversion in `get()` then.

tests/unit/sftp_test.py
Jakuje: too many assertions.

Jakuje:
I will probably remove the SFTPv4 attributes as there is really not a good use of them ...

src/pylibsshext/sftp.pyx
Jakuje: Yeah, this already does the conversion, but for error logging, passing the str/origianl between these two is better, I think.

src/pylibsshext/sftp.pyx
Jakuje: Added.

src/pylibsshext/sftp.pyx
Jakuje: If attrs is null, I guess this should be caught in the factory and the object should not be constructed. I can add a check about that in there.

src/pylibsshext/sftp.pyx
Jakuje: Added. Thanks!

tests/unit/sftp_test.py
Jakuje: Added.

tests/unit/sftp_test.py
Jakuje: Typed

tests/unit/sftp_test.py
Jakuje: Moved to parametrized

src/pylibsshext/sftp.pyx
Jakuje: But the null check is already in the stat(). But better be safe.

PR #875, opened 2026-08-11
Remove use of CPython internal PyBytes_AS_STRING API

Jakuje:
##### SUMMARY

The CPython API PyBytes_AS_STRING is needed only when writing Python C Extensions in pure C. In CPython, this is handled automatically.

This is requirement for #819 (compiling with limited API).

##### ISSUE TYPE
- Misc Pull Request

### DIFF
--- a/docs/changelog-fragments/875.breaking.rst
+++ b/docs/changelog-fragments/875.breaking.rst
@@ -1,0 +1,6 @@
+Changed the ``Channel.write()`` API to accept only :class:`bytes` type.
+Previously, this API used to accept anything that could have been converted
+to a pointer using :c:func:`PyBytes_AS_STRING` and that would implement
+:meth:`~object.__len__()` interface, which could result in unexpected behavior
+even for UTF8 strings. Anything else than :class:`bytes` now raises
+:exc:`TypeError` -- by :user:`Jakuje`.
--- a/docs/changelog-fragments/875.misc.rst
+++ b/docs/changelog-fragments/875.misc.rst
@@ -1,0 +1,3 @@
+Removed the needless use of CPython internal API
+:c:func:`PyBytes_AS_STRING` throughout the code to simplify code
+and improve the compatibility with limited Python API -- by :user:`Jakuje`.
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -19,7 +19,6 @@ import signal
 import time
 from io import BytesIO
 
-from cpython.bytes cimport PyBytes_AS_STRING
 from libc.string cimport memset
 
 from pylibsshext.errors cimport LibsshChannelException
@@ -126,7 +125,8 @@ cdef class Channel:
         return self.read_nonblocking(size=size, stderr=stderr)
 
     def write(self, data):
-        written = libssh.ssh_channel_write(self._libssh_channel, PyBytes_AS_STRING(data), len(data))
+        cdef const char* c_buf = data
+        written = libssh.ssh_channel_write(self._libssh_channel, c_buf, len(data))
         if written == libssh.SSH_ERROR:
             raise LibsshChannelException("Failed to write to ssh channel")
         return written
--- a/src/pylibsshext/channel.pyx
+++ b/src/pylibsshext/channel.pyx
@@ -124,8 +124,16 @@ cdef class Channel:
     def recv(self, size=1024, stderr=0):
         return self.read_nonblocking(size=size, stderr=stderr)
 
-    def write(self, data):
+    def write(self, data: bytes) -> int:
+        """
+        Write data to the SSH channel.
+
+        :param data: The payload to write to the channel
+
+        :return: Number of bytes written
+        """
         cdef const char* c_buf = data
+
         written = libssh.ssh_channel_write(self._libssh_channel, c_buf, len(data))
         if written == libssh.SSH_ERROR:
             raise LibsshChannelException("Failed to write to ssh channel")
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -175,9 +175,14 @@ cdef class Session(object):
             value_long = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_long)
         else:
+            # We need to keep temporary python variable before assigning it
+            # to c pointer, otherwise GC would collect the object and we
+            # would crash on use-after-free
+            # https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#encoding-text-to-bytes
+            value_b = value
             if isinstance(value, str):
-                value = value.encode("utf-8")
-            c_buf = value
+                value_b = value.encode("utf-8")
+            c_buf = value_b
             libssh.ssh_options_set(self._libssh_session, key_m, c_buf)
             if key in OPTS_DIR_MAP:
                 self._opts[key] = value
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -175,7 +175,7 @@ cdef class Session(object):
             value_long = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_long)
         else:
-            if isinstance(value, basestring):
+            if isinstance(value, str):
                 value = value.encode("utf-8")
             c_buf = value
             libssh.ssh_options_set(self._libssh_session, key_m, c_buf)
--- a/src/pylibsshext/session.pyx
+++ b/src/pylibsshext/session.pyx
@@ -17,8 +17,6 @@
 import inspect
 import logging
 
-from cpython.bytes cimport PyBytes_AS_STRING
-
 from pylibsshext.channel import Channel
 from pylibsshext.errors cimport LibsshSessionException
 from pylibsshext.logging import _initialize_logging, _set_level
@@ -158,6 +156,7 @@ cdef class Session(object):
         cdef int value_int
         cdef unsigned int value_uint
         cdef long value_long
+        cdef const char* c_buf
 
         key_m = None
         if key in OPTS_DIR_MAP:
@@ -178,7 +177,8 @@ cdef class Session(object):
         else:
             if isinstance(value, basestring):
                 value = value.encode("utf-8")
-            libssh.ssh_options_set(self._libssh_session, key_m, PyBytes_AS_STRING(value))
+            c_buf = value
+            libssh.ssh_options_set(self._libssh_session, key_m, c_buf)
             if key in OPTS_DIR_MAP:
                 self._opts[key] = value
 
--- a/tests/unit/channel_test.py
+++ b/tests/unit/channel_test.py
@@ -1,11 +1,13 @@
 """Tests suite for channel."""
 
 import gc
+import pathlib
 import signal
 import time
 
 import pytest
 
+from pylibsshext.channel import Channel
 from pylibsshext.errors import LibsshChannelException
 from pylibsshext.session import Session
 
@@ -195,6 +197,45 @@ def test_is_eof(ssh_channel):
     ssh_channel.request_exec('exit 0')
     ssh_channel.poll(timeout=POLL_TIMEOUT)
     assert ssh_channel.is_eof
+
+
+@pytest.fixture
+def server_side_tmp_file_path(tmp_path: pathlib.Path) -> pathlib.Path:
+    """Create a new unique path that can be written."""
+    return tmp_path / 'test-file.txt'
+
+
+@pytest.fixture
+def _remote_tmp_file_data_writer(
+    server_side_tmp_file_path: pathlib.Path,
+    ssh_channel: Channel,
+) -> None:
+    """Run a server-side process writing network-received bytes to disk."""
+    ssh_channel.request_exec(f'cat > {server_side_tmp_file_path}')
+
+
+@pytest.mark.usefixtures('_remote_tmp_file_data_writer')
+def test_write_bytes(
+    server_side_tmp_file_path: pathlib.Path,
+    ssh_channel: Channel,
+) -> None:
+    """Test that write() takes bytes correctly."""
+    data_to_write = b'just pure ASCII bytes'
+    ssh_channel.write(data_to_write)
+    ssh_channel.send_eof()
+    ssh_channel.poll(timeout=POLL_TIMEOUT)
+    assert server_side_tmp_file_path.read_bytes() == data_to_write
+
+
+@pytest.mark.usefixtures('_remote_tmp_file_data_writer')
+def test_write_utf8_str(ssh_channel: Channel) -> None:
+    """Test that write() rejects non-bytes."""
+    data_to_write = '∮ E⋅da = Q,  n → ∞, ∑ f(i)'
+    error_msg = (
+        r"^Argument 'data' has incorrect type \(expected bytes, got str\)$"
+    )
+    with pytest.raises(TypeError, match=error_msg):
+        ssh_channel.write(data_to_write)
 
 
 def test_destructor(ssh_session_connect):
--- a/src/pylibsshext/scp.pyx
+++ b/src/pylibsshext/scp.pyx
@@ -90,7 +90,7 @@ cdef class SCP:
                     c_buf = read_buffer
 
                     # Write to the open file
-                    rc = libssh.ssh_scp_write(scp, c_buf, read_bytes)
+                    rc = libssh.ssh_scp_write(scp, c_buf, len(read_buffer))
                     if rc != libssh.SSH_OK:
                         raise LibsshSCPException("Can't write to remote file: %s" % self._get_ssh_error_str())
                     remaining_bytes_to_read -= read_bytes
--- a/src/pylibsshext/scp.pyx
+++ b/src/pylibsshext/scp.pyx
@@ -17,7 +17,6 @@
 
 import os
 
-from cpython.bytes cimport PyBytes_AS_STRING
 from cpython.mem cimport PyMem_Free, PyMem_Malloc
 
 from pylibsshext.errors cimport LibsshSCPException
@@ -42,6 +41,7 @@ cdef class SCP:
         :param remote_file: The path on the remote host where the file should be placed
         :type remote_file: str
         """
+        cdef const char* c_buf
         remote_file_b = remote_file
         if isinstance(remote_file_b, unicode):
             remote_file_b = remote_file.encode("utf-8")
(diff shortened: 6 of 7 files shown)


src/pylibsshext/session.pyx
```diff
@@ -176,9 +175,11 @@ cdef class Session(object):
             value_long = value
             libssh.ssh_options_set(self._libssh_session, key_m, &value_long)
         else:
-            if isinstance(value, basestring):
-                value = value.encode("utf-8")
-            libssh.ssh_options_set(self._libssh_session, key_m, PyBytes_AS_STRING(value))
+            value_b = value
+            if isinstance(value, unicode):
```
webknjaz: TIL: even though Python 3's `str` is text (as in unicode) and what used to be the `unicode` built-in in Python 2 no longer exists in the global namespace, Cython still shims it as such making the name available: https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#python-string-types-in-cython-code.

This makes things confusing compared to reading just standard Python 3.

Though, I have been considering making an effort to stick to a unified value of the `language_level` compiler directive: https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#compiler-directives. A quick repo search shows that I've never implemented that idea, however.

src/pylibsshext/session.pyx
```diff
@@ -176,9 +175,11 @@ cdef class Session(object):
...
+                value_b = value.encode("utf-8")
+            c_buf = value_b
```
webknjaz, line 185: How about something like this for clarity?
```suggestion
            if isinstance(value, str):
                c_buf = value.encode("utf-8")
            else:
                c_buf = value
```

src/pylibsshext/channel.pyx
```diff
@@ -126,7 +125,23 @@ cdef class Channel:
...
+        Write data to the SSH channel
+
+        :param data: The payload to write to the channel
+        :type data: str or bytes
+
+        :return: Number of bytes written
+        :rtype: int
+        """
+        data_b = data
+        if isinstance(data, unicode):
```
webknjaz:
```suggestion
        if isinstance(data, str):
```

src/pylibsshext/session.pyx
Jakuje: I am fine with that. I just used the same constructs as we already had in the existing code.

Jakuje: Actually, I think this will not work. The python garbage collection will throw away the result of `value.encode("utf-8")` when it will not be assigned to any python variable (only the `c_buf` pointer), which will cause random crashes.

webknjaz: How so? That sounds weird. Do you have any proof? Repro? Docs?

src/pylibsshext/channel.pyx
```diff
@@ -125,8 +124,23 @@ cdef class Channel:
...
+        :param data: The payload to write to the channel
+        :type data: str or bytes
+
+        :return: Number of bytes written
+        """
+        data_b = data
+        if isinstance(data, str):
+            data_b = data.encode("utf-8")
+        elif not isinstance(data, bytes):
+            raise TypeError("write() requires bytes or str, got %s" % type(data).__name__)
```
webknjaz: Let's use f-strings for any new interpolation, not the slow old-style python2 render
```suggestion
            raise TypeError(
                f"write() requires bytes or str, got {data!r}.",
            )
```

src/pylibsshext/channel.pyx
```diff
@@ -125,8 +124,23 @@ cdef class Channel:
     def recv(self, size=1024, stderr=0):
         return self.read_nonblocking(size=size, stderr=stderr)
 
-    def write(self, data):
-        written = libssh.ssh_channel_write(self._libssh_channel, PyBytes_AS_STRING(data), len(data))
+    def write(self, data) -> int:
```
webknjaz: Is it really acceptable to get str here, though?
```suggestion
    def write(self, data: str | bytes) -> int:
```

src/pylibsshext/channel.pyx
```diff
@@ -125,8 +124,23 @@ cdef class Channel:
...
+
+        :param data: The payload to write to the channel
+        :type data: str or bytes
+
+        :return: Number of bytes written
+        """
+        data_b = data
+        if isinstance(data, str):
+            data_b = data.encode("utf-8")
+        elif not isinstance(data, bytes):
```
webknjaz: With this new logic, we need to have a test added to cover it.

src/pylibsshext/session.pyx
Jakuje: https://cython.readthedocs.io/en/latest/src/tutorial/strings.html#encoding-text-to-bytes

> ```
> cdef char* c_string = py_unicode_string.encode('UTF-8')
> ```
> 
> Here, the Cython compiler notices that the code takes a pointer to a temporary string result that will be garbage collected after the assignment. Later access to the invalidated pointer will read invalid memory and likely result in a segfault. Cython will therefore refuse to compile this code.

src/pylibsshext/channel.pyx
```diff
@@ -125,8 +124,23 @@ cdef class Channel:
...
         return self.read_nonblocking(size=size, stderr=stderr)
 
-    def write(self, data):
-        written = libssh.ssh_channel_write(self._libssh_channel, PyBytes_AS_STRING(data), len(data))
+    def write(self, data) -> int:
+        """
+        Write data to the SSH channel
+
+        :param data: The payload to write to the channel
+        :type data: str or bytes
```
webknjaz: This free-form data being accepted smells bad.. Is there any way to know what people actually pass here in the wild?

src/pylibsshext/session.pyx
webknjaz: Okay, thanks. Explain that in a code comment, plz and attach that link too, then.

src/pylibsshext/channel.pyx
Jakuje: The `PyBytes_AS_STRING()` converted anything to pointer so people could have passed anything in. But it really makes sense to pass only bytes so lets stick with bytes.
The attempt to convert str to bytes was made in this assumption somebody might have done that, but I can make it just bytes.

src/pylibsshext/channel.pyx
Jakuje: Removed. I changed to accepting just bytes.

src/pylibsshext/channel.pyx
Jakuje: APplied.

src/pylibsshext/session.pyx
Jakuje: Added. PTAL

src/pylibsshext/session.pyx
Jakuje: Applied.

src/pylibsshext/channel.pyx
```diff
@@ -125,8 +124,19 @@ cdef class Channel:
...
+    def write(self, data: bytes) -> int:
+        """
+        Write data to the SSH channel.
+
+        :param data: The payload to write to the channel
+
+        :return: Number of bytes written
+        """
+        if not isinstance(data, bytes):
+            raise TypeError("write() requires bytes, got {data!r}.")
```
webknjaz: Now that we're rejecting non-byte data types, we definitely need tests to check both code branches.

docs/changelog-fragments/875.misc.rst
webknjaz, line 1: With the behavior change (stricter input data type), I think this could qualify for a `feature` fragment or even `breaking`. Not just a `misc`. WDYT?

https://ansible-pylibssh.rtfd.io/en/latest/contributing/guidelines/#alright-so-how-do-i-add-a-news-fragment

src/pylibsshext/channel.pyx
Jakuje: Turns out this check is not needed anymore (thanks to test!) and just the right type signature creates the right TyperError.

docs/changelog-fragments/875.misc.rst
Jakuje: Good point. Kept this as an informative, but added separate for the breaking change.

src/pylibsshext/channel.pyx
webknjaz: @Jakuje great! I've added some of the things we've been talking about to the in-project agent memory. Hope that next time your robot will just attempt self-testing earlier in the process...

docs/changelog-fragments/875.breaking.rst
```diff
@@ -0,0 +1,5 @@
+Changed the ``Channel.write()`` API to accept only ``bytes`` type. Previously,
+this API used to accept anything that could have been converted to a pointer
+using ``PyBytes_AS_STRING`` and that would implement ``__len__()`` method,
+which could result in unexpected behavior even for UTF8 strings. Anything
+else than ``bytes`` now raises in ``TypeError`` -- by :user:``Jakuje``.
```
webknjaz: We can link the things from CPython docs
```suggestion
Changed the ``Channel.write()`` API to accept only :class:`bytes` type. Previously,
this API used to accept anything that could have been converted to a pointer
using :c:function:`PyBytes_AS_STRING` and that would implement :meth:`~object.__len__` method,
which could result in unexpected behavior even for UTF8 strings. Anything
else than :class:`bytes` now raises a :exc:`TypeError` -- by :user:``Jakuje``.
```

docs/changelog-fragments/875.misc.rst
```diff
@@ -0,0 +1,3 @@
+Removed the needless use of CPython internal API ``PyBytes_AS_STRING``
```
webknjaz: Let's try linking this
```suggestion
Removed the needless use of CPython internal API :c:function:`PyBytes_AS_STRING`
```

src/pylibsshext/session.pyx
webknjaz: Ack. Resolved.

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
     assert ssh_channel.is_eof
 
 
+@pytest.fixture
+def tmp_file(tmp_path):
```
webknjaz: Please use a specific informative name for the fixture that would correspond to how it's used in the tests. Sounds like you wanted it write-only or read-write?

Additionally, please add type annotations to all the newly added test function signatures.

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
...
+    ssh_channel.write(data_to_write)
+    ssh_channel.send_eof()
+    ssh_channel.poll(timeout=POLL_TIMEOUT)
+    assert tmp_file.read_bytes() == data_to_write
+
+
+def test_write_utf8_str(ssh_channel, tmp_file):
+    """Test that write() rejects non-byte."""
+    data_to_write = '∮ E⋅da = Q,  n → ∞, ∑ f(i)'
+    ssh_channel.request_exec(f'cat > {tmp_file}')
```
webknjaz: Why do we need this call here?

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
...
 
 
+@pytest.fixture
+def tmp_file(tmp_path):
+    """Create a new unique path that can be written."""
+    return tmp_path / 'test-file.txt'
+
+
+def test_write_bytes(ssh_channel, tmp_file):
+    """Test that write() takes bytes correctl."""
```
webknjaz: missing "y"

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
...
+    data_to_write = b'just pure ASCII bytes'
+    ssh_channel.request_exec(f'cat > {tmp_file}')
```
webknjaz: if we're testing a "write", why do we need a "request_exec"?

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
...
+    ssh_channel.write(data_to_write)
+    ssh_channel.send_eof()
+    ssh_channel.poll(timeout=POLL_TIMEOUT)
+    assert tmp_file.read_bytes() == data_to_write
```
webknjaz: What are we testing here?

Jakuje: that we wrote the data to the file over ssh and they match the data we fed into the write().

tests/unit/channel_test.py
Jakuje: we need to write them somewhere. And there needs to be something running on the other side so we would be able to verify they were written correctly.

tests/unit/channel_test.py
Jakuje: Same as in the other test -- we need something on the other side of ssh channel to receive the data

tests/unit/channel_test.py
webknjaz: Ah, okay. I didn't immediately understand that this is a hanging subprocess on the receiving end because of a bunch of low-level interactions that aren't really being named/abstracted as what they are. Let me think if it's possible to communicate this better through code+comments...

tests/unit/channel_test.py
webknjaz: Here's a clearer name for the fixture with annotations:
```suggestion
def server_side_tmp_file_path(tmp_path: pathlib.Path) -> pathlib.Path:
```

(don't forget to import pathlib at the top of the module)

tests/unit/channel_test.py
```diff
@@ -197,6 +197,33 @@ def test_is_eof(ssh_channel):
     assert ssh_channel.is_eof
 
 
+@pytest.fixture
+def tmp_file(tmp_path):
+    """Create a new unique path that can be written."""
+    return tmp_path / 'test-file.txt'
+
+
+def test_write_bytes(ssh_channel, tmp_file):
```
webknjaz: So instead of having a confusing "cat into tmp" call in the middle of tests, we can move it into a fixture and inject that fixture into both tests like so. This way, we'll have a server-side `cat`-receiver waiting to get bytes before each test execution starts and having a better name w/ its purpose explained.

```suggestion
@pytest.fixture
def _remote_tmp_file_data_writer(
    server_side_tmp_file_path: pathlib.Path,
    ssh_channel: 'pylibsshext.channel.Channel',
) -> None:
    """Run a server-side process writing network-received bytes to disk."""
    ssh_channel.request_exec(f'cat > {server_side_tmp_file_path}')


@pytest.mark.usefixtures('_remote_tmp_file_data_writer')
def test_write_bytes(
    server_side_tmp_file_path: pathlib.Path,
    ssh_channel: 'pylibsshext.channel.Channel',
) -> None:
```

tests/unit/channel_test.py
webknjaz: I think I've found a better way: https://github.com/ansible/pylibssh/pull/875/changes#r3816327667

```suggestion
```

tests/unit/channel_test.py
webknjaz: With https://github.com/ansible/pylibssh/pull/875/changes#r3816327667
```suggestion
@pytest.mark.usefixtures('_remote_tmp_file_data_writer')
def test_write_utf8_str(ssh_channel: 'pylibsshext.channel.Channel') -> None:
    """Test that ``write()`` rejects non-bytes."""
    data_to_write = '∮ E⋅da = Q,  n → ∞, ∑ f(i)'
```
should be enough here.

Jakuje:
Thanks for suggestions! Applied.

tests/unit/channel_test.py
Jakuje: Thanks! This looks cleaner!

tests/unit/channel_test.py
Jakuje: Done

docs/changelog-fragments/875.breaking.rst
```diff
@@ -0,0 +1,6 @@
+Changed the ``Channel.write()`` API to accept only :class:``bytes`` type.
+Previously, this API used to accept anything that could have been converted
+to a pointer using :c:function:``PyBytes_AS_STRING`` and that would implement
+:meth:``~object.__len__()`` method, which could result in unexpected behavior
+even for UTF8 strings. Anything else than :class:``bytes`` now raises
+:exp:``TypeError`` -- by :user:``Jakuje``.
```
webknjaz: There's `:exc:` role that exists, not `:exp:`. Also, your agent got drunk and didn't accept my suggested changes as I originally wrote them (https://github.com/ansible/pylibssh/pull/875#discussion_r3814439822) effectively breaking the RST syntax in a way that it was still kinda “valid” RST but resulted in a broken render by double backticking role arguments and adding undesired parens — here, see it for yourself: https://ansible-pylibssh--875.org.readthedocs.build/en/875/changelog/#vrelease-unreleased-draft.

This should work:
```suggestion
Changed the ``Channel.write()`` API to accept only :class:`bytes` type.
Previously, this API used to accept anything that could have been converted
to a pointer using :c:func:`PyBytes_AS_STRING` and that would implement
:meth:`~object.__len__` interface, which could result in unexpected behavior
even for UTF8 strings. Anything else than :class:`bytes` now raises
:exc:`TypeError` -- by :user:`Jakuje`.
```

I'm pretty sure, this will work if you apply it verbatim, as is. Without any additional changes. The only possible problem could be `:c:func:` as I haven't tested this locally.

But once you apply this change, please don't ask me for review before you actually open the PR docs build and look at the render to see that it does not have raw strings leaking into HTML.

docs/changelog-fragments/875.misc.rst
```diff
@@ -0,0 +1,3 @@
+Removed the needless use of CPython internal API
+:c:function:``PyBytes_AS_STRING`` throughout the code to simplify code
```
webknjaz: And here as well.
```suggestion
:c:func:`PyBytes_AS_STRING` throughout the code to simplify code
```

docs/changelog-fragments/875.breaking.rst
Jakuje: Sorry, no agent involved here. JUst my sloppiness in manually applying changes. Should have been doing that with the GH interface. Verified now that it looks like working all well including the c:func:

https://ansible-pylibssh--875.org.readthedocs.build/en/875/changelog/#vrelease-unreleased-draft

docs/changelog-fragments/875.misc.rst
Jakuje: Fixed also this one. Sorry.

PR #879, opened 2026-08-20
sftp: Handle remote file opening in context manager

Jakuje:
##### SUMMARY
This change refactors get and put methods to use shared code for managing lifetime of remote file descriptors, avoiding code duplication and improving memory handling. This obsoletes one of the changes (fd leaks from #862).

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/879.bugfix.rst
+++ b/docs/changelog-fragments/879.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed memory leaks of remote file descriptors inside ``get()`` and ``put()``
+methods in ``SFTP`` by replacing the manual memory handling with a context
+manager -- by :user:`Jakuje`.
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -17,7 +17,11 @@
 
 from posix.fcntl cimport O_CREAT, O_RDONLY, O_TRUNC, O_WRONLY
 
-from cpython.mem cimport PyMem_Free, PyMem_Malloc
+import typing as _t  # noqa: WPS111
+
+
+if _t.TYPE_CHECKING:  # pragma: no cover
+    from types import TracebackType
 
 from pylibsshext.errors cimport LibsshSFTPException
 from pylibsshext.session cimport get_libssh_session
@@ -69,44 +73,20 @@ cdef class SFTP:
         :param remote_file: The path to upload the file on the remote system
         :type remote_file: str or bytes
         """
-        cdef sftp.sftp_file rf
-        cdef const char* c_buf
-        with open(local_file, "rb") as f:
-            remote_file_b = remote_file
-            if isinstance(remote_file_b, unicode):
-                remote_file_b = remote_file.encode("utf-8")
+        cdef bytearray write_buffer = bytearray(SFTP_MAX_CHUNK)
 
-            rf = sftp.sftp_open(
-                self._libssh_sftp_session,
-                remote_file_b,
-                O_WRONLY | O_CREAT | O_TRUNC,
-                sftp.S_IRWXU,
-            )
-            if rf is NULL:
-                raise LibsshSFTPException(
-                    "Opening remote file [%s] for write failed with error [%s]"
-                    % (
-                        remote_file,
-                        self._get_sftp_error_str()
-                    ),
-                )
-            read_buffer = f.read(SFTP_MAX_CHUNK)
+        with open(local_file, "rb") as local_fd, _RemoteFile(
+            sftp_obj=self,
+            path=remote_file,
+            flags=O_WRONLY | O_CREAT | O_TRUNC,
+        ) as remote_fd:
+            while True:
+                bytes_read = local_fd.readinto(write_buffer)
+                is_eof = bytes_read == 0
+                if is_eof:
+                    break
 
-            while read_buffer != b"":
-                c_buf = read_buffer
-                length = len(read_buffer)
-                written = sftp.sftp_write(rf, c_buf, length)
-                if written != length:
-                    sftp.sftp_close(rf)
-                    raise LibsshSFTPException(
-                        "Writing to remote file [%s] failed with error [%s]"
-                        % (
-                            remote_file,
-                            self._get_sftp_error_str(),
-                        )
-                    )
-                read_buffer = f.read(SFTP_MAX_CHUNK)
-            sftp.sftp_close(rf)
+                remote_fd.write(write_buffer, bytes_read)
 
     def get(self, remote_file, local_file):
         """
@@ -118,8 +98,6 @@ cdef class SFTP:
         :param local_file: The path on the local file system to place the downloaded file
         :type local_file: str or os.PathLike
         """
-        cdef sftp.sftp_file rf
-        cdef char *read_buffer = NULL
         cdef sftp.sftp_attributes attrs
 
         remote_file_b = remote_file
@@ -136,57 +114,28 @@ cdef class SFTP:
                 ),
             )
         file_size = attrs.size
+        buffer_size = min(SFTP_MAX_CHUNK, file_size)
+        cdef bytearray read_buffer = bytearray(buffer_size)
+        buffer_view = memoryview(read_buffer)
 
-        rf = sftp.sftp_open(self._libssh_sftp_session, remote_file_b, O_RDONLY, sftp.S_IRWXU)
-        if rf is NULL:
-            raise LibsshSFTPException(
-                "Opening remote file [%s] for read failed with error [%s]"
-                % (
-                    remote_file,
-                    self._get_sftp_error_str(),
-                ),
-            )
+        with _RemoteFile(
+            sftp_obj=self,
+            path=remote_file,
+            flags=O_RDONLY,
+        ) as remote_fd, open(local_file, 'wb') as local_fd:
+            while True:
+                file_data = remote_fd.read(read_buffer)
+                is_eof = file_data == 0
+                if is_eof:
+                    break
 
-        try:
-            with open(local_file, 'wb') as f:
-                buffer_size = min(SFTP_MAX_CHUNK, file_size)
-                read_buffer = <char *>PyMem_Malloc(buffer_size)
-                if read_buffer is NULL:
-                    raise LibsshSFTPException("Memory allocation error")
-
-                while True:
-                    file_data = sftp.sftp_read(rf, <void *>read_buffer, sizeof(char) * buffer_size)
-                    if file_data == 0:
-                        break
-                    elif file_data < 0:
-                        sftp.sftp_close(rf)
-                        raise LibsshSFTPException(
-                            "Reading data from remote file [%s] failed with error [%s]"
-                            % (
-                                remote_file,
-                                self._get_sftp_error_str(),
-                            ),
-                        )
-
-                    bytes_written = f.write(read_buffer[:file_data])
-                    if bytes_written and file_data != bytes_written:
-                        sftp.sftp_close(rf)
-                        raise LibsshSFTPException(
-                            "Number of bytes [%s] read from remote file [%s]"
-                            " does not match number of bytes [%s] written to"
-                            " local file [%s] due to error [%s]"
-                            % (
-                                file_data,
-                                remote_file,
-                                bytes_written,
-                                local_file,
-                                self._get_sftp_error_str(),
-                            ),
-                        )
-        finally:
-            if read_buffer is not NULL:
-                PyMem_Free(read_buffer)
-        sftp.sftp_close(rf)
+                bytes_written = local_fd.write(buffer_view[:file_data])
+                if bytes_written and file_data != bytes_written:
+                    raise LibsshSFTPException(
+                        f"Number of bytes [{file_data}] read from remote file "
+                        f"[{remote_file!s}] does not match number of bytes "
+                        f"[{bytes_written}] written to local file [{local_file!s}]"
+                    )
 
     def close(self):
         if self._libssh_sftp_session is not NULL:
@@ -198,3 +147,90 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """
+    Remote file handle facade.
+
+    Implements a standard context manager interface for
+    convenient access lifetime management.
+    """
+    def __cinit__(self, *, sftp_obj: SFTP, path: str | bytes, int flags) -> None:
+        self._sftp = sftp_obj
+        self._path = path
+        b_path = path
+        if isinstance(b_path, str):
+            b_path = path.encode("utf-8")
+
+        self._fd = sftp.sftp_open(
+            sftp_obj._libssh_sftp_session,
+            b_path,
+            flags,
+            sftp.S_IRWXU,
+        )
+        if self._fd is not NULL:
+            return
+
+        file_open_mode = "writing" if flags & O_WRONLY else "reading"
+        sftp_error = sftp_obj._get_sftp_error_str()
+        raise LibsshSFTPException(
+            f"Opening remote file [{path!s}] for {file_open_mode!s} "
+            f"failed with error [{sftp_error!s}]",
+        )
+
+    def write(self, data: bytearray, length: int) -> int:
+        """
+        Writes a bytearray directly to the remote file.
+
+        :raises LibsshSessionException: When writing failed or wrong amount of data was written.
+        """
+        cdef const char* c_buf = data
+        written = sftp.sftp_write(self._fd, c_buf, length)
+        if written == length:
+            return written
+
+        sftp_error = self._sftp._get_sftp_error_str()
+        raise LibsshSFTPException(
+            f"Writing to remote file [{self._path!s}] failed with error [{sftp_error}]"
+        )
+
+    def read(self, data: bytearray) -> int:
+        """
+        Reads data from the remote file descriptor into provided bytearray.
+
+        :raises LibsshSessionException: When reading failed.
+        """
+        cdef char* c_buf = data
+        cdef size_t length = len(data)
+
+        file_data = sftp.sftp_read(self._fd, c_buf, length)
+        if file_data >= 0:
+            return file_data
+
+        sftp_error = self._sftp._get_sftp_error_str()
+        raise LibsshSFTPException(
+            f"Reading data from remote file [{self._path!s}] failed with error [{sftp_error!s}]"
+        )
+
+    def __dealloc__(self) -> None:
+        if self._fd is NULL:
+            return
+
+        sftp.sftp_close(self._fd)
+        self._fd = NULL
+
+    def __enter__(self) -> _t.Self:
+        return self
+
+    def __exit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> bool | None:
+        if self._fd is NULL:
+            return False
+
+        sftp.sftp_close(self._fd)
+        self._fd = NULL
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -1,11 +1,20 @@
 """Tests suite for sftp."""
 
+from __future__ import annotations
+
+import os
 import random
+import typing as _t  # noqa: WPS111
 import uuid
+
+
+if _t.TYPE_CHECKING:  # pragma: no cover
+    import pathlib
+    from collections.abc import Callable
 
 import pytest
 
-from pylibsshext.sftp import SFTP_MAX_CHUNK
+from pylibsshext.sftp import SFTP, SFTP_MAX_CHUNK, _RemoteFile
 
 
 SMALL_PAYLOAD = 32
@@ -122,3 +131,26 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+@pytest.mark.parametrize(
+    'path_modifier',
+    (str, lambda path: str(path).encode('utf-8')),
+    ids=('str', 'bytes'),
+)
+def test_remote_file_cm(
+    sftp_session: SFTP,
+    dst_path: pathlib.Path,
+    path_modifier: Callable[[pathlib.Path], str | bytes],
+) -> None:
+    """Test the ``_RemoteFile`` context manager works with both :class:`str` and :class:`bytes`."""
+    assert not dst_path.exists()
+    formatted_path = path_modifier(dst_path)
+
+    with _RemoteFile(
+        sftp_obj=sftp_session,
+        path=formatted_path,
+        flags=os.O_WRONLY | os.O_CREAT,
+    ):
+        assert dst_path.exists()
+    assert dst_path.read_bytes() == b''
--- a/src/pylibsshext/sftp.pxd
+++ b/src/pylibsshext/sftp.pxd
@@ -23,3 +23,8 @@ from pylibsshext.session cimport Session
 cdef class SFTP:
     cdef Session session
     cdef sftp.sftp_session _libssh_sftp_session
+
+cdef class _RemoteFile:
+    cdef sftp.sftp_file _fd
+    cdef SFTP _sftp
+    cdef _path


src/pylibsshext/sftp.pyx
```diff
@@ -197,3 +165,41 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
+    def __cinit__(self, sftp_obj: SFTP, path: str | bytes, int flags):
```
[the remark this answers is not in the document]
Jakuje: @webknjaz do we care about Python 3.9 or is Copilot off?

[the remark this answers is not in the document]
webknjaz: Technically, Python does not interpret what's inside the `.pyx` files, it can't. Cython does. It'll only hit `.py` files. The type annotations can be free-form strings too. But if there's a type checker, it would attempt to evaluate them (again, in case of pure Python). With Cython, we might want to have separate type annotations in `.pyi` files eventually for external consumers.

Copilot is probably drunk here. We don't run any type-checking within the project (unfortunately). But when we start doing that, we may attempt linting the tests as they are pure-python first.

It seems like Cython itself attempts to infer C-types out of annotations so that you could skip using C-like type syntax: https://cython.readthedocs.io/en/latest/src/userguide/migrating_to_cy30.html#annotation-typing. But I haven't explored all there is to it. There are some notes @ https://cython.readthedocs.io/en/latest/src/tutorial/pure.html#pep-484-type-annotations.

I'd say unless something breaks, I wouldn't care about it too much.

But it's probably a good idea to make sure that anything we change or anything new is covered with tests. And review it manually. We've lost the coverage collection for Cython some time ago for some reason and I haven't had a chance to bisect when and why that happened yet. I think https://github.com/ansible/pylibssh/commit/6f42d20e21c4e36f593cec7fa4340ecff0a6a337 / https://app.codecov.io/gh/ansible/pylibssh/commit/6f42d20e21c4e36f593cec7fa4340ecff0a6a337/tree/src/pylibsshext?dropdown=coverage is the last report known to contain data for `.pyx` files.

src/pylibsshext/sftp.pyx
```diff
@@ -197,3 +165,41 @@ cdef class SFTP:
...
+
+    def __dealloc__(self):
+        if self._fd is not NULL:
+            sftp.sftp_close(self._fd)
+            self._fd = NULL
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
```
webknjaz: Let's have type annotations on this

src/pylibsshext/sftp.pyx
```diff
@@ -197,3 +165,41 @@ cdef class SFTP:
...
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
+    def __cinit__(self, sftp_obj: SFTP, path: str | bytes, int flags):
+        mode = "write" if flags & O_WRONLY else "read"
+        path_b = path
+        if isinstance(path_b, str):
+            path_b = path.encode("utf-8")
```
webknjaz, line 182: I think this should only accept strings and let the caller do the conversion to bytes. Otherwise, the initializer should somehow track the remote file system encoding which might be accessible via that sftp session, but I'm not sure. Encoding conversion is a tricky topic but the agreement on the Python side is to never assume it (in the Python 3 world), which is why it's always a part of the `str` object. And when we're getting `bytes` assuming arbitrary encodings only works for as long as the guess is correct. So at least don't do so in the generic abstractions and push this to the callers so they can fix this later.

docs/changelog-fragments/879.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed memory leaks of remote file descriptors inside ``get()`` and ``put()``
+methods in ``SFTP`` by replacing the manual memory handling with context
```
webknjaz:
```suggestion
methods in ``SFTP`` by replacing the manual memory handling with a context
```

webknjaz (commented):
Is there any testing we can add? If not e2e, maybe for the CM abstraction?

Jakuje:
> Is there any testing we can add? If not e2e, maybe for the CM abstraction?

We test the CM with all the transfers we do get/put in SFTP e2e. If we decide to accept only str, we can write negative test to make sure the bytes are not accepted. For the CM abstraction, the only thing I can try is to try open a file and verify the _fd is non-NULL, I think.

src/pylibsshext/sftp.pyx
Jakuje: Again, what I used here was mostly following existing conventions of the get/put API that handles both of the types.

We need to pass bytes to the libssh C-API,  so accepting str only means we need to convert it to bytes internally for the API. That is easy. We can easily convert str to bytes, but not the other way round (?). If we would like to accept only str here, I think we would need to change the `get/put` API to accept str only too to avoid a need to convert bytes to str. This would be technically breaking change and I wanted to avoid that in this "housekeeping" PR, though I think it can be done as I think its the right direction.

src/pylibsshext/sftp.pyx
```diff
@@ -136,57 +113,44 @@ cdef class SFTP:
...
-                                self._get_sftp_error_str(),
-                            ),
-                        )
-        finally:
-            if read_buffer is not NULL:
-                PyMem_Free(read_buffer)
-        sftp.sftp_close(rf)
+        with _RemoteFile(self, remote_file, O_RDONLY) as remote:
+            try:
+                with open(local_file, 'wb') as f:
```
Jakuje, line 128: The last though I had was if we should make the `RemoteFile` API more like `open()` (mostly in regards to the last argument and do the mapping inside of the API and name it, for example like `remote_open()`). This would also allow us to expose this as public API and make it useful for potential users.

Jakuje (commented):
Just rebased on top of latest changes.

src/pylibsshext/sftp.pyx
```diff
@@ -59,73 +62,47 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
-    def put(self, local_file, remote_file):
+    def put(self, local_file: str | os.PathLike, remote_file: str):
```
webknjaz: Looks like this is lacking a `->`

src/pylibsshext/sftp.pyx
webknjaz: This usually makes sense as a CM-returning method on some sort of a "connection" abstraction. We do this type of thing in the HTTP world. Here's an example from my other project (it's an async CM, but the idea of layering still applies): https://docs.aiohttp.org/en/stable/client_quickstart.html#make-a-request.

So while the `open()` is a global Python built-in that works for local processes without any imports and isn't attached to a class or other explicit object you could do a `with globals().open()` I suppose, but that's silly. Here, we'd need the API to be along the lines of a

```python
with some_maybe_externally_initialized_long_living_connection.open() as remote_fd:
    ... = remote_fd.read()
    remote_fd.write(...)
```
^^^ this is if we want it to feel like a twin of `open()` ^^^

However, the high-level stdlib API for file reads are in pathlib nowadays: it's just `pathlib.Path.read_text()` (or `read_bytes()`). These read the entire contents in-memory.

It feels like we need to define more clearly what we want to get out of this and maybe return to the drawing board and do some designing first.

Meanwhile, maybe we could scope this PR to just the non-breaking changes somehow?

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +162,43 @@ cdef class SFTP:
...
+            self._fd = NULL
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self,
+                 exc_type: type[BaseException] | None,
+                 exc_val: BaseException | None,
+                 exc_tb: TracebackType | None,
+                 ) -> None:
```
webknjaz: This is what typical Python formatters would usually get you. Also, the signature seems wrong. I'm pretty sure it's supposed to return a boolean.
```suggestion
    def __exit__(
        self,
        exc_type: type[BaseException] | None,
        exc_val: BaseException | None,
        exc_tb: TracebackType | None,
    ) -> bool:
```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +162,43 @@ cdef class SFTP:
...
+        if self._fd is not NULL:
```
webknjaz, line 214: If you just invert the guard, you can dedent the entire block below:
```suggestion
        if self._fd is NULL:
            return False
```

webknjaz: Ah, never mind. The incorrect signature confused me.

tests/unit/sftp_test.py
```diff
@@ -122,3 +122,71 @@ def test_put_existing(
...
+        sftp_session.get(
+            str(bogus_src_path).encode('utf-8'),
+            str(bogus_dst_path),
+        )
+
+
+def test_remote_file_cm(sftp_session, src_path):
+    """Test the ``_RemoteFile`` context manager works."""
+    with _RemoteFile(sftp_session, str(src_path), 0) as remote:
+        assert remote
```
webknjaz: Does `_RemoteFile` have a working `__bool__` interface?

tests/unit/sftp_test.py
```diff
@@ -122,3 +122,71 @@ def test_put_existing(
...
+    assert not path.exists()
+    return path
+
+
+@pytest.fixture
+def bogus_dst_path(bogus_file_paths_pair):
+    """Return a data destination path."""
+    path = bogus_file_paths_pair[1]
+    assert not path.exists()
+    return path
```
webknjaz:
```suggestion
@pytest.fixture
def bogus_src_path(tmp_path: pathlib.Path) -> pathlib.Path:
    """Return a data source path."""
    return tmp_path / 'bogus-src-file.txt'


@pytest.fixture
def bogus_dst_path(tmp_path: pathlib.Path) -> pathlib.Path:
    """Return a data destination path."""
    return tmp_path / 'bogus-dst-file.txt'
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +122,71 @@ def test_put_existing(
...
+            str(bogus_src_path),
+            str(bogus_dst_path).encode('utf-8'),
+        )
+
+
+def test_get_bytes(
+    bogus_dst_path,
+    bogus_src_path,
+    sftp_session,
+):
```
webknjaz:
```suggestion
    bogus_dst_path: pathlib.Path,
    bogus_src_path: pathlib.Path,
    sftp_session: 'SFTP',
) -> None:
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +122,71 @@ def test_put_existing(
...
+    path = bogus_file_paths_pair[1]
+    assert not path.exists()
+    return path
+
+
+def test_put_bytes(
+    bogus_dst_path,
+    bogus_src_path,
+    sftp_session,
+):
```
webknjaz:
```suggestion
    bogus_dst_path: pathlib.Path,
    bogus_src_path: pathlib.Path,
    sftp_session: 'SFTP',
) -> None:
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +122,71 @@ def test_put_existing(
...
+        )
+
+
+def test_remote_file_cm(sftp_session, src_path):
+    """Test the ``_RemoteFile`` context manager works."""
+    with _RemoteFile(sftp_session, str(src_path), 0) as remote:
+        assert remote
+
+
+def test_remote_file_cm_bytes(sftp_session, bogus_dst_path):
```
webknjaz:
```suggestion
def test_remote_file_cm_bytes(sftp_session: 'SFTP', bogus_dst_path: pathlib.Path) -> None:
```

tests/unit/sftp_test.py
webknjaz: It's probably more useful to copy something back and forth put+get under the with block and then read the contents.

The `assert` will just confuse a casual reader.

src/pylibsshext/sftp.pyx
Jakuje: Thanks for formatting hings.

I see some `__exit__` is `bool | None` here. Some other are either of them:

https://github.com/python/typeshed/blob/c2944c46cb980246c5e68ca713ee81e53ac0c019/stdlib/contextlib.pyi#L78-L85

tests/unit/sftp_test.py
Jakuje: We really do not have any API that would use the remote file descriptor like this. (unless we would call direct libssh API). The only less confusing option I can think of would be just calling `pass` here.

src/pylibsshext/sftp.pyx
Jakuje: Yeah, that would be for another time I guess. Moved the breaking part to #885.

src/pylibsshext/sftp.pyx
Jakuje: Actually the cython-lint complains about this formatting. Thats the reason why I used this specific format: to make the cython lint happy. With what you suggest, we are getting:
```
- hook id: cython-lint
- exit code: 1

/home/jjelen/devel/pylibssh/src/pylibsshext/sftp.pyx:204:9: E128 continuation line under-indented for visual indent
/home/jjelen/devel/pylibssh/src/pylibsshext/sftp.pyx:205:9: E128 continuation line under-indented for visual indent
/home/jjelen/devel/pylibssh/src/pylibsshext/sftp.pyx:206:9: E128 continuation line under-indented for visual indent
/home/jjelen/devel/pylibssh/src/pylibsshext/sftp.pyx:207:5: E124 closing bracket does not match visual indentation
```

tests/unit/sftp_test.py
Jakuje: lints are not happy with `pass` either and suggest adding some doc string. Not sure if it is right. If you have strong opinion please, let me know.
I changed this also to open the file in write mode, which allows us to check the (empty) file was created, which is likely the most we can do with existing API.

src/pylibsshext/sftp.pyx
Jakuje: Added. Sorry

src/pylibsshext/sftp.pyx
webknjaz: Weird. I've applied it locally and it didn't complain. Let me try pushing and see what the CI says...

tests/unit/sftp_test.py
```diff
@@ -122,3 +127,33 @@ def test_put_existing(
...
+def test_remote_file_cm(sftp_session: SFTP, dst_path: pathlib.Path) -> None:
+    """Test the ``_RemoteFile`` context manager works."""
+    assert not dst_path.exists()
+    with _RemoteFile(
+        sftp_session,
+        str(dst_path),
+        POSIX_O_WRONLY | POSIX_O_CREAT,
+    ):
+        """ File created """
+    assert dst_path.exists()
```
webknjaz: My understanding is that the file will be created right once it's open, right? If that's correct, we would check that it's there upon the call to `__enter__()`. A string literal node looks weird here, otherwise.
```suggestion
        assert dst_path.exists()
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +127,33 @@ def test_put_existing(
...
+    """Test the ``_RemoteFile`` context manager accepts :class:`bytes`."""
+    assert not dst_path.exists()
+    path_b = str(dst_path).encode('utf-8')
+    with _RemoteFile(
+        sftp_session,
+        path_b,
+        POSIX_O_WRONLY | POSIX_O_CREAT,
+    ):
+        """ File created """
+    assert dst_path.exists()
```
webknjaz:
```suggestion
        assert dst_path.exists()
```
same?

tests/unit/sftp_test.py
```diff
@@ -122,3 +127,33 @@ def test_put_existing(
...
+        sftp_session,
+        str(dst_path),
+        POSIX_O_WRONLY | POSIX_O_CREAT,
+    ):
+        """ File created """
+    assert dst_path.exists()
+    assert dst_path.read_bytes() == b''
+
+
+def test_remote_file_cm_bytes(
```
webknjaz: Looks like this is exactly the same test with the only difference being the argument type. Apply `@pytest.mark.parametrize` to the first one with `str` and `lambda p: str(p).bytes('utf-8')` as params with explicit `ids=` to reduce duplication but keep both cases covered. With that, you'll get a callable injected as a test param — you can then wrap what you pass to the remote file initializer with that callable, and it'll get whatever data type each variant returns.

src/pylibsshext/sftp.pyx
Jakuje: Interesting. It might be I have some old versions in tox?

tests/unit/sftp_test.py
Jakuje: Right. Added parametrize

tests/unit/sftp_test.py
Jakuje: Yeah. This works. Thanks for the suggestion!

tests/unit/sftp_test.py
Jakuje: Right. Thanks.

tests/unit/sftp_test.py
```diff
@@ -1,14 +1,27 @@
...
 
+# The Callable[[pathlib.Path], str | bytes] syntax is not supported natively by Python 3.9
+from __future__ import annotations
+
 import random
 import uuid
+from typing import TYPE_CHECKING
+
+
+if TYPE_CHECKING:
```
webknjaz:
```suggestion
import typing as _t


if _t.TYPE_CHECKING:
```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +167,46 @@ cdef class SFTP:
...
+                    sftp_obj._get_sftp_error_str(),
+                ),
+            )
+
+    def __dealloc__(self) -> None:
+        if self._fd is not NULL:
+            sftp.sftp_close(self._fd)
+            self._fd = NULL
+
+    def __enter__(self):
```
webknjaz: This could probably use some typing
```suggestion
    def __enter__(self) -> _t.Self:
```

tests/unit/sftp_test.py
Jakuje: Adjusted with the noqa to make linter happy

src/pylibsshext/sftp.pyx
Jakuje: Added. Resolved also the macos test issue.

src/pylibsshext/sftp.pyx
webknjaz: It's not in `tox`'s directly managed virtualenvs, but in `pre-commit`'s (one layer lower) — it manages its own virtualenvs somewhere under `~/.cache/pre-commit` — you could try wiping it to see if that helps. But I suspect that you might've accidentally copied (or pasted) the suggestion with one less whitespace which would also cause the linter to produce such output.

Is it still unhappy on your machine? The CI's chill with the patch I've pushed to your branch otherwise.

Jakuje: You are probably right. I probably copied it wrong initially.Tox is now happy both on my machine and in CI.

tests/unit/sftp_test.py
```diff
@@ -1,11 +1,21 @@
 """Tests suite for sftp."""
 
+# The Callable[[pathlib.Path], str | bytes] syntax is not supported natively by Python 3.9
+from __future__ import annotations
+
+import os
 import random
+import typing as _t  # noqa: WPS111
```
webknjaz, line 8: partial context: https://github.com/wemake-services/wemake-python-styleguide/pull/3657

*(at some point, we'll get this fixed by bumping WPS and maybe tweaking the flake8 config when I have a minute)*

tests/unit/sftp_test.py
```diff
@@ -1,11 +1,21 @@
 """Tests suite for sftp."""
 
+# The Callable[[pathlib.Path], str | bytes] syntax is not supported natively by Python 3.9
```
webknjaz, line 3: Let's drop this comment. This `__future__` import idiom is a pretty common convention and so it does not require explaining.
```suggestion
```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +172,46 @@ cdef class SFTP:
...
+        if self._fd is NULL:
+            mode = "writing" if flags & O_WRONLY else "reading"
+            raise LibsshSFTPException(
+                "Opening remote file [%s] for %s failed with error [%s]"
+                % (
+                    path,
+                    mode,
+                    sftp_obj._get_sftp_error_str(),
+                ),
+            )
```
webknjaz, line 198: Let's flip this to reduce the indentation of that block below. It's easier on the eyes with early interruptions. Also, since you've started adding some refactoring in, instead of strictly moving blocks of code, I suppose we could also replace percent-interpolation with an f-string too.
```suggestion
        if self._fd is not NULL:
            return

        file_open_mode = "writing" if flags & O_WRONLY else "reading"
        sftp_error = sftp_obj._get_sftp_error_str()
        raise LibsshSFTPException(
            f"Opening remote file [{path!s}] for {file_open_mode!s} "
            f"failed with error [{sftp_error!s}]",
        )
```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +172,46 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
+    def __cinit__(self, sftp_obj: SFTP, path: str | bytes, int flags) -> None:
+        path_b = path
```
webknjaz, line 180: FTR, in `ansible/ansible`, there's a decade-long convention of prefixing vars meant to hold bytes with `b_`. I've never seen people using suffixes. Not sure if there's any existing practice of this in the repo here. But keep this in mind for the future.

src/pylibsshext/sftp.pyx
```diff
@@ -69,44 +75,25 @@ cdef class SFTP:
...
-                if written != length:
-                    sftp.sftp_close(rf)
-                    raise LibsshSFTPException(
-                        "Writing to remote file [%s] failed with error [%s]"
-                        % (
-                            remote_file,
-                            self._get_sftp_error_str(),
-                        )
-                    )
+            with _RemoteFile(self, remote_file, O_WRONLY | O_CREAT | O_TRUNC) as remote:
```
webknjaz, line 81: I somehow missed this initially, but 
we can actually have a chain of CMs in the same block. This allows us to avoid unnecessary indentation which improves readability, while still preserves the nice cleanup perks of CMs in our code. So let's do this.

Additionally, I think we have to think about naming things and interfaces a little more here:
* `remote` is rather generic and it sorta confuses me what it's supposed to be. It's probably a resource or a file descriptor. But it's unclear because if I scroll below, there's something accessing its private attributes
* modern Python versions have a concept of positional-only and keyword-only parameters/args, when designing new APIs (even internal), it's a good practice to make use of those, especially when the list of params grows — passing explicitly named arguments will make it easier for readers to understand what's called and how


```python
with open(local_file, "rb") as f, _RemoteFile(
    sftp_obj=self,
    path=remote_file,
    flags=O_WRONLY | O_CREAT | O_TRUNC,
) as remote_file_resource:
```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +172,46 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
+    def __cinit__(self, sftp_obj: SFTP, path: str | bytes, int flags) -> None:
```
webknjaz, line 179: This is how to make the args keyword-only
```suggestion
    def __cinit__(self, *, sftp_obj: SFTP, path: str | bytes, int flags) -> None:
```

src/pylibsshext/sftp.pyx
```diff
@@ -69,44 +75,25 @@ cdef class SFTP:
...
+                    length = len(read_buffer)
+                    written = sftp.sftp_write(remote._fd, c_buf, length)
+                    if written != length:
+                        raise LibsshSFTPException(
+                            "Writing to remote file [%s] failed with error [%s]"
+                            % (
+                                remote_file,
+                                self._get_sftp_error_str(),
+                            )
+                        )
```
webknjaz, line 95: Since this accesses the private `remote._fd` attribute, I have a feeling that it attempts to implement something that should be a public method — should we just have a   `_RemoteFile.write_buffer(c_buf, /)` for exactly this? And inside, it'd abstract away most of the while block (except for the last line) in a nice method?

src/pylibsshext/sftp.pyx
Jakuje: I used existing conventions from this file:

https://github.com/ansible/pylibssh/blob/41868ef599daffc341a4541d88b5e7cd7c9bc92c/src/pylibsshext/sftp.pyx#L75-L77

If ansible has different conventions, I am happy to change.

I can see one use of the `b_` convention in session:
```
src/pylibsshext/session.pyx:        cdef bytes b_password
```
while several more in scp, sftp and session:
```
src/pylibsshext/session.pyx:                value_b = value.encode("utf-8")
src/pylibsshext/sftp.pyx:            remote_file_b = remote_file.encode("utf-8")
src/pylibsshext/scp.pyx:            remote_file_b = remote_file.encode("utf-8")
src/pylibsshext/scp.pyx:            remote_file_b = remote_file.encode("utf-8")
...
```

I really do not have a strong opinion towards either.

src/pylibsshext/sftp.pyx
Jakuje: Good idea. Moving the read/write into the _RemoteFile makes perfect sense. Please have a look.

src/pylibsshext/sftp.pyx
Jakuje: Adjusted to match this.

src/pylibsshext/sftp.pyx
Jakuje: Great suggestion! I also though about that but did not want to pile up more changes here. But changed that accordingly, which makes it significantly easier to follow. I used `remote_fd` name, which should be hopefully clear. And changed also the symmetric part in get().

src/pylibsshext/sftp.pyx
Jakuje: I used the `b_name` in the new code.

src/pylibsshext/sftp.pyx
Jakuje: Thanks for suggestion. Did that exchange and changed all the exception strings that I touched.

tests/unit/sftp_test.py
Jakuje: Dropped. Thanks

src/pylibsshext/sftp.pyx
```diff
@@ -69,44 +73,19 @@ cdef class SFTP:
...
-                        )
-                    )
-                read_buffer = f.read(SFTP_MAX_CHUNK)
-            sftp.sftp_close(rf)
+        cdef bytearray write_buffer = bytearray(SFTP_MAX_CHUNK)
+
+        with open(local_file, "rb") as local_fd, _RemoteFile(
+            sftp_obj=self,
+            path=remote_file,
+            flags=O_WRONLY | O_CREAT | O_TRUNC
```
webknjaz: (non-blocking, nice-to-have) not *very* important, but trailing commas are usually very nicer on diffs
```suggestion
            flags=O_WRONLY | O_CREAT | O_TRUNC,
```

src/pylibsshext/sftp.pyx
```diff
@@ -69,44 +73,19 @@ cdef class SFTP:
...
+        ) as remote_fd:
+            while True:
+                bytes_read = local_fd.readinto(write_buffer)
+                if bytes_read == 0:
+                    break  # EOF
```
webknjaz: Readability trick — naming our breakout conditionals makes comments unnecessary:
```suggestion
                is_eof = bytes_read == 0
                if is_eof:
                    break
```

webknjaz: This could also be case for the walrus operator but it's arguably uglier, it's say this is a "no":
```suggestion
            while (bytes_read := local_fd.readinto(write_buffer)) != 0:
```

src/pylibsshext/sftp.pyx
```diff
@@ -136,57 +113,27 @@ cdef class SFTP:
...
+        buffer_view = memoryview(read_buffer)
+
+        with _RemoteFile(
+            sftp_obj=self,
+            path=remote_file,
+            flags=O_RDONLY
+        ) as remote_fd, open(local_file, 'wb') as local_fd:
+            while True:
+                file_data = remote_fd.read(read_buffer)
+                if file_data == 0:
```
webknjaz: (same with naming the check)

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +145,85 @@ cdef class SFTP:
         if error in MSG_MAP and error != sftp.SSH_FX_FAILURE:
             return MSG_MAP[error]
         return "Generic failure: %s" % self.session._get_session_error_str()
+
+
+cdef class _RemoteFile:
+    """Helper class managing lifetime of remote file handle."""
```
webknjaz: How about
```suggestion
    """Remote file handle facade.

    Implements a standard context manager interface for
    convenient access lifetime management.
    """

```

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +145,85 @@ cdef class SFTP:
...
+        written = sftp.sftp_write(self._fd, c_buf, length)
+        if written == length:
+            return written
+
+        sftp_error = self._sftp._get_sftp_error_str()
+        raise LibsshSFTPException(
+            f"Writing to remote file [{self._path!s}] failed with error [{sftp_error}]"
+        )
+
+    def read(self, buffer: bytearray) -> int:
```
webknjaz: Plz avoid using `buffer` as a local variable name so that the `buffer` from `globals()` is not shaddowed.

src/pylibsshext/sftp.pyx
```diff
@@ -198,3 +145,85 @@ cdef class SFTP:
...
+        """
+        Reads data from the remote file descriptor into provided bytearray.
+
+        :raises LibsshSessionException: When reading failed.
+        """
+        cdef char* c_buf = buffer
```
webknjaz: `buffer` is highlighted with blue here because it's in stdlib.

src/pylibsshext/sftp.pyx
Jakuje: originally, we had two calls to read() (before cycle and at the end). It was ugly. Gemini suggested to use the `while True`, but this is also an antipatern. While the walrus is very dense, it arguably makes the code cleaner so I think I will make use of it here.

src/pylibsshext/sftp.pyx
Jakuje: sorry, forgot about that again. Is there some static analysis check that could catch this?

src/pylibsshext/sftp.pyx
Jakuje: Used your wording. Sounds good.

src/pylibsshext/sftp.pyx
Jakuje: ok, RHEL9's cython (0.29.37) does not support walrus so we cant use this:
```
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
          with open(local_file, "rb") as local_fd, _RemoteFile(
              sftp_obj=self,
              path=remote_file,
              flags=O_WRONLY | O_CREAT | O_TRUNC,
          ) as remote_fd:
              while (bytes_read := local_fd.readinto(write_buffer)) != 0:
                               ^
  ------------------------------------------------------------

  src/pylibsshext/sftp.pyx:83:30: Expected ')', found ':'
```
Reverting back to the while True:

PR #885, opened 2026-08-26
sftp: Restrict remote path arguments of get()/put() to str

Jakuje:
##### SUMMARY
Previously, this API used to accept also raw `bytes` type, which
required something on the way to guess the real encoding. Anything else
than `str` now raises `TypeError`.

##### ISSUE TYPE
- Bugfix Pull Request

### DIFF
--- a/docs/changelog-fragments/885.breaking.rst
+++ b/docs/changelog-fragments/885.breaking.rst
@@ -1,0 +1,6 @@
+Changed the ``SFTP.get()`` and ``SFTP.put()`` API to accept only
+:class:`str` type for remote paths.
+
+Previously, this API used to accept also raw :class:`bytes` type, which
+required something on the way to guess the real encoding. Anything else
+than :class:`str` now raises :exc:`TypeError` -- by :user:`Jakuje`.
--- a/src/pylibsshext/sftp.pyx
+++ b/src/pylibsshext/sftp.pyx
@@ -15,6 +15,12 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 
+import typing as _t  # noqa: WPS111
+
+
+if _t.TYPE_CHECKING:  # pragma: no cover
+    import os
+
 from posix.fcntl cimport O_CREAT, O_RDONLY, O_TRUNC, O_WRONLY
 
 from cpython.mem cimport PyMem_Free, PyMem_Malloc
@@ -59,15 +65,13 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
-    def put(self, local_file, remote_file):
+    def put(self, local_file: str | os.PathLike[str], remote_file: str) -> None:
         """
         Upload local file to remote server.
 
         :param local_file: The file name on the local file system to upload
-        :type local_file: str or os.PathLike
 
         :param remote_file: The path to upload the file on the remote system
-        :type remote_file: str or bytes
         """
         cdef sftp.sftp_file rf
         cdef const char* c_buf
@@ -108,24 +112,19 @@ cdef class SFTP:
                 read_buffer = f.read(SFTP_MAX_CHUNK)
             sftp.sftp_close(rf)
 
-    def get(self, remote_file, local_file):
+    def get(self, remote_file: str, local_file: str | os.PathLike[str]) -> None:
         """
         Download remote file to local path.
 
         :param remote_file: The file path on the remote system to download
-        :type remote_file: str or bytes
 
         :param local_file: The path on the local file system to place the downloaded file
-        :type local_file: str or os.PathLike
         """
         cdef sftp.sftp_file rf
         cdef char *read_buffer = NULL
         cdef sftp.sftp_attributes attrs
 
-        remote_file_b = remote_file
-        if isinstance(remote_file_b, unicode):
-            remote_file_b = remote_file.encode("utf-8")
-
+        remote_file_b = remote_file.encode("utf-8")
         attrs = sftp.sftp_stat(self._libssh_sftp_session, remote_file_b)
         if attrs is NULL:
             raise LibsshSFTPException(
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -1,11 +1,12 @@
 """Tests suite for sftp."""
 
+import pathlib
 import random
 import uuid
 
 import pytest
 
-from pylibsshext.sftp import SFTP_MAX_CHUNK
+from pylibsshext.sftp import SFTP, SFTP_MAX_CHUNK
 
 
 SMALL_PAYLOAD = 32
@@ -122,3 +123,43 @@ def test_put_existing(
     """Check that SFTP file upload works when target file exists."""
     sftp_session.put(str(src_path), str(pre_existing_dst_path))
     assert pre_existing_dst_path.read_bytes() == transmit_payload
+
+
+@pytest.fixture
+def bogus_src_path(tmp_path: pathlib.Path) -> pathlib.Path:
+    """Return a bogus source path that does not exist."""
+    return tmp_path / 'bogus-dst-file.txt'
+
+
+@pytest.fixture
+def bogus_dst_path(tmp_path: pathlib.Path) -> pathlib.Path:
+    """Return bogus destination path that does not exist."""
+    return tmp_path / 'bogus-src-file.txt'
+
+
+def test_put_bytes(
+    bogus_dst_path: pathlib.Path,
+    bogus_src_path: pathlib.Path,
+    sftp_session: SFTP,
+) -> None:
+    """Check that SFTP put() API does not accept :class:`bytes` as remote file path."""
+    error_msg = r"^Argument 'remote_file' has incorrect type \(expected str, got bytes\)$"
+    with pytest.raises(TypeError, match=error_msg):
+        sftp_session.put(
+            str(bogus_src_path),
+            str(bogus_dst_path).encode('utf-8'),
+        )
+
+
+def test_get_bytes(
+    bogus_dst_path: pathlib.Path,
+    bogus_src_path: pathlib.Path,
+    sftp_session: SFTP,
+) -> None:
+    """Check that SFTP ``get()`` API does not accept :class:`bytes` as remote file path."""
+    error_msg = r"^Argument 'remote_file' has incorrect type \(expected str, got bytes\)$"
+    with pytest.raises(TypeError, match=error_msg):
+        sftp_session.get(
+            str(bogus_src_path).encode('utf-8'),
+            str(bogus_dst_path),
+        )
--- a/tests/unit/sftp_test.py
+++ b/tests/unit/sftp_test.py
@@ -1,8 +1,16 @@
 """Tests suite for sftp."""
 
+# The Callable[[pathlib.Path], str | bytes] syntax is not supported natively by Python 3.9
+from __future__ import annotations
+
 import pathlib
 import random
+import typing as _t  # noqa: WPS111
 import uuid
+
+
+if _t.TYPE_CHECKING:  # pragma: no cover
+    from collections.abc import Callable
 
 import pytest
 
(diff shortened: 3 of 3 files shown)


tests/unit/sftp_test.py
```diff
@@ -122,3 +122,43 @@ def test_put_existing(
...
+    bogus_src_path: pathlib.Path,
+    sftp_session: 'SFTP',
+) -> None:
+    """Check that SFTP put() API does not accept :class:`bytes` as remote file path."""
+    error_msg = r"^Argument 'remote_file' has incorrect type \(expected str, got bytes\)$"
+    with pytest.raises(TypeError, match=error_msg):
+        sftp_session.put(
+            str(bogus_src_path),
+            str(bogus_dst_path).encode('utf-8'),
+        )
```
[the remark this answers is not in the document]
Jakuje: did the import.

src/pylibsshext/sftp.pyx
```diff
@@ -108,24 +108,19 @@ cdef class SFTP:
                 read_buffer = f.read(SFTP_MAX_CHUNK)
             sftp.sftp_close(rf)
 
-    def get(self, remote_file, local_file):
+    def get(self, remote_file: str, local_file: str | os.PathLike) -> None:
```
webknjaz: let's make the pathlike objects a bit more specific
```suggestion
    def get(self, remote_file: str, local_file: str | os.PathLike[str]) -> None:
```

webknjaz: Is it possible to add support for accepting pathlib objects for local paths or does Cython reject them by default now?

src/pylibsshext/sftp.pyx
```diff
@@ -59,15 +61,13 @@ cdef class SFTP:
             sftp.sftp_free(self._libssh_sftp_session)
             self._libssh_sftp_session = NULL
 
-    def put(self, local_file, remote_file):
+    def put(self, local_file: str | os.PathLike, remote_file: str) -> None:
```
```suggestion
    def put(self, local_file: str | os.PathLike[str], remote_file: str) -> None:
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +123,43 @@ def test_put_existing(
...
+    return tmp_path / 'bogus-dst-file.txt'
+
+
+@pytest.fixture
+def bogus_dst_path(tmp_path: pathlib.Path) -> pathlib.Path:
+    """Return bogus destination path that does not exist."""
+    return tmp_path / 'bogus-src-file.txt'
+
+
+def test_put_bytes(
```
webknjaz, line 172: *(given `import typing as _t` at the top)*


We could try passing both `str` and pathlib variants of the local path:

```suggestion
@pytest.fixture(
    ids=repr,
    params=(
        str,
        pathlib.Path,
    ),
)
def normalize_path(request) -> _t.Callable[[pathlib.Path], _t.T | str]:
    return request.param


def test_put_bytes(
    normalize_path: _t.Callable[[pathlib.Path], _t.T | str],
```

tests/unit/sftp_test.py
```diff
@@ -122,3 +123,43 @@ def test_put_existing(
...
+    bogus_dst_path: pathlib.Path,
+    bogus_src_path: pathlib.Path,
+    sftp_session: SFTP,
+) -> None:
+    """Check that SFTP put() API does not accept :class:`bytes` as remote file path."""
+    error_msg = r"^Argument 'remote_file' has incorrect type \(expected str, got bytes\)$"
+    with pytest.raises(TypeError, match=error_msg):
+        sftp_session.put(
+            str(bogus_src_path),
```
webknjaz, line 181:
```suggestion
            normalize_path(bogus_src_path),
```
(hopefully this should be enough to test both cases)

webknjaz: Though, it doesn't really matter for the test that checks the failure scenario of the second argument. We probably need a successful scenario for the pathlib input dual-type testing instead.

docs/changelog-fragments/885.breaking.rst
webknjaz, line 1: Are you sure this API accepted `bytes` in the past? Is there any evidence of this anywhere? Type annotations don't count as they mean nothing in runtime. Are you able to dig up any of the past builds published to PyPI where it's possible to do this, and it accepts `bytes` on the API level? I thought you mentioned Cython rejecting them somehow automatically. If that's true, it may be the case that this never worked, in which case, this wouldn't be breaking. Can we verify this somehow?

src/pylibsshext/sftp.pyx
Jakuje: It should be supported. Because local_file is kept as a standard Python object and passed directly to Python's built-in open(), Python natively unwraps the pathlib.Path object for us. Cython doesn't reject it because we aren't extracting a C-pointer from it.

What we do in the tests is actually a pathlib.Path already

tests/unit/sftp_test.py
Jakuje: I guess this would better do for the test_put() where we can verify both of the representations are accepted.

docs/changelog-fragments/885.breaking.rst
Jakuje: It indeed does accept bytest:
```
$ python3 -m venv check_old_pylibssh
$ source check_old_pylibssh/bin/activate
(check_old_pylibssh) $ pip install --extra-index-url=https://ansible.github.io/pylibssh/simple/  --pre ansible-pylibssh==1.4.0
Looking in indexes: https://pypi.org/simple, https://ansible.github.io/pylibssh/simple/
Collecting ansible-pylibssh==1.4.0
  Downloading ansible_pylibssh-1.4.0-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl.metadata (68 kB)
Downloading ansible_pylibssh-1.4.0-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl (2.7 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 2.7/2.7 MB 50.4 MB/s  0:00:00
Installing collected packages: ansible-pylibssh
Successfully installed ansible-pylibssh-1.4.0

[notice] A new release of pip is available: 26.0.1 -> 26.2.1
[notice] To update, run: pip install --upgrade pip
(check_old_pylibssh) $ cat test.py
import os

with open("test_upload.txt", "w") as f:
    f.write("Hello world")

ssh = Session()
ssh.connect(host="localhost", user="jjelen")

sftp = ssh.sftp()

try:
    print("Attempting to put() using bytes...")
    sftp.put(
        local_file=b"test_upload.txt", 
        remote_file=b"/tmp/test_upload.txt"
    )
    print("SUCCESS: The old API accepted bytes!")
except Exception as e:
    print(f"FAILED: The old API crashed with {type(e).__name__}: {e}")
finally:
    os.remove("test_upload.txt")
(check_old_pylibssh) $ python3 test.py
ssh_strict_fopen: Failed to open a file /etc/ssh/ssh_known_hosts for reading: No such file or directory
ssh_packet_process: Couldn't do anything with packet type 80
Attempting to put() using bytes...
SUCCESS: The old API accepted bytes!
```
If you have other test in mind, please let me know.

webknjaz: @Jakuje do we know what exact change makes it stop accepting bytes?

tests/unit/sftp_test.py
```diff
@@ -1,11 +1,20 @@
 """Tests suite for sftp."""
 
+# The Callable[[pathlib.Path], str | bytes] syntax is not supported natively by Python 3.9
+from __future__ import annotations
+
+import pathlib
 import random
+import typing
```
webknjaz: plz follow this convention. I'll add a linter for it later.
```suggestion
import typing as _t
```

docs/changelog-fragments/885.breaking.rst
webknjaz: @Jakuje also, could you show me what the current trace looks like? Where does it originate?

src/pylibsshext/sftp.pyx
```diff
@@ -15,6 +15,8 @@
 # License along with this library; if not, see file LICENSE.rst in this
 # repository.
 
+import os
```
webknjaz: I know there's no visible existing practice, but it'd be a really good idea to start doing imports in public modules like so:
```suggestion
import os as _os
```

This makes them marked as conventionally private. `import os` puts `os` into the module's `globals()` and by default, unless `__all__` is specified if somebody were to do a `from pylibsshext.sftp import *`, they'd have an `os` imported into their scope. This is a messy practice and should in general be avoided, but historically this project wasn't set up with a good structure to follow.
But I think we should start doing this at some point and be more intentional about making things exposed public API since technically unintentionally this makes `pylibsshext.sftp import os` a public API. https://www.hyrumslaw.com

docs/changelog-fragments/885.breaking.rst
Jakuje: > @Jakuje do we know what exact change makes it stop accepting bytes?

The typing we added to the get/put functions.

There is no technical reason to not accept bytes, but as we discussed, this is bad API and using  str sounds cleaner.

> @Jakuje also, could you show me what the current trace looks like? Where does it originate?

```
$ tox -e build-dists
$ python3 -m venv check_new_pylibssh
$ source check_new_pylibssh/bin/activate
(check_new_pylibssh) $ pip install --force-reinstall --no-deps dist/ansible_pylibssh-1.4.1.dev115+gef18f88.tar.gz
Processing ./dist/ansible_pylibssh-1.4.1.dev115+gef18f88.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Installing backend dependencies ... done
  Preparing metadata (pyproject.toml) ... done
Building wheels for collected packages: ansible-pylibssh
  Building wheel for ansible-pylibssh (pyproject.toml) ... done
  Created wheel for ansible-pylibssh: filename=ansible_pylibssh-1.4.1.dev115+gef18f88-cp314-cp314-linux_x86_64.whl size=410299 sha256=5bd49f4ed2b205d53443aae703dc4f3af5e8b9c6e989043c0b91cb94f9340f6e
  Stored in directory: /home/jjelen/.cache/pip/wheels/d2/d2/fd/0645ac28fd244469014399a5643cc1f3eeb5aee6d9c4d5d421
Successfully built ansible-pylibssh
Installing collected packages: ansible-pylibssh
Successfully installed ansible-pylibssh-1.4.1.dev115+gef18f88

[notice] A new release of pip is available: 26.0.1 -> 26.2.1
[notice] To update, run: pip install --upgrade pip
(check_new_pylibssh) $ python3 test.py
ssh_strict_fopen: Failed to open a file /etc/ssh/ssh_known_hosts for reading: No such file or directory
Attempting to put() using bytes...
FAILED: The old API crashed with TypeError: Argument 'remote_file' has incorrect type (expected str, got bytes)
```

tests/unit/sftp_test.py
Jakuje: The lint complains about this short variable name. But I can noqa it if you prefer

src/pylibsshext/sftp.pyx
Jakuje: given the `os` is only in the typing, we can do the same trick we do with the tests, right?
```
if _t.TYPE_CHECKING:
    import os
```
or is this having also some drawbacks?

Jakuje: code coverage complains about this on.

webknjaz: Technically, not importing things in runtime is good because it improves the startup time. Python 3.15 even has native lazy imports. 

@Jakuje for coverage, you can add a  `# pragma: no cover` on the `if ...:` line. I should actually adjust the config to do this automatically like so: https://github.com/jazzband/pip-tools/blob/c58d5d945c0f20e9713e10f5fea2491a5c984f39/.coveragerc#L14. Feel free to send a PR with this.

docs/changelog-fragments/885.breaking.rst
webknjaz: @Jakuje you've suppressed the traceback. Should you show me the entirety of it? As in, everything? Just add another `raise` on a separate line after that print in the except block.

Jakuje: re-raising does not really show much more:
```
$ python3 test.py
ssh_strict_fopen: Failed to open a file /etc/ssh/ssh_known_hosts for reading: No such file or directory
Attempting to put() using bytes...
FAILED: The old API crashed with TypeError: Argument 'remote_file' has incorrect type (expected str, got bytes)
Traceback (most recent call last):
  File "/home/jjelen/devel/pylibssh/test.py", line 21, in <module>
    raise e
  File "/home/jjelen/devel/pylibssh/test.py", line 14, in <module>
    sftp.put(
    ~~~~~~~~^
        local_file=b"test_upload.txt",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        remote_file=b"/tmp/test_upload.txt"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
TypeError: Argument 'remote_file' has incorrect type (expected str, got bytes)
```

webknjaz: Is this difference pre/post this PR? Could you bisect? I'm a bit confused because I didn't expect that Cython would cause this behavior based on typing alone. Could you confirm what change exactly makes Cython validate the params?

Jakuje: Yes, it does. At `devel` (41868ef599daffc341a4541d88b5e7cd7c9bc92c):
```
(typecheck) jjelen@fedora:~/devel/pylibssh (devel)$ python3 test.py
...
Attempting to put() using bytes...
SUCCESS: The old API accepted bytes!
```
and the previous attempt was at Jakuje:fixup-sftp10 (37703751fd02aab496ec30fc55d1b010c64c42ef). Reran again:
```
(typecheck) jjelen@fedora:~/devel/pylibssh (fixup-sftp10)$ python3 test.py
...
Attempting to put() using bytes...
FAILED: The old API crashed with TypeError: Argument 'remote_file' has incorrect type (expected str, got bytes)
Traceback (most recent call last):
  File "/home/jjelen/devel/pylibssh/test.py", line 21, in <module>
    raise e
  File "/home/jjelen/devel/pylibssh/test.py", line 14, in <module>
    sftp.put(
    ~~~~~~~~^
        local_file=b"test_upload.txt",
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        remote_file=b"/tmp/test_upload.txt"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
TypeError: Argument 'remote_file' has incorrect type (expected str, got bytes)
```

This PR (added typing) is causing the TypeError. I do not know how it work and since in cython this works.

PR #888, opened 2026-09-11
replace expandvars with the Python standard library functionality

f321x:
##### SUMMARY

Replaces the build backend's `expandvars` dependency with `os.path.expandvars`. 

It's great that python provides a built-in alternative to this dependency, so why not use it.
This allows all projects transitively depending on pylibssh to shrink their dependency graph and reduce their supply-chain risk.
The [`expandvars`](https://github.com/sayanarijit/expandvars/) library is maintained by a single person and hasn't seen a commit in 9 months, it seems more sane to depend on a well-maintained implementation of this functionality by the CPython project as such dependencies are one phishing E-Mail away from disaster.

##### ADDITIONAL INFORMATION

Initializes unset configured environment variables to empty strings before expansion. This preserves `LDFLAGS = "-lssh ${LDFLAGS}"` when `LDFLAGS` is unset, while retaining caller-supplied flags. Restore the original environment even if expansion fails.

Expansion follows Python's `$VAR` and `${VAR}` rules. Unknown variables outside the configured environment mapping remain literal, and shell parameter operators such as `${VAR:-default}` are no longer supported. The current build configuration uses only the self-reference above.


docs/changelog-fragments/888.packaging.rst
```diff
@@ -0,0 +1,2 @@
+Removed the ``expandvars`` build dependency while preserving existing
```
webknjaz:
```suggestion
Removed the :pypi:`expandvars` build dependency while preserving existing
```

f321x: Changed in https://github.com/ansible/pylibssh/compare/e7984845376fba6880b3d96b75a047f6848f829a..c8b4f2d61ff78dfa7bc32389ca42b54802fbb44c

packaging/pep517_backend/_cython_configuration.py
webknjaz, line 1: Why did this all move into the try-block so suddenly?

webknjaz:
I'll need to look into why you decided to make the diff so big. Based on the PR description, it doesn't sound like it needs to be. Also, I need to try to remember why I needed the lib in the first place. It's likely that it was due to the need to support Python 2.7, not the extended Bash syntax so dropping that sounds fine. Might need to stare at `git blame` for a bit.

> [!tip]
> Though, using the commit cadence to decide if something is maintained as an excuse for security theater is a shady practice. It doesn't really correlate to the supply chain scenario with a phishing email. So this point sounds like nonsense to me, I won't accept it as valid. I have a lot of things that don't get a lot of commits, but they also don't need to get a lot of commits either — having a script that adds one empty commit a day on cron is a good way to game the metric but does not contribute anything valuable.

webknjaz:
@rodrigobnogueira heads up: this may conflict w/ your changes in https://github.com/aio-libs/yarl/pull/1808 that need to be synched here.

packaging/pep517_backend/_cython_configuration.py
f321x: So every modification to the environment is guaranteed to be cleaned up by the finally block in case anything raises an exception. `setdefault`, `os.environ.update` and `expandvars` could all raise if incorrect arguments are passed, so it seems saner to just cover them in the try finally block.

f321x:
> why you decided to make the diff so big

Not sure what specific part you mean by this, assuming you mean the new `setdefault` loop through the env variables, this is required to keep the behavior of the external library as it would internally replace unset vars with empty strings which `os.path.expandvars` does not do. (see https://github.com/sayanarijit/expandvars/blob/bfdd3f6cf86934f2c3318dd949676aa382be4a89/expandvars.py#L341-L348).

> using the commit cadence to decide if something is maintained as an excuse for security theater

I'd say it depends on what the library does, for `expandvars` you are right I guess, for a library which does networking or is exposed to untrusted input I`d consider it a red-flag if they haven't received any commits in the last months considering the significant amount of vulnerabilities uncovered by newer LLMs. Even if the library is not security critical, I'd still argue that you can find loads of bugs with newer LLMs if the developer hasn't done so himself in the last months, which might then indirectly get chained in some larger scheme to exploit something. 
You're right that commit cadence probably is not an indication of how easy it is for someone to compromise a repository, but having few stars and few consumers of a library makes it easier to go unnoticed for longer.
Whatever, in general I think it is just saner to remove dependence on 3rd parties wherever reasonable. Having a nearly drop-in stdlib alternative seems to be a good alternative.

webknjaz:
> It's likely that it was due to the need to support Python 2.7, not the extended Bash syntax so dropping that sounds fine. Might need to stare at `git blame` for a bit.

Looking at [`e33e4f5` (#32)](https://github.com/ansible/pylibssh/pull/32/changes/e33e4f55bc2aecf9a9c0a73df20338e8fa1b9a6a) and the surrounding commits, I think it was indeed Python 2 support so moving to just stdlib sounds okay. I think at some point we'll migrate aiohttp to also use this backend (or maybe move it out into a shared project) so it's important that all the cases where this might be used are covered: https://github.com/aio-libs/aiohttp/pull/10760#discussion_r2052859065.

docs/changelog-fragments/888.packaging.rst
webknjaz: Ack. Marking as resolved.

packaging/pep517_backend/_cython_configuration.py
webknjaz: Well, let's not sneak in any new behavior into a dependency swap change here. Or mix it with reformatting. It makes the patch messy and impossible to see where the refactoring starts, formatting begins, feature kicks in etc. There's even no atomic commits to review this in isolation. I agreed to accept a small dep swap and I will once it's there. The try-blocks should always be limited to one instruction, but this is dragging in a big refactoring debate on how exactly errors should bubble up into crashes, it seems.

f321x: Not trying to start any debate about error handling here. 
I made the change behavior preserving, see: https://github.com/ansible/pylibssh/compare/c8b4f2d61ff78dfa7bc32389ca42b54802fbb44c..17cd7046d8c9979c9159a406bcd869bdd2f72164

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -170,7 +169,10 @@ def patched_env(
     :yields: None
     """
     orig_env = os.environ.copy()
-    expanded_env = {name: expandvars(var_val) for name, var_val in env.items()}  # type: ignore[no-untyped-call]
```
webknjaz, line 173: Why did you have to remove the type ignore here? I think we're currently not running MyPy here so it won't catch anything in this repo but might fail once the change is ported into yarl and other places that do. Let's keep it and make sure to keep the patch as small as possible.

@f321x additionally, mirror the changes here into the PRs in @aio-libs — I'll want to look at all of them together before doing any merges to make sure that no corner cases that are only hit in other repos are broken.

packaging/pep517_backend/_cython_configuration.py
```diff
@@ -170,7 +169,10 @@ def patched_env(
...
+    # Unset self-references such as ${LDFLAGS} must expand to empty strings.
+    for env_var in env:
+        os.environ.setdefault(env_var, '')
+    expanded_env = {name: expandvars(var_val) for name, var_val in env.items()}
```
webknjaz, line 175:
```suggestion
    expanded_env = {name: expandvars(var_val) for name, var_val in env.items()}  # type: ignore[no-untyped-call]
```

webknjaz: @f321x unless you have a proof that it'll be typed now?

packaging/pep517_backend/_cython_configuration.py
f321x: > Why did you have to remove the type ignore here?

Because the builtin expandvars function provides type hints, unlike the 3rd party expandvars. 
If I re-introduce the flag mypy could actually warn if being run with `--strict` due to `warn_unused_ignores`.
Sure I should re-introduce it?

packaging/pep517_backend/_cython_configuration.py
webknjaz: ack. marking as resolved.

packaging/pep517_backend/_cython_configuration.py
f321x: `expandvars` is typed in the typeshed bundled by mypy: 
https://github.com/python/mypy/blob/53688f7e9385ba738cb34c5964ad96dee5af0868/mypy/typeshed/stdlib/posixpath.pyi#L120-L123

packaging/pep517_backend/_cython_configuration.py
webknjaz: Okay, thanks for explaining. We're good, then. Just trying to cover all the bases since this isn't linted very well in all the projects. Marking as resolved.

packaging/pep517_backend/_cython_configuration.py
webknjaz: Disregarding the request per above and marking as resolved.

webknjaz (approved):
@f321x thanks! I think this is ready, but could you get the aio-libs project PRs in-sync w/ this one so we're sure they're all in the same shape before merging?

f321x:
Thanks for reviewing.

> but could you get the aio-libs project PRs in-sync

I updated all my PRs to aiolibs:
https://github.com/aio-libs/yarl/pull/1879
https://github.com/aio-libs/propcache/pull/294
https://github.com/aio-libs/frozenlist/pull/847

PR #889, opened 2026-09-14
Restore Cython coverage collection in CI

### LINKED ISSUE #781: [TODO] Fix coverage collection for Cython modules

rodrigobnogueira:
##### SUMMARY
Restores coverage collection for the Cython modules, which has reported nothing since #688. Three settings, one commit each:

1. The test jobs install the dist under test with `tox --installpkg` on the pre-populating run and then call `tox` again to run the commands. That second run has no package to install, so tox builds one from the source tree and installs it over the dist, quietly under `-qq`. The rebuild never sees the tracing request, so the modules under test were untraced on every job and every Python version. Before #688 the env reinstalled the pre-built dist in `commands_pre` on every run and never built anything. `--skip-pkg-install` on the second run keeps the dist, as the re-run step already does.
2. The inline test jobs used to gate `ANSIBLE_PYLIBSSH_CYTHON_TRACING` on `profiling-enabled` like the wheels; the reusable workflow that replaced them in 47e2f4f took a `cython-tracing` input with a `0` default instead, and nothing ever passed it, so the source dists were compiled untraced in the test jobs. They now get the same gate as the wheels.
3. Cython 3.1+ routes tracing through `sys.monitoring` on CPython 3.13+, which the `Cython.Coverage` plugin never receives. Traced builds now define `CYTHON_USE_SYS_MONITORING=0`, as Cython's changelog prescribes for coverage. This is the pylibssh half of aio-libs/yarl#1808; the `linetrace` directive is already unconditional here, so yarl's other change does not apply.

Fixes #781.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
Codecov shows `.pyx` data on the devel commits up to 6f42d20 and on none from 83c22c2, the #688 merge, onwards. In the binary-dist jobs the wheel installs in under a second, then the second `tox -qq` is silent for about 30 seconds before pytest starts, consistent with the from-source build (the sdist compile in the first invocation takes 24 seconds). The wheels themselves are traced: the CI-built modules carry `__Pyx_call_line_trace_func`, which Cython compiles only under `CYTHON_TRACE`, and on 3.13 only without `sys.monitoring`, so the define took effect too. The same two `tox` invocations run locally swap the installed module on the second run and drop every `.pyx` from the coverage data; with `--skip-pkg-install` the traced module stays and the `.pyx` files report.

With the dist kept, the binary-dist jobs exercise the manylinux wheels for the first time since #688; those bundle libssh 0.12.2, where the rebuild they had been testing linked the runner's 0.9.6 or 0.10.6. On the Linux runners `tests/unit/channel_test.py::test_open_session_small_timeout` then fails with `DID NOT RAISE`, twice out of two runs, and fail-fast cancels the other Linux jobs before any source-dist job finishes. The aarch64 wheel reproduces it in an Ubuntu 24.04 container: 10 of 10 pass without coverage, 0 of 10 with it; a source build against Ubuntu's libssh 0.10.6 passes 10 of 10 either way, and the macOS wheels with the same libssh pass on CI. The test expects `ssh_channel_open_session()` to return `SSH_AGAIN` within a 1 ms `timeout_usec` with no retries, which the shipped wheels do not do on Linux once a tracer slows the client down. It is left alone here.


webknjaz:
@rodrigobnogueira what's missing for the coverage collection to start working again, then? I was pretty sure the infra was the same as in yarl and it stopped working around the same time too.

webknjaz:
@rodrigobnogueira looks like there are no conflicts to solve after all, but I'm still waiting for ideas on why the coverage hasn't improved..

rodrigobnogueira:
What was missing: since #688 the second tox run in the test jobs rebuilt the package from source, untraced, over the dist it had just installed, quietly under -qq; and the tests workflow's cython-tracing input was never passed, so the sdists compiled untraced too. Both fixed, plus the sys.monitoring define for 3.13+. The macOS jobs now report all seven .pyx (Codecov 67.21%).

rodrigobnogueira:
The Linux matrix is red for a separate reason: the binary-dist jobs now test the manylinux wheels themselves (bundled libssh 0.12.2) for the first time, and test_open_session_small_timeout fails there under a tracer — 0/10 with coverage vs 10/10 without in an Ubuntu container; source builds against the system libssh pass either way.

.github/workflows/reusable-tests.yml
```diff
@@ -243,12 +243,16 @@ jobs:
...
+      # NOTE: Without `--skip-pkg-install`, this second run would rebuild the
+      # NOTE: package from the source tree, untraced, and install it over the
+      # NOTE: pre-built dist put in place above.
       run: >-
         python -m
         tox
         --parallel auto
         --parallel-live
         --skip-missing-interpreters false
+        --skip-pkg-install
```
webknjaz, line 255: Oh, so this was the infra bug, then. I'll pull it out separately as a fix for actually using the passed-in dists rather than the source tree.

webknjaz:
@rodrigobnogueira I've pulled in the last commits. Let's scope this back to py313+ (commit msg + PR description + change note). Note that I'm using natural merges in this repo, so each commit is preserved as is — I won't be squashing your commits or rewriting their contents in a merge.

webknjaz:
@rodrigobnogueira that's a bit too much text (LLM?) but are you saying that `tests/unit/channel_test.py::test_open_session_small_timeout` is failing because with tracing on the lib is slower and needs the timeouts tweaked? Can we do this in a separate PR?

webknjaz:
@Jakuje WDYT about the `tests/unit/channel_test.py::test_open_session_small_timeout` race here?

Jakuje:
> @Jakuje WDYT about the `tests/unit/channel_test.py::test_open_session_small_timeout` race here?

Its odd. It fails when ran together with other tests but works during rerun with max verbosity. Or the rerun is without coverage/tracing?

Timing will always be flaky. We can try to workaround it with reruns (I think we talked about that already, but not sure if in the context of this test), but thats probably all we could do. The value used in `test_open_session_small_timeout` is already the smallest possible for libssh and if the server happens to reply during that time window (and it likely will as its not instrumented), it will just fail like this.

The only option would be somehow make the server slower or inject some hooks into the channel opening, but it sounds to me like to too much effort for unknown gains.

webknjaz:
@Jakuje the rerun is performed w/ `--no-cov` and no parallelism in pytest because that tends to be problematic so it's a little different, yes.

webknjaz:
@Jakuje my understanding is that now that we're always running tests w/ tracing, the client got slower than before, not the server. Right?

Jakuje:
> @Jakuje my understanding is that now that we're always running tests w/ tracing, the client got slower than before, not the server. Right?

That is my understanding. I think the timeout starts only after the packet is (encrypted &) sent, which means that what we are waiting for is really mostly the server operation, which is not affected by tracing.

OTOH, I do not understand why it fails more frequently now. If I see right, its happening in master since 3b3d4d11d5953009efc7c87acf2daa56732cd4cc which "added" the coverage tracing itself.

webknjaz:
@Jakuje yes, `devel` now has tracing for older runtimes, but not for Python 3.13, which is the last thing to pick up from his PR. I'm trying to understand if there's any other influences I haven't taken into account.

webknjaz:
@rodrigobnogueira I've clicked `Rebase` on the GH UI. I think the previous commit's of the change note needs to be recovered, though.

Meanwhile, I'm still trying to understand what's going on with the test races.

PR #890, opened 2026-09-15
Pass the extra compiler flags through CPPFLAGS so wheels keep -O3

bdraco:
##### SUMMARY
The build backend sets `CFLAGS` to add `-DCYTHON_TRACE_NOGIL=1` for coverage builds. setuptools' distutils treats a `CFLAGS` environment variable as a replacement for the interpreter's compiler flags rather than an addition, so those builds were compiled without `-O3` or `-DNDEBUG`. The macro now goes through `CPPFLAGS`, which both the vendored and the legacy distutils append; MSVC ignores both variables, so Windows is unchanged.

The released Linux wheels were affected through a second path: the `cibuildwheel` Linux environment sets `CFLAGS` to the bundled `libssh` include path, which replaces the interpreter's flags the same way, so `ansible-pylibssh` 1.4.0 on PyPI is compiled at `-O0` on Linux (its `session` and `channel` extensions show the frame spill pattern of unoptimized code, while the PR's earlier build with only the backend change still did). That include path now goes through `CPPFLAGS` too, which is the proper variable for `-I`. This is the same latent pattern that made every propcache wheel since 0.5.0 ship unoptimized (aio-libs/propcache#302); the fix is being applied to every project sharing this backend so they stay in sync. Verified locally that a tracing build is optimized and still contains the Cython trace hooks. The new unit tests drive `patched_env` and setuptools' real compiler customization, and fail against the previous backend.

##### ISSUE TYPE
- Bugfix Pull Request

##### ADDITIONAL INFORMATION
With `CFLAGS=' -DCYTHON_TRACE_NOGIL=1'` in the environment, setuptools' compiler customization produces the whole compile command shown below; the interpreter's own flags are gone. With the same value in `CPPFLAGS` they are kept and the macro is appended.

```paste below
CFLAGS unset                     -> clang -fno-strict-overflow -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -O3 -Wall
CFLAGS=' -DCYTHON_TRACE_NOGIL=1' -> clang -DCYTHON_TRACE_NOGIL=1
```


bdraco:
should probably wait to undraft until https://github.com/ansible/pylibssh/pull/888 merges and than switch expandvars here

tests/conftest.py
```diff
@@ -18,6 +20,10 @@
 from pylibsshext import __libssh_version__
 
 
+# Make the in-tree PEP 517 build backend importable from the tests:
+_BUILD_BACKEND_DIR = Path(__file__).parent.parent / 'packaging'
```
webknjaz, line 24: It's better to add such a hack into `pytest.ini`'s `pythonpath` instead of patching the import machinery in runtime.

.coveragerc
```diff
@@ -13,6 +13,10 @@ source =
   *\Lib\site-packages
 
 [report]
+# The in-tree build backend is exercised by tests but is not library
+# code held to the coverage gate.
+omit =
```
webknjaz, line 18: The options here should be sorted alphabetically.

tests/unit/pep517_backend_test.py
```diff
@@ -0,0 +1,65 @@
...
+from pep517_backend._cython_configuration import patched_env
+
+
+CFLAGS_VAR = 'CFLAGS'
+CPPFLAGS_VAR = 'CPPFLAGS'
+TRACE_MACRO = '-DCYTHON_TRACE_NOGIL=1'
+
+
+@pytest.mark.parametrize('tracing', (True, False))
+def test_tracing_macro_goes_through_cppflags(monkeypatch, tracing):
```
webknjaz, line 19: Can we add type annotations to all the new test functions?

```suggestion
def test_tracing_macro_goes_through_cppflags(monkeypatch: pytest.MonkeyPatch, tracing: bool) -> None:
```

pyproject.toml
```diff
@@ -292,3 +303,6 @@ skip = "*"
 
 [tool.setuptools_scm]
 write_to = "src/pylibsshext/_scm_version.py"
+
+[tool.codespell]
```
webknjaz, line 307: Instead of adding non-packaging tool configs to this file, use dedicated tool configs. Add this to a `.codespellrc` instead.

tox.ini
```diff
@@ -14,6 +14,10 @@ deps =
   coverage >= 5.3
   pytest
   pytest-cov
+  # Dependencies of the in-tree PEP 517 build backend under test:
+  expandvars
```
webknjaz, line 18: With the in-flight removal PR #888, this line is up for deletion upon rebase.

pyproject.toml
```diff
@@ -202,7 +202,15 @@ profile = "True"
 [tool.cibuildwheel]
 build-frontend = "build"
 # container-engine = "podman"  # FIXME?
-test-requires = "pytest pytest-cov pytest-xdist"
+test-requires = [
+  "pytest",
+  "pytest-cov",
+  "pytest-xdist",
+  # Dependencies of the in-tree PEP 517 build backend under test:
+  "expandvars",
```
webknjaz, line 210: This will need to be removed after #888
```suggestion
```

.coveragerc
```diff
@@ -13,6 +13,10 @@ source =
...
+  packaging/*
```
webknjaz, line 19: Actually, I'm not sure if I want to outright exclude this as it wasn't before this PR and I don't want to add the coverage report representation discussion to the scope.

tests/unit/pep517_backend_test.py
```diff
@@ -0,0 +1,65 @@
...
+    """Append the macro to CPPFLAGS and leave CFLAGS alone.
+
+    A CFLAGS environment variable replaces the interpreter's own compiler
+    flags in setuptools' distutils, which silently drops -O3 from the
+    build, while CPPFLAGS is appended to them.
+    """
+    monkeypatch.setenv(CFLAGS_VAR, '-fuser-cflag')
```
webknjaz, line 26: Do we need some sort of assertion that it was different before monkey-patching?

tests/unit/pep517_backend_test.py
```diff
@@ -0,0 +1,65 @@
...
+
+from pep517_backend._cython_configuration import patched_env
+
+
+CFLAGS_VAR = 'CFLAGS'
+CPPFLAGS_VAR = 'CPPFLAGS'
+TRACE_MACRO = '-DCYTHON_TRACE_NOGIL=1'
+
+
+@pytest.mark.parametrize('tracing', (True, False))
```
webknjaz, line 18:
```suggestion
@pytest.mark.parametrize(
    'tracing',
    (True, False),
    ids=(
        'with-cython-tracing',
        'without-cython-tracing',
    ),
)
```

tests/unit/pep517_backend_test.py
```diff
@@ -0,0 +1,65 @@
...
+    assert cppflags[0] == '-DUSER=1'
+    assert (TRACE_MACRO in cppflags) is tracing
+    assert os.environ[CPPFLAGS_VAR] == '-DUSER=1'
+
+
+@pytest.mark.skipif(
+    sys.platform == 'win32',
+    reason='MSVC does not read CFLAGS or CPPFLAGS',
+)
+def test_interpreter_flags_survive_the_build_env(monkeypatch):
```
webknjaz, line 42:
```suggestion
def test_interpreter_flags_survive_the_build_env(monkeypatch: pytest.MonkeyPatch) -> None:
```

pyproject.toml
webknjaz: @bdraco we can proceed with this now

